PR #10, opened 2015-12-07
Reduce Krantzberg syndrome in magic getopt

fanf2:
As we know from Charlie Sross's Laundry Files novels, excessive use of magic can cause your brain to be eaten from the inside by terrible entities from another dimension.

This patch series removes `longjmp()` and `goto` from magic getopt, replacing them with `continue` and `if()`. This eliminates a lot of `sigprocmask()` calls, and makes volatile variables unnecessary thereby avoiding a large class of eldritch compiler bugs.

### DIFF
--- a/tests/getopt/Makefile
+++ b/tests/getopt/Makefile
@@ -4,7 +4,7 @@ all:	test_getopt
 
 test:	all
 	@echo -n "Testing command-line options parsing... "
-	@/bin/sh -e test_getopt.sh 2>test_getopt.log
+	@/bin/sh -e test_getopt.sh 2>test_getopt.log || true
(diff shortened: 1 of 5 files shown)


cperciva:
Good point about sigprocmask calls though... I'll switch to using sigsetjmp/siglongjmp so I can indicate that the sigprocmask calls are unnecessary. :-)

cperciva:
I've taken the `return (0);` for `tests/getopt/main.c` and the `|| true` fix for `tests/getopt/Makefile`, and eliminated the unnecessary sigprocmask calls.  But I don't like your other changes so I'm going to close this now.

PR #15, opened 2016-03-31
Add rfc3986 url-encoding

gperciva:


cperciva:
Need to hook this into tests/buildall

Issue #6, opened 2015-08-21
events_network should use poll instead of select


cperciva:
select has a hard limit on the highest socket number it supports; poll has the same (less than ideal) performance with large numbers of sockets, but doesn't have any hard limit.  See https://github.com/Tarsnap/spiped/issues/24.

gperciva:
(for the record, I just edited cperciva's comment so that the `spiped#24` was an actual link.  Interestingly, I can see no record that I changed this; even https://github.com/Tarsnap/spiped/issues/24 says "cperciva just referenced this issue".  I guess github issue editing operates on the honour system?)

cperciva:
Ah, thanks -- I thought I used the right syntax and was confused when it didn't autolinkify.

gperciva:
Easiest way: just copy&paste the actual URL.  github automatically turns it into the syntax you see above.

mc0:
Thank you for all of the great tools from your small team. Just pinging this issue to give some feedback.

This one issue makes the usage of spiped way more complicated for anything involving a decent amount of connections on the server end. We are having to set up an haproxy in front of many spiped processes. I'd love to work on this but I think it'd take me a long time to get a handle on all of how libcperciva works with regard to sockets. Also, it looks like spiped has its own version of libcperciva?[1]

Also, would it be better to also add backwards-compat support for epoll as part of this small rewrite?

1: https://github.com/Tarsnap/spiped/tree/master/libcperciva

gperciva:
Hi @mc0,

I can't speak to the epoll / sockets stuff, but I can clarify libcperciva.  The libcperciva files inside spiped should be the same as those found in the libcperciva respository and other Tarsnap repositories.  If you notice a difference, please let us know so that we can update those files.  :)
(that said, spiped does not have all the libcperciva files -- only those relevant to spiped)

cperciva:
@mc0 As @gperciva says, the copy of libcperciva in spiped is a sparse copy which (for the parts used) should be in sync with libcperciva.  Given that most users of libcperciva only take a few functions, it didn't make sense to build libcperciva into shared object files which other programs would use.

You should be able to replace select with poll without touching anything beyond of https://github.com/Tarsnap/libcperciva/blob/master/events/events_network.c .

I'd prefer to start by switching to poll, since that's portable, before adding support for unportable improvements like epoll / kqueue / etc.

cperciva:
[...]

PR #18, opened 2016-05-17
Use poll(2) instead of select(2) in events/events_network.c

TimWolla:
Closes Tarsnap/libcperciva#6
---

I've done some basic checks using spiped and valgrind. spiped transfers data and valgrind does not report any memory errors when running spiped with the `-1` option. I'm sure I still did something horribly wrong, as I never really worked with event based C networking before. I'm sure you'll examine the patch closely, as always :wink:

### DIFF
--- a/events/events_network.c
+++ b/events/events_network.c
@@ -1,4 +1,4 @@
-#include <sys/select.h>
+#include <poll.h>
 
 #include <errno.h>
 #include <stdlib.h>
@@ -19,12 +19,12 @@ struct socketrec {
 ELASTICARRAY_DECL(SOCKETLIST, socketlist, struct socketrec);
 static SOCKETLIST S = NULL;
 
-/* File descriptor sets containing unevented ready sockets. */
-static fd_set readfds;
-static fd_set writefds;
+/* File descriptors to be polled. */
+static struct pollfd * fds = NULL;
+static nfds_t nfds;
 
 /* Position to which events_network_get has scanned in *fds. */
-static size_t fdscanpos;
+static nfds_t fdscanpos;
 
 /* Number of registered events. */
 static size_t nev;
@@ -46,7 +46,7 @@ initsocketlist(void)
 	nev = 0;
 
 	/* There are no unevented ready sockets. */
-	fdscanpos = FD_SETSIZE;
+	fdscanpos = 0;
 
 done:
 	/* Success! */
@@ -102,7 +102,7 @@ events_network_register(int (*func)(void *), void * cookie, int s, int op)
 		goto err0;
 
 	/* Sanity-check socket number. */
-	if ((s < 0) || (s >= (int)FD_SETSIZE)) {
+	if (s < 0) {
 		warn0("Invalid file descriptor for network event: %d", s);
 		goto err0;
 	}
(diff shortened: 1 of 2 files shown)


events/events_network.c
```diff
@@ -239,30 +249,39 @@ events_network_select(struct timeval * tv)
...
-	FD_ZERO(&writefds);
-
-	/* ... and add the ones we care about. */
-	for (i = 0; i < socketlist_getsize(S); i++) {
-		if (socketlist_get(S, i)->reader)
-			FD_SET(i, &readfds);
-		if (socketlist_get(S, i)->writer)
-			FD_SET(i, &writefds);
+	if (fds == NULL) {
+		if ((fds = calloc(socketlist_getsize(S), sizeof(struct pollfd))) == NULL) {
```
gperciva: Comment here, please.  (something like "Allocate socket list" ?)

events/events_network.c
```diff
@@ -341,6 +376,11 @@ events_network_shutdown(void)
 	if (nev > 0)
 		return;
 
+	if (fds != NULL) {
```
gperciva: Comment here, please.  ("Free the fd list" ?)

events/events_network.c
```diff
@@ -136,6 +136,15 @@ events_network_register(int (*func)(void *), void * cookie, int s, int op)
 		goto err0;
 
 	/*
+	 * There is a new socket/operation pair:
+	 * Rebuild the fd list on the next call to events_network_select().
+	 */
+	if (fds != NULL) {
```
cperciva: Not good.  In the common case that a callback reads/writes data and then immediately re-registers itself to read/write more, this will result in the fd list being regenerated after every network I/O.  Worse, if there are many readable/writable sockets, the lowest-numbered one will always get the chance to do something and the higher-numbered sockets will get starved.

TimWolla: I think this problem existed before as well: The fd sets were cleared for every call to select, no? If not I'd like to understand the difference, as I said: I never worked with event based C networking before.

see: https://github.com/Tarsnap/libcperciva/pull/18/files#diff-6745fd3f7945293c64b6fbcc79828177L243

cperciva: Yes, the fd sets were being cleared for every call to select.  But not for every register/deregister call.  The scenario I'm concerned about here is
1. select gets called (once)
2. 100 active connections get callbacks to read/write data and register new callbacks.

TimWolla: Ah, I see. I'll think about this and see what I'll do about this, might take a bit (depending on my free time :smile:).

cperciva: Taking a bit of time is fine.  This issue has sat open for quite a while already. :-)

But I wonder if it would be enough to set a flag for "regenerate list" rather than throwing it away entirely here...?

gperciva:
@TimWolla, are you still interested in this change?  I just added a few updates to https://github.com/Tarsnap/libcperciva/commits/TimWolla-poll-updated.

My next step is to tackle the "regenerate list" comment that @cperciva made, although like you, I'm new to event-based C networking.  If you're interested and think you have a handle on this problem, I'd be happy to leave it up to you.  :)

TimWolla:
> @TimWolla, are you still interested in this change?

Not at this moment. I lack the time to dive deeper into this.

cperciva:
Closing this for now since the submitter is busy and @gperciva is working on the same functionality in #39.

PR #19, opened 2016-06-18
Include headers

gperciva:


gperciva:
Not certain what you want the test directory to be called.  I initially thought that asprintf needed a similar fix, but once I'd added the infrastructure to test it on its own, I discovered that it was fine.

gperciva:
Nope, cancel my "not certain".  I've got more fixes for files which fail when compiled on their own coming up.  :)

gperciva:
I'm moving the "test compiling headers individually" to a separate PR, since I encountered some difficulty (simply including `warnp.h` wasn't enough to catch the "missing NULL"; it only occurs when the `WARNP_INIT` macro was used).

gperciva:
For the record, there's a few more header stuff to investigate, but they'll require more attention from me and you, so I figured it would be good to get these out of the way first.

gperciva:
So far I've been using 1 commit per file, in the theory that this could avoid problems pulling a commit into a different repository.  If you think that's not a concern, I could group them more.  (for example, putting all of the "include `<errno.h>`" changes into 1 commit)

crypto/crypto_aesctr.h
```diff
@@ -6,6 +6,7 @@
 
 /* Opaque type. */
```
cperciva: type -> types

cperciva:
Note to self: Looks good aside from needing the grammar fix in crypto_aesctr.h

gperciva:
Updated PR.

I could have grouped the files together a bit -- all projects use (warp.h, getopt.c); if (monoclock.c, network_read.c, sock.c) are used in a project then all three are used, likewise (crypto_aesctr.h, entropy.c) are found as a pair or not at all.  But the actual types of header changes in those groups didn't really seem to go together all that well, so I've held off on merging those commits.

That said, I don't think that merging with other projects will be a problem if there's lots of separate commits.  I'd export all of the commits with

    git format-patch 65635c8a60e472ad11f0e1e3b23ace46ddcba9f0

, do a `sed` on the files to adjust the directory name, then do

    git am location/of/patches/*.patch

and simply ignore any patch that failed to apply.  (and I'm also writing a script to check that libcperciva is sync'd between our projects, so any mistakes in the above process would be easily detected)

PR #21, opened 2016-06-19
test/buildall/Makefile*: add CFLAGS

gperciva:


gperciva:
This lets us easily test the build with stuff like `-Wall -Weverything -Werror -Wno-#warning`.

PR #23, opened 2016-06-19
Crypto int conversion

gperciva:


gperciva:
This PR obviously requires careful attention, since "janitorial fixes to crypto code" is a recipe for disaster.  :)

For the `crypto_dh.c` change, at first I added an `int rlen_openssl` and checked if that was less than 0, but then I decided it was overkill since any overflow would be caught anyway.  But if you'd rather be paranoid, I can easily add such a check.

crypto/crypto_aes.c
```diff
@@ -29,9 +29,11 @@ aesnitest(uint8_t ptext[16], uint8_t * key, size_t keylen)
 	void * kexp_aesni;
 	uint8_t ctext_openssl[16];
 	uint8_t ctext_aesni[16];
+	int newkeylen;
 
 	/* Expand the key. */
-	AES_set_encrypt_key(key, keylen * 8, &kexp_openssl);
+	newkeylen = (int) (keylen * 8);
+	AES_set_encrypt_key(key, newkeylen, &kexp_openssl);
```
cperciva: This silences the warning, but doesn't really accomplish anything.  Also, `newkeylen` is unnecessary; you could just put the explicit cast into the function call.

I'd add an assertion (keylen < INT_MAX / 8, since that guarantees the conversion to int won't overflow) and then put the explicit cast into the function call.

crypto/crypto_aes.c
```diff
@@ -57,7 +59,7 @@ useaesni(void)
 	static int aesnigood = -1;
 	uint8_t key[32];
 	uint8_t ptext[16];
-	size_t i;
+	uint8_t i;
```
cperciva: Do the warnings go away if you keep this as a size_t, but add a `& 0xff` where it's used?  That would make for saner code.

crypto/crypto_aes.c
```diff
@@ -89,30 +91,32 @@ useaesni(void)
 #endif /* CPUSUPPORT_X86_AESNI */
 
 /**
- * crypto_aes_key_expand(key, len):
- * Expand the ${len}-byte AES key ${key} into a structure which can be passed
- * to crypto_aes_encrypt_block.  The length must be 16 or 32.
+ * crypto_aes_key_expand(key, keylen):
```
cperciva: I'm not sure I see the point of the global `s/len/keylen/` here.  Also, if you do have a reason for changing this, you should change the _aesni version too.

crypto/crypto_aes.c
```diff
@@ -89,30 +91,32 @@ useaesni(void)
...
 #endif
 
 	/* Allocate structure. */
 	if ((kexp = malloc(sizeof(AES_KEY))) == NULL)
 		goto err0;
 
 	/* Expand the key. */
-	AES_set_encrypt_key(key, len * 8, kexp);
+	newkeylen = (int) (keylen * 8);
+	AES_set_encrypt_key(key, newkeylen, kexp);
```
cperciva: As above, assert and put the explicit cast into the function call.

crypto/crypto_dh.c
```diff
@@ -126,7 +126,7 @@ blinded_modexp(uint8_t r[CRYPTO_DH_PUBLEN], BIGNUM * a,
 		warn0("%s", ERR_error_string(ERR_get_error(), NULL));
 		goto err8;
 	}
-	rlen = BN_num_bytes(r1);
+	rlen = (size_t) BN_num_bytes(r1);
```
cperciva: Probably better to change `rlen` to an int, and add a check for < 0.

crypto/crypto_aes.c
gperciva: Well, I wanted to standardize the names so that I could make exactly the same change in both functions.  You've got

    static int
    aesnitest(uint8_t ptext[16], uint8_t * key, size_t keylen)

and

    struct crypto_aes_key *
    crypto_aes_key_expand(const uint8_t * key, size_t len)

I found the `keylen` name slightly more informative than `len`.  Granted, since the only other argument to `crypto_aes_key_expand` is `key`, I wouldn't say that `len` is a terrible name.  But changing `aesnitest` to use `len` seemed like a bad idea, since then it wouldn't be obvious at first glance whether `len` referred to `ptext` or `key`.

I'm not wedded to this change, though.

crypto/crypto_aes.c
gperciva: My mind boggles that that this is considered "saner", but yes, `& 0xff` works, and I'll include that when I update this PR.

cperciva: So, the idea here is that a "straightforward" reading of the code should match what it actually does.  `i` is an array index, so it should be a `size_t`; where we set a byte equal to `i` or `0x11 * i` what we really mean is "set this byte to the lowest byte of..." and `& 0xff` is a way of making that explicit rather than simply relying on the implicit type conversion.

crypto/crypto_aes.c
cperciva: I think we can normalize on `len` then; any time there's a `size_t len` parameter it should refer to the immediately preceding pointer value, so I don't think there's much risk of confusion.

gperciva:
Updated PR with almost completely different commits.  :)

crypto/crypto_aes.c
```diff
@@ -70,9 +73,9 @@ useaesni(void)
 
 		/* Test cases: key is 0x00010203..., ptext is 0x00112233... */
 		for (i = 0; i < 16; i++)
-			ptext[i] = 0x11 * i;
+			ptext[i] = 0x11 * (i & 0xff);
```
cperciva: I meant `(0x11 * i) & 0xff`.  We're taking the final byte of the multiplication result.

crypto/crypto_aes.c
```diff
@@ -70,9 +73,9 @@ useaesni(void)
...
 		for (i = 0; i < 32; i++)
-			key[i] = i;
+			key[i] = i & 0xff;
```
cperciva: Is this masking necessary?  I thought clang would recognize that `i` was always within the range for a uint8_t here.

crypto/crypto_dh.c
```diff
@@ -127,11 +127,15 @@ blinded_modexp(uint8_t r[CRYPTO_DH_PUBLEN], BIGNUM * a,
...
+	if (rlen < 0) {
+		warn0("Unexpected error in OpenSSL");
+		goto err8;
+	}
 	if (rlen > CRYPTO_DH_PUBLEN) {
 		warn0("Exponent result too large!");
 		goto err8;
 	}
-	memset(r, 0, CRYPTO_DH_PUBLEN - rlen);
+	memset(r, 0, CRYPTO_DH_PUBLEN - (size_t) rlen);
```
cperciva: No whitespace in the typecast `(foo)bar`.

crypto/crypto_dh.c
```diff
@@ -127,11 +127,15 @@ blinded_modexp(uint8_t r[CRYPTO_DH_PUBLEN], BIGNUM * a,
...
 	BN_bn2bin(r1, r + CRYPTO_DH_PUBLEN - rlen);
```
cperciva: And if you needed the `(size_t)` on the previous line, why not here too?

gperciva: I looked this up previously:
```
void *memset(void *s, int c, size_t n);
int BN_bn2bin(const BIGNUM *a, unsigned char *to);
```
Evidently clang is happy to mix signed and unsigned ints when it comes to pointer math?

crypto/crypto_aes.c
gperciva: Nope, it doesn't recognize it:
```
../../crypto/crypto_aes.c:78:13: warning: implicit conversion loses integer
      precision: 'size_t' (aka 'unsigned long') to 'uint8_t'
      (aka 'unsigned char') [-Wconversion]
                        key[i] = i;
```

gperciva:
Updated PR.  Haven't changed anything about the arguments to `BN_bn2bin()`; I think it's ok as it is, but if you'd like a comment explaining that the second argument is pointer math, I'm cool with adding that.

PR #24, opened 2016-06-19
Build each header file individually

gperciva:


gperciva:
This ended up being not very interesting; the only problem it found was `sock_util.h`.  `warnp.h` is only detected if `WARNP_INIT` is used.  That said, having written the code, I figured I might as well toss it up here.  If you don't think it's worth adding, I won't be offended.  :)

NB: if this is pulled, it should come after #19 , otherwise `make test` will report an error (the `sock_util.h` issue).

cperciva:
Note to self: Looks good, merge after #19 .

PR #25, opened 2016-06-22
Style comment return

gperciva:


gperciva:
I left these as separate commits because, although kivaloo uses both of them, tarsnap and spiped only use timerqueue.
(I'm working on a script that investigates which files are used in which projects, so that I can group changes more sensibly)

PR #26, opened 2016-06-23
Include headers

gperciva:


gperciva:
I grouped these commits because in both cases, either all 3 files were used, or none.

cperciva:
For the <sys/types.h> bit, please add the following to the commit message: "The <sys/types.h> header has historically been required on some systems when <sys/socket.h> is used, but the POSIX standard states that it is not necessary.  In keeping with our policy of strict compliance, we omit these inclusions; but it may be necessary to add them back on some buggy platforms."

cperciva:
For the <sys/time.h> bit, I'd prefer to keep the inclusion; while declaring the struct is *probably* sufficient, I'm not sure that it's guaranteed to be good enough if, for example, a platform has `#define timeval timeval64` or similar evilness which might make our declaration of 'struct timeval' different from the one the OS provides.

So my policy of declaring opaque struct types in headers should only apply to our library code, not to anything coming from "elsewhere" (OS, OpenSSL, etc.).

gperciva:
Should I add something to the `posix-cflags.sh` for `sys/socket.h` without `sys/types.h`?

cperciva:
> Should I add something to posix-cflags.sh for sys/socket.h without sys/types.h?

Not yet.  I don't know if any of the buggy platforms still exist, and I don't want to clutter /POSIX/ with workarounds which are no longer relevant.

gperciva:
`sys/types.h` now submitted as #29.

gperciva:
Still thinking about how best to approach `<sys/time.h>`.

cperciva:
I think the best approach to `<sys/time.h>` is to simply leave it there.  Consider its definition of `struct timeval` to be equivalent to `<stddef.h>` defining `size_t`.

gperciva:
I'm not disputing keeping `<sys/time.h>`; I was trying to find a way to tell include-what-you-use to stop bugging me about it via its mapping file.  I've successfully dealt with some other false positives this way, but so far it hasn't worked for `<sys/time.h>`.

I've made some notes on paper to come back to it, so I can close this issue now.

PR #27, opened 2016-06-27
Fix warnings

gperciva:


alg/crc32c.c
```diff
@@ -63,7 +63,7 @@ init(void)
 
 	/* Fill in tables. */
 	for (i = 0; i < 256; i++) {
-		r = reverse(i);
+		r = reverse(i & 0xffff);
```
cperciva: Not sure why you're masking to take the bottom 16 bits here... in any case, I'd go with a cast to `(uint32_t)`.

network/network_read.c
```diff
@@ -69,7 +69,7 @@ callback_buf(void * cookie)
 		goto tryagain;
 
 	/* Invoke the callback and return. */
-	return (docallback(C, C->bufpos));
+	return (docallback(C, (ssize_t)C->bufpos));
```
cperciva: Maybe also add an assertion that bufpos <= SSIZE_MAX?  We have an assertion in network_read but not one close to here.

alg/crc32c.c
gperciva: erm, yes, well, as a cellist, counting isn't my strong point.  :)

Curious, though: elsewhere, you've preferred `& 0xff` instead of a cast.  Is that because this is 32 bits, or because it's not in the crypto directory, or...?

util/entropy.c
```diff
@@ -55,7 +55,7 @@ entropy_read(uint8_t * buf, size_t buflen)
 
 		/* We've filled a portion of the buffer. */
 		buf += lenread;
```
cperciva: I'd cast lenread to (size_t) both times, even if clang doesn't care about pointer arithmetic.

util/noeintr.c
```diff
@@ -42,11 +42,11 @@ noeintr_write(int d, const void * buf, size_t nbyte)
 		 * pointer and adjust the remaining write length.
 		 */
 		p += lenwrit;
```
cperciva: I'd cast lenread to (size_t) both times, even if clang doesn't care about pointer arithmetic.

alg/crc32c.c
cperciva: In that other code, we were doing some arithmetic and then saying "take the last byte of this".  Here we're saying "this is really a 32-bit integer and has been all along".

Either way works (and they should produce identical code) but I like to have the "natural reading" of the code match my mental model of what we're trying to do, if that makes sense.

gperciva: but what if "what we're trying to do" doesn't make sense?

... ok, I'll go order pizza now.

gperciva:
Updated PR

PR #32, opened 2016-08-06
cpusupport.h: add forward declaration

gperciva:


gperciva:
This came from https://github.com/Tarsnap/scrypt/pull/59

PR #35, opened 2016-08-07
Fix warnings

gperciva:


gperciva:
I'm less certain that these are the correct fixes.  Mostly because I think they don't fix all theoretical problems -- we can get an integer overflow if we exceed `INT64_MAX` entries in the map.

I think, at least.  `seqptrmap_add()` doesn't check any size limit, nor does `elasticqueue_add()`. `elasticarray_append()` does check for an overflow:

```
        /* Check for overflow. */
        if ((nrec > SIZE_MAX / reclen) ||
            (nrec * reclen > SIZE_MAX - EA->size)) {
                errno = ENOMEM;
```

however, if `reclen = 1`, then I think we can still fail.  On 64-bit Ubuntu,
```
INT64_MAX =  9223372036854775807
SIZE_MAX  = 18446744073709551615UL
```
so something can fit into `SIZE_MAX` but still overflow `INT64_MAX`.

(granted, even if that overflow check doesn't trigger an error, the next step is to `resize()` the elastic array, and I admit that it's unlikely that anybody will be running this code on a system with that much RAM, because 640k should be enough for anybody.)


So yeah.  I might have missed some weird corner case (or maybe those corner cases are protected in other ways), but in any case, I think I've done as much digging into these fixes as I can.

datastruct/seqptrmap.c
```diff
@@ -57,7 +57,7 @@ seqptrmap_add(struct seqptrmap * M, void * ptr)
 	M->len += 1;
```
cperciva: How about adding an assertion here to check that we don't overflow int64_t when we compute the value to return?  Being careful to not have an integer overflow while checking the assertion, of course...

datastruct/seqptrmap.c
```diff
@@ -74,7 +74,7 @@ seqptrmap_add(struct seqptrmap * M, void * ptr)
 void *
 seqptrmap_get(struct seqptrmap * M, int64_t i)
 {
-	size_t pos = i - M->offset;
+	size_t pos = (size_t)(i - M->offset);
```
cperciva: Hmm, this code is actually buggy with a 32-bit size_t... if `i` is 2^32 away from a valid value, we could return a pointer when we should have returned NULL.

gperciva:
Updated PR

datastruct/seqptrmap.c
```diff
@@ -56,8 +57,11 @@ seqptrmap_add(struct seqptrmap * M, void * ptr)
 		goto err0;
 	M->len += 1;
 
+	/* Check for overflow. */
+	assert(INT64_MAX - (int64_t)M->len >= M->offset);
```
cperciva: Need to assert that M->len <= (uint64_t)(INT64_MAX) first, since otherwise there could be a signed integer overflow when performing the cast.

datastruct/seqptrmap.c
```diff
@@ -74,7 +78,12 @@ seqptrmap_add(struct seqptrmap * M, void * ptr)
 void *
 seqptrmap_get(struct seqptrmap * M, int64_t i)
 {
-	size_t pos = i - M->offset;
+	size_t pos;
+
+	/* Calculate position after checking that it will fit into size_t. */
+	if (i - M->offset >= (int64_t)SIZE_MAX)
```
cperciva: Does this work?  What if i < M->offset?

gperciva: Bette reason why it won't work: `(int64_t)SIZE_MAX` can be `-1`.

gperciva:
Updated PR, imported from https://github.com/Tarsnap/kivaloo/pull/20.

PR #37, opened 2016-08-26
style: use parentheses for return

gperciva:


cperciva:
Yeah, ok.  I hadn't bothered with STYLE in these simple test files, but I guess consistency is good...

gperciva:
I noticed it while doing `git grep "return 0"` so I figured it would bug me if I didn't fix it.

PR #39, opened 2016-10-04
Use poll instead of select; updated version

gperciva:


gperciva:
This is an updated version of #18.  (should I continue it here, or move it over to spiped so that we can see it in action?)

It's a tentative first step; some parts are suspect.  The spiped tests pass, but I haven't done any benchmarking on latency or bandwidth or whatever.

events/events_network.c
```diff
@@ -233,36 +240,60 @@ events_network_cancel(int s, int op)
...
+
 	/* Initialize if necessary. */
 	if (initsocketlist())
 		goto err0;
 
-	/* Zero the fd sets... */
-	FD_ZERO(&readfds);
-	FD_ZERO(&writefds);
+	/* Clear list if a rebuild is requested. */
+	if (fds_rebuild_needed) {
```
cperciva: Why free the memory allocation just to reallocate it a moment later?

events/events_network.c
```diff
@@ -292,30 +323,47 @@ struct eventrec *
...
+	if (fds == NULL)
+		return (r);
+
 	/* Scan through the fd sets looking for ready sockets. */
 	for (; fdscanpos < nfds; fdscanpos++) {
+		struct pollfd * current = &fds[fdscanpos];
+
+		/* File descriptor was closed, ignore it in future polls. */
+		if (current->revents & POLLNVAL) {
+			current->fd = -1;
```
cperciva: This should probably be an assert, not a silent ignore.  In the select-based code, we would never even get this far -- select returns EBADF if you ask to select on an invalid descriptor.

events/events_network.c
```diff
@@ -233,36 +240,60 @@ events_network_cancel(int s, int op)
...
+		}
+		nfds = 0;
+		fds_rebuild_needed = 0;
+
+		/* ... and add the ones we care about. */
+		for (i = 0; i < socketlist_getsize(S); i++) {
+			if (socketlist_get(S, i)->reader ||
+			    socketlist_get(S, i)->writer) {
+				fds[nfds].fd = (int)i;
+				if (socketlist_get(S, i)->reader)
```
cperciva: We should check for POLLHUP and POLLERR too -- caused by the arrival of TCP FIN and RST packets, for example -- since these mark the socket as "readable" for the purpose of select (and this behaviour is assumed by users of this code).

events/events_network.c
```diff
@@ -292,30 +323,47 @@ struct eventrec *
...
+			current->fd = -1;
+			continue;
+		}
+
 		/* Are we ready for reading? */
-		if (FD_ISSET(fdscanpos, &readfds)) {
-			r = socketlist_get(S, fdscanpos)->reader;
-			socketlist_get(S, fdscanpos)->reader = NULL;
+		if (current->revents & POLLIN) {
+			r = socketlist_get(S, (size_t)current->fd)->reader;
```
cperciva: If `r` is NULL here, we should keep scanning.  This can happen if _cancel has been called, since we're no longer clearing the socket readable/writable bits.

cperciva: ... actually, it's even worse than that: Imagine the sequence:
1. _select gets called
2. _cancel gets called
3. the socket is closed
4. a new socket is opened
5. _register gets called
6. _get is called

In that case we'd be reporting an event *on a completely different socket* to the one which was previously detected as ready.  We need some sort of validity marker which _cancel can clear, or possibly a freshness value we can compare between the event registration and when poll was called.

events/events_network.c
gperciva: fixed

events/events_network.c
gperciva: Changing it to an assert broke stuff; there isn't anywhere else that handles "file descriptor was closed".

cperciva:
I decided to rewrite this from scratch, with pollfd kept up-to-date as events are registered and cancelled instead of being completely regenerated when `events_network_select` is called.  Committed as 91f42f7f50f9e1c61ada00f4724b0d761a4b56c0.

Please let me know if you see any problems...

PR #41, opened 2016-10-27
Avoid tautology on 32-bit systems

gperciva:


gperciva:
This was part of https://github.com/Tarsnap/kivaloo/pull/26

PR #48, opened 2017-01-22
Switch events_network from select to poll

cperciva:
This has several effects: Most notably, it removes the FD_SETSIZE
limitation on descriptors, but it will also improve performance in
the unlikely case of waiting on a very sparse set of descriptors
(since poll is only passed the active list), and it may yield a
modest speedup in general (since the list is no longer regenerated
on each call to events_network_select, although the kernel still
needs to examine the entire list).

Note that this change is likely to result in events being returned
in a different order if multiple descriptors become ready at the
same time; this may expose previously-hidden bugs.

This change adds a dependency on ctassert.h; projects which do not
already include that file will need to add it.

### DIFF
--- a/events/events_network.c
+++ b/events/events_network.c
@@ -1,37 +1,77 @@
 #include <sys/select.h>
 
+#include <assert.h>
 #include <errno.h>
+#include <limits.h>
+#include <poll.h>
+#include <stdint.h>
 #include <stdlib.h>
+#include <string.h>
 
+#include "ctassert.h"
 #include "elasticarray.h"
 #include "warnp.h"
 
 #include "events.h"
 #include "events_internal.h"
 
+/*
+ * Sanity checks on the nfds_t type: POSIX simply says "an unsigned integer
+ * type used for the number of file descriptors", but it doesn't make sense
+ * for it to be larger than size_t unless there's an undocumented limit on
+ * the number of descriptors which can be polled (since poll takes an array,
+ * the size of which must fit into a size_t); and nfds_t should be able to
+ * the value INT_MAX + 1 (in case every possible file descriptor is in use
+ * and being polled for).
+ */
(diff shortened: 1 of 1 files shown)


events/events_network.c
```diff
@@ -1,37 +1,77 @@
...
+/**
+ * Invariants:
+ * 1. Initialized entries in S and fds point to each other:
+ *     S[i].pollpos < nfds ==> fds[S[i].pollpos].fd == i
+ *     j < nfds ==> S[fds[j].fd].pollpos == j
+ * 2. Descriptors with events registered are in the right place:
+ *     S[i].reader != NULL ==> S[i].pollpos < nfds
+ *     S[i].writer != NULL ==> S[i].pollpos < nfds
+ * 3. Descriptors without events registered aren't in the way:
+ *     (S[i].reader == NULL && S[i].writer != NULL) ==> S[i].pollpos == -1
```
TimWolla: Should this read: `(S[i].reader == NULL && S[i].writer == NULL)`?

gperciva:
@cperciva wrote:
> Note that this change is likely to result in events being returned in a different order if multiple descriptors become ready at the same time; this may expose previously-hidden bugs.

Yeah, `spipe` now sometimes passes a `struct timeval * tv=NULL` to `events_network_select()`.

gperciva:
Ok, answering my own comment: passing `NULL` to `events_network_select()` was not invalid in the previous version.  Instead,

>  If the timeout parameter is a null pointer, then the call to pselect() or select() shall block indefinitely until at least one descriptor meets the specified criteria.
http://pubs.opengroup.org/onlinepubs/9699919799/functions/select.html

So I'll make `poll()` block in the new version (if `tv=NULL`).

events/events_network.c
cperciva: Good catch.  @gperciva, can you add this to your PR?

PR #49, opened 2017-01-22
Poll fixes

gperciva:


gperciva:
With these changes, the tests for spiped and kivaloo (without S3 tests) complete successfully.

events/events_network.c
```diff
@@ -360,10 +363,14 @@ events_network_select(struct timeval * tv)
 	 * Convert timeout to an integer number of ms.  We round up in order
 	 * to avoid creating busy loops when 0 < ${tv} < 1 ms.
 	 */
-	if (tv->tv_sec >= INT_MAX / 1000)
-		timeout = INT_MAX;
-	else
-		timeout = tv->tv_sec * 1000 + (tv->tv_usec + 999) / 1000;
+	if (tv != NULL) {
```
cperciva: Can you put this backwards, so that it becomes if ... else if ... else ?  No point having unnecessary indentation.

cperciva:
Huh.  I'm not surprised by the missing casts, but I must have been tired to miss the static declaration and the NULL timeout handling.  Well, that's why we have tests, I guess...

In case you didn't get an email: Please add Tim's comment fix (https://github.com/Tarsnap/libcperciva/pull/48/files/ff803364d8df10c718e56876fab1cdc02792d056#r97227373).

gperciva:
Updated PR; still works with the spiped test.

PR #53, opened 2017-02-09
poll: fix spelling "initialzed".

gperciva:


gperciva:
It's amazing what I can discover while waiting for our parents to return my phone call asking whether I should shovel their driveway this morning!

PR #54, opened 2017-03-13
Parsenum additions

gperciva:


gperciva:
This isn't really a "serious" PR; it's just more convenient to have a central place to discuss parsenum.  The last commit (so far) ["add more testing for parsenum"] tests including the header in a file which doesn't actually use the functionality.  That demonstrates the value of the "parsenum: silence warnings" commit.

gperciva:
General thought: I'm slightly disappointed by the suggested boilerplate-ness of:
```
GETOPT("-n"):
	if (PARSENUM(&opt_n, optarg, 0, SIZE_MAX)) {
		warnp("Invalid option: -n %s", optarg);
		exit(1);
	}
	break;
```

What about:
```
GETOPT("-n"):
	PARSENUM(&opt_n, optarg, 0, SIZE_MAX, "-n");
	break;
```

or even
```
GETOPT("-n", &opt_n, 0, SIZE_MAX):
	break;
```

?  (naturally, the last option would need to be rolled into getopt.h)

gperciva:
Note: compiling with strict `$CFLAGS` produces a whole raft of errors, mainly about "comparing floating point with == or != is unsafe" and implicit conversions.  I'm sorting out the non-parsenum `$CFLAGS` problems first, then I'll come back to this.

util/parsenum.h
```diff
@@ -18,13 +18,13 @@
 #define PARSENUM(x, s, min, max)					\
 	(								\
 		errno = 0,						\
-		(((x) = 1, (x) /= 2) != 0) ?				\
-			((x) = parsenum_float((s), (min), (max))) :	\
-		(((x) = -1) <= 0) ?					\
-			((x) = parsenum_signed((s), (min), (max))) :	\
+		(((*x) = 1, (*x) /= 2) != 0) ?				\
```
cperciva: This is broken.  You need to replace `(x)` with `(*(x))` in order to avoid problems if `x` is something like `&foo + 5`.

tests/parsenum/main.c
```diff
@@ -1,4 +1,3 @@
-#include <inttypes.h>
```
cperciva: You're going to want it when you declare a bunch more variables and parse numbers into them. :-)

cperciva:
On 03/12/17 23:41, Graham Percival wrote:
> General thought: I'm slightly disappointed by the suggested boilerplate-ness...

That might happen in the future.  But we need the basic routines for switching
on command-line options and parsing numbers before we consider wrapping them
into a more sophisticated macro.

-- 
Colin Percival

util/parsenum.h
gperciva: why on earth would somebody try testing `&foo + 5`, i don't even.

C can be amazingly horrifying.  :)

gperciva:
I'm giving up on clang-happifying parsenum for tonight.  Other than one minor change (`!=` into `>` in the "is this a float" test), I'm not optimistic that we can avoid the problems.

tl;dr:
```
    int x;
    (((x) = 1, (x) /= 2) > 0) ?
        ((x) = parsenum_float((s), (min), (max)))
```

We know that the computer isn't going to reach the `parsenum_float()`, but clang-4.0 doesn't figure that out, so it complains about implicitly converting a float into an int.

cperciva:
On 03/13/17 01:07, Graham Percival wrote:
> We know that the computer isn't going to reach the |parsenum_float()|, but
> clang-4.0 doesn't figure that out, so it complains about implicitly converting
> a float into an int.

The solution here might be to define a macro `SET(x, y)` which is normally
`(x) = (y)` but becomes `(x) = (typeof (x))(y)` with some compilers.  Ugly,
but it should work...

-- 
Colin Percival

gperciva:
`typeof`?  ick, that's a GNU-ism.  :)

One option would be to switch to C11 and use `_Generic`.  I'm pretty sure that's a non-option, though.

For laughs, I coded up a separate macro for `#ifdef __clang__` which disables the relevant variable conversion warnings.  Not certain if you'll prefer it to `typeof`, but I figured it was worth writing so you can take a look.

util/parsenum.h
```diff
@@ -15,8 +15,14 @@
  * nonzero with an unspecified value of ${x} and errno set to EINVAL or ERANGE
  * as appropriate.
  */
+#ifdef __clang__
+/* Disable clang warnings arising from implicit conversion in PARSENUM. */
 #define PARSENUM(x, s, min, max)					\
 	(								\
+_Pragma("clang diagnostic push")					\
```
cperciva: Ooh, I didn't realize `_Pragma` was C99.  OK, let's do it this way, but rather than having two definitions for PARSENUM, define PARSENUM_PROLOGUE and PARSENUM_EPILOGUE as being either these pragmas or nothing.

gperciva:
Updated with `PARSENUM_PROLOGUE` and `_EPILOGUE`.

gperciva:
Some fixes for clang.  What strategy do you want to use for the rebasing?  Just smoosh everything together (while keeping `parsenum.h` and the test code separate)?  Or 4 commits: orig PARSENUM, orig test, 2-arg PARSNUM, 2-arg test?  Or something else?

gperciva:
This code now needs
```
#define GCC diagnostic ignored "-Wtype-limits"
```
otherwise it gets
```
../../util/parsenum.h:78:26: error: comparison of unsigned expression < 0 is always false [-Werror=type-limits]
       (((uintmax_t)(min) > (uintmax_t)(*(x))) ||  \
                          ^
```

Unfortunately, when I try to do the same `PARSENUM_PROLOGUE` trick, gcc borks with:
```
main.c:26:1: error: expected expression before ‘#pragma’
  TEST(&d, "123.456", 0, 1000);
 ^
```

Apparently they changed their pragma-handling sometime between 4.7.2 and 4.7.4, and the later version is stricter about where pragma can appear.

gperciva:
In case you weren't already certain, 92860df makes gcc stop complaining.

cperciva:
See https://github.com/Tarsnap/libcperciva/pull/56

PR #55, opened 2017-03-13
tests: use ${CFLAGS}

gperciva:


gperciva:
Heh, not much point having a strict `$CFLAGS` variable if it's not used in (most) of the tests!

gperciva:
Not certain if you'll like splitting up the getopt test, but it works.

cperciva:
I don't really want to waste time working around compiler-is-stupid warnings when they only related to test code.  Is this all we need?

gperciva:
Yes, that's why this PR passed the Travis-CI tests...?

I'm still working on parsenum stuff, but that's not part of this particular PR.

cperciva:
I wasn't sure if you had all the warnings turned on which you wanted. :-)

gperciva:
Ah.  Yeah, that's the full set of warnings that we can use with clang-3.5.  (note that this commit doesn't modify the actual `CFLAGS` at all; it merely tells the `Makefile`s to use them.  The sha256 test already did so, probably because you needed them for `posix-cflags.sh`.)

As per the commit message in fe75d3acb359006b79ee2a37a8636bd56b44af3f (the one that added the strict tests to Travis-CI), once Travis-CI uses a newer clang, we'll need to add in `-Wno-reserved-id-macro`.

tests/getopt/main.c
```diff
@@ -11,15 +11,12 @@ usage(void)
 	exit(1);
 }
 
-int
-main(int argc, char * argv[])
+static void
+parse_first(int argc, char * argv[])
```
cperciva: I'd call these `pass1` and `pass2`.

tests/getopt/main.c
```diff
@@ -81,5 +87,18 @@ main(int argc, char * argv[])
...
+}
+
+int
+main(int argc, char * argv[])
+{
+
+	/* Process arguments. */
+	parse_first(argc, argv);
+
+	/* We want to make sure that reset works. */
```
cperciva: `/* Reset getopt state and process arguments again. */`

gperciva: For the record, that comment was in the original file.

gperciva:
Updated with `pass1()`, `pass2()`, and changing the `Reset getopt...` comment.  I rebased those changes with the previous commit, since they seemed simple enough.

tests/getopt/main.c
cperciva: Yeah, but it made far more sense in its original context. :-)

PR #56, opened 2017-03-18
Add PARSENUM macros

cperciva:


gperciva:
Problem with the current setup: if I want to test
```
TEST2(&s, "-1");
```
then on my 64-bit desktop, I'll get
```
234.567000 18446744073709551615 0
```
but on a 32-bit machine, we'd see
```
234.567000 4294967295 0
```

We could avoid this by only printing `d, s, i` if we didn't receive an error, but I'm not certain if you'd like that.  I mean, do you want to see the exactly value of those variables even if there was an error?

util/parsenum.h
```diff
@@ -64,7 +64,9 @@ _Pragma("clang diagnostic pop")
 			((*(x)) = parsenum_float((s), (min), (max))) :	\
 		(((*(x)) = -1) <= 0) ?					\
 			((*(x)) = parsenum_signed((s), (min), (max))) :	\
-			(((*(x)) = parsenum_unsigned((s),		\
+			(assert(((min) >= 0) && "PARSENUM unsigned with negative minimum bound"),			\
+			assert(((max) <= *(x)) && "PARSENUM unsigned upper bound does not fit in datatype"),			\
```
cperciva: I don't think we want this.  It would be entirely reasonable to say "parse a size_t and make sure the value is less than 5 billion"; this assertion would then pass on a 64-bit system but fail on a 32-bit system.

(As for a negative minimum bound... it's hard to see why someone would do that, but I can't imagine someone doing that by accident either.)

gperciva: when you wrote
```
TEST(&s, "1234", -123, 4000);
```
was it deliberate?
(BTW, it took me a few minutes before I trained myself to read `s` as `size_t` instead of as `signed`)

tests/parsenum/main.c
```diff
@@ -0,0 +1,55 @@
...
+			warn0("PARSENUM: EINVAL");	\
+		} else if (errno == ERANGE) {		\
+			warn0("PARSENUM: ERANGE");	\
+		} else {				\
+			warnp("PARSENUM");		\
+		}					\
+	} else {					\
+		fprintf(stderr, "%f %zu %d\n", d, s, i);\
+	}						\
+} while (0)
```
cperciva: The more I think about it, the more I think that these are not the right way to write testing macros for PARSENUM.  Something more like

    TEST2_SUCCESS("123.456", double, 123.456);
    TEST2_FAILURE("123.456", size_t, EINVAL);
    TEST4_FAILURE("1234", int, -10, 10 ERANGE);

producing output of

    Parsing "123.456" as type double yields 123.456... PASSED!
    Parsing "123.456" as type size_t yields EINVAL... PASSED!
    Parsing "1234" as type int between -10 and 10 yields ERANGE... PASSED!

might be better?

gperciva: looks much better to me!

gperciva: continued this part in https://github.com/Tarsnap/libcperciva/pull/57

PR #57, opened 2017-03-23
Parsenum2 alt macro

gperciva:


tests/parsenum/main.c
```diff
@@ -16,16 +16,36 @@ main(int argc, char * argv[])
 
 	WARNP_INIT;
 
-#define TEST(x, y, z, w) do {						\
-	fprintf(stderr, "PARSENUM(\"%s\", %s, %s)\n", y, #z, #w);	\
-	if (PARSENUM(x, y, z, w))					\
-		warnp("PARSENUM");					\
-	fprintf(stderr, "%f %zu %d\n", d, s, i);			\
+#define TEST4_PROLOGUE(str, var, min, max, target)		\
```
gperciva: I thought it was cute to avoid repetition by having this `_PROLOGUE` bit, but I figure there's 40% chance you'll tell me to nuke it and just copy the lines.

tests/parsenum/main.c
```diff
@@ -16,16 +16,36 @@ main(int argc, char * argv[])
...
+	var parsenum_x;						\
+	int parsenum_ret;					\
+	fprintf(stderr, "Parsing \"%s\" as %s between %s and %s (incl.) yields " target "... ", str, #var, #min, #max);			\
+	parsenum_ret = PARSENUM(&parsenum_x, str, min, max);
+
+#define TEST4_SUCCESS(str, var, min, max, target) do {		\
+	TEST4_PROLOGUE(str, var, min, max, #target);		\
+	if (parsenum_ret == 0) {				\
+		if ((parsenum_x >= target) && (parsenum_x <= target)) {	\
```
gperciva: ugliness due to trying to avoid floating-point comparisons.  The original code compared floats by printing them to 6 decimal places and comparing the strings.  I could still use this method, of course.  Or I could do a more standard "`fabs(x-y) < 1e-10`" thing.

gperciva: * make that "floating-point equality comparisons".

tests/parsenum/main.c
```diff
@@ -16,16 +16,36 @@ main(int argc, char * argv[])
...
+			fprintf(stderr, "PASSED!\n");		\
+		} else {					\
```
gperciva: oops, I should remove the braces here.

gperciva:
If this looks ok, I'll do it for `TEST2()` as well, then prepare everything suitable for rebasing the existing tests, before adding new tests.

tests/parsenum/main.c
cperciva: Either copy the lines, or pick a different name.  A prologue is something which isn't really part of the routine itself, e.g. fiddling with the stack pointer: https://en.wikipedia.org/wiki/Function_prologue .  What you have here is the the first half of the test (in the sense that a test is 1. running a command 2. checking the results).

BTW you probably want to adjust how you organize the macros anyway, since the check-the-results bits for PARSENUM4 and PARSENUM2 should be the same...

tests/parsenum/main.c
cperciva: I'd go with `fabs(x - y) < 1 && fabs(x - y) <= fabs(x + y) * 0.000001`.  The compiler warning is quite right about testing for equality being dangerous... especially if you're using gcc, which has broken floating-point arithmetic.  Checking `>= && <=` might silence the warning, but it doesn't actually solve the problem.

cperciva:
Did https://github.com/Tarsnap/libcperciva/pull/57#discussion_r107565870 and https://github.com/Tarsnap/libcperciva/pull/57#discussion_r107565448 get lost due to a forced push?

gperciva:
No, I saw them, and I'm working on them.  Comparison is a bit tricky, since the compiler isn't font of shoving unsigned numbers into `fabs()`.  I think I have it working now, though, provided that the compiler doesn't optimize
```
(x/x/2.0 > 0) ?
```
into something useless.

... and also, now that I look at it again, provided that `x!=0`.  Hmm, back to the drawing board.

[basic problem: writing a `PARSENUM_CMP(type, x, y)` that does the right thing.]

cperciva:
On 03/22/17 18:59, Graham Percival wrote:
> No, I saw them, and I'm working on them.

Ok, I was confused by github saying that those comments were on outdated patches.

> Comparison is a bit tricky, since the
> compiler isn't font of shoving unsigned numbers into |fabs()|.

Cast to double?

-- 
Colin Percival

gperciva:
Are you saying that you want me to cast all my problems away?

cperciva:
Not necessarily *all* your problems... but in this case, it seems like a cast would work just fine?  If the difference between two integers is less than one, they're equal. :-)

gperciva:
Updated PR

gperciva:
Should this be valid?
```
TEST4_SUCCESS("inf", double, 0, (double)INFINITY, (double)INFINITY);
```
- it doesn't compile with `gcc`:
    ```
    ../../util/parsenum.h:66:42: error: overflow in implicit constant conversion [-Werror=overflow]
         ((*(x)) = parsenum_signed((s), (min), (max))) : \
    ```
    (now that I see that error again, I think that was why I started adding `assert()` into `PARSENUM`).
- it compiles with `clang`, but fails the test:

    ```
    Parsing "inf" as double between 0 and (double)(__builtin_inff()) (incl.) yields (double)INFINITY... FAILED!
    ```
    If `inf` support is desired, I could add a bunch of `isinf()` / `isnan()` into `PARSENUM_CMP()`, unless there's some sneakier way of doing this.

gperciva:
(err, of course with normal `CFLAGS`, `gcc` merely gives a warning rather than an error.  Still, would be nice to avoid it.)

cperciva:
On 03/23/17 20:35, Graham Percival wrote:
> Should this be valid?
> 
> |TEST4_SUCCESS("inf", double, 0, (double)INFINITY, (double)INFINITY); |

Looks reasonable.

>     it doesn't compile with |gcc|:
> 
>     |../../util/parsenum.h:66:42: error: overflow in implicit constant
>     conversion [-Werror=overflow] ((*(x)) = parsenum_signed((s), (min),
>     (max))) : \ |

Ugh, and we can't disable this warning because gcc has broken _Pragma support?

I don't think we can do much to avoid this -- even without infinities, we
would run into this problem with constants which don't fit into the integer
ranges (whether signed or unsigned).

>     it compiles with |clang|, but fails the test:
> 
>     |Parsing "inf" as double between 0 and (double)(__builtin_inff()) (incl.)
>     yields (double)INFINITY... FAILED! |
> 
>     If |inf| support is desired, I could add a bunch of |isinf()| / |isnan()|
>     into |PARSENUM_CMP()|, unless there's some sneakier way of doing this.

Ah yes, INFINITY - INFINITY is a NaN.  Right, please make PARSENUM_CMP accept
that INFINITY == INFINITY, -INFINITY == -INFINITY, and NaN == NaN.

-- 
Colin Percival

gperciva:
> Ugh, and we can't disable this warning because gcc has broken _Pragma support?

Yep.  Adding a `#pragma` outside of `PARSENUM_PROLOGUE` isn't sufficient, because the macro is expanded in the `.c` file, not the `.h` file.

The "cleanest" option I could find is adding this to `parsenum.h`:

```
/* Urgh, ugly hack. */
#define PARSENUM_GCC_PUSH                                       \
_Pragma("GCC diagnostic push");                                 \
_Pragma("GCC diagnostic ignored \"-Woverflow\"");
        
#define PARSENUM_GCC_POP                                        \
_Pragma("GCC diagnostic pop");
```

and then in `main.c`:

```
        /* Disable misleading GCC warning. */
        PARSENUM_GCC_PUSH;

        TEST4_SUCCESS("123.456", double, 0, 1000, 123.456);
        TEST4_SUCCESS("inf", double, 0, (double)INFINITY, (double)INFINITY);
        TEST4_SUCCESS("1234", size_t, -123, 4000, 1234);
        TEST4_FAILURE("12345", int, -10, 100, ERANGE);

        TEST2_SUCCESS("234.567", double, 234.567);
        TEST2_SUCCESS("2345", size_t, 2345);
        TEST2_FAILURE("abcd", size_t, EINVAL);
        
        /* Re-enable original GCC warnings. */
        PARSENUM_GCC_POP;\
```

I think it's worth considering it -- this makes it clear that we know that there's a GCC issue, we've looked for workarounds, and this is the best we can do.  Otherwise we'll get well-intentioned bug reports about this.

(no wait, there's an even cleaner method: use C11's `_Generic` keyword.  It's been almost 6 years since the finalized draft!  :)

cperciva:
This seems to work:
```
-                       ((*(x)) = parsenum_signed((s), (min), (max))) : \
+                       ((*(x)) = parsenum_signed((s), (*(x) <= 0) ? (min) : 0, (*(x) <= 0) ? (max) : 0)) :     \
```

This ensures that the values in question are not "compile-time constants", even though the optimizer is able to recognize that they are in fact constants. :-)

gperciva:
right, the build fails due to the GCC `-Woverflow`.  I could add `-Wno-overflow` to the `.travis.yml`, or go with the `PARSENUM_GCC_HACK_PUSH`, or just leave it for now.

tests/parsenum/main.c
```diff
@@ -5,8 +5,11 @@
 #include "warnp.h"
 
 #define PARSENUM_CMP(x, y)						\
-	(fabs((double)x - (double)y) < 1 &&				\
-	    fabs((double)x - (double)y) <= fabs((double)x + (double)y) * 1e-6)
+	((isnan(x) == isnan(y) && isnan(x) != 0) ||			\
```
cperciva: Or you could just write this as `isnan(x) && isnan(y)` ?

tests/parsenum/main.c
```diff
@@ -5,8 +5,11 @@
...
+	    (isinf(x) == isinf(y) && isinf(x) != 0) ||		\
```
cperciva: As above, but you need to check the sign as well.

gperciva: err... so
```
isinf((double)x) && isinf((double)y) && ((x > 0) && (y > 0) || (x < 0 && y <0))
```
?  I think my original version is simpler.

[I did change the `isnan()` version, though.]

gperciva: oh wait, I can remove the final `!= 0` bit.

gperciva:
Should this pass or fail?
```
TEST4_SUCCESS("-0", double, 0, 0, 0);
```
I mean, is `-0` the same as `0` for the purpose of PARSENUM?  Right now it is.

tests/parsenum/main.c
```diff
@@ -0,0 +1,97 @@
+#include <stdint.h>
+#include <stdio.h>
+
+#include "parsenum.h"
+#include "warnp.h"
+
+#define PARSENUM_CMP(x, y)						\
+	((isnan((double)x) && isnan((double)y)) ||			\
+	    (isinf((double)x) == isinf((double)y) && isinf((double)x)) || \
```
cperciva: This is a synonym for isinf(x) && isinf(y), I think?

Also, doesn't this make PARSENUM_CMP(INFINITY, -INFINITY) return nonzero?

gperciva: Growl.  glibc says:
> isinf(x)      returns 1 if x is positive infinity, and -1 if x is negative infinity.

but I just checked in POSIX and it makes no such guarantees; merely
> The isinf() macro shall return a non-zero value if and only if its argument has an infinite value.

gperciva: (that's why I thought that `isinf() == isinf()` was more useful than it actually is)

cperciva: Huh, it never occurred to me that isinf might return *different* nonzero values.

cperciva:
I'm not incredibly worried about the `0` vs. `-0` case... but hey, why not?  Easy enough to check if fpclassify returns FP_ZERO and then check if signbit returns the same value.

tests/parsenum/main.c
```diff
@@ -7,7 +7,8 @@
 #define PARSENUM_CMP(x, y)						\
 	((isnan((double)x) && isnan((double)y)) ||			\
 	    ((isinf((double)x) && isinf((double)y)) &&			\
-	     (((x > 0) && (y > 0)) || ((x < 0) && (y < 0)))) ||		\
+		((((double)x > 0) && ((double)y > 0)) ||		\
```
cperciva: FWIW, I'd write this as `(double)x * (double)y > 0` since we've already determined that they're both infinities...

gperciva:
Updated PR to use a function.

cperciva:
Looks good but needs more test cases.  If nothing else, octal and hex need to work; but it would be good to check maximum and minimum values of `uint32_t` and `int32_t` as well.

PR #58, opened 2017-03-26
Parsenum3

gperciva:


gperciva:
Did the rebasing, and added more test cases.  Am I correct that we don't need to test `TEST2` as much if we've covered them in `TEST4` ?

tests/parsenum/main.c
```diff
@@ -0,0 +1,141 @@
...
+	TEST4_SUCCESS("-0x7f", double, -1000, 0, -127);
+	TEST4_SUCCESS("-1e2", double, -1000, 0, -100);
+	TEST4_SUCCESS("1e-2", double, 0, 1000, 0.01);
+
+	TEST4_SUCCESS("1234", size_t, -123, 4000, 1234);
+	TEST4_FAILURE("7f", size_t, 0, 1000, EINVAL);
+	TEST4_SUCCESS("0x7f", size_t, 0, 1000, 127);
+	TEST4_SUCCESS("0x77", size_t, 0, 1000, 119);
+	TEST4_SUCCESS("077", size_t, 0, 1000, 63);
+	TEST4_SUCCESS("0xffffffff", size_t, 0, (size_t)0xffffffff, UINT32_MAX);
```
cperciva: Maybe you mean to parse this into a uint32_t ?

tests/parsenum/main.c
```diff
@@ -0,0 +1,141 @@
...
+	TEST4_FAILURE("0xfffffffff", size_t, 0, (size_t)0xffffffff, ERANGE);
```
cperciva: We should check that PARSENUM returns ERANGE for values outside of the range of the type even if they're within the range passed to the macro.

tests/parsenum/main.c
```diff
@@ -0,0 +1,141 @@
...
+
+	TEST4_FAILURE("12345", int, -10, 100, ERANGE);
+	TEST4_SUCCESS("0x7fffffff", int, 0, INT32_MAX, INT32_MAX);
+	TEST4_SUCCESS("-0x80000000", int, INT32_MIN, 0, INT32_MIN);
```
cperciva: int32_t ?

tests/parsenum/main.c
gperciva: Given that the variable type is `size_t`, why would it be preferable to use `uint32_t`?  I do see that `size_t` can have a max as little as 65535.

tests/parsenum/main.c
gperciva: eh?  Isn't that what I proposed here?  where you said
> It would be entirely reasonable to say "parse a size_t and make sure the value is less than 5 billion"; this assertion would then pass on a 64-bit system but fail on a 32-bit system.

https://github.com/Tarsnap/libcperciva/pull/56#discussion_r107533552

... oh did you mean that you wanted it to return ERANGE, as opposed to having an assertion fail?

tests/parsenum/main.c
gperciva: POSIX specifies that an int must hold 2^31-1 and -(2^31-1).

... wait, what?!  `INT_MIN... -2 147 483 647` http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html

huh, TIL.  Ok, switching it to `int32_t`.

tests/parsenum/main.c
gperciva: Adding
```
TEST4_FAILURE("0xfffffffff", uint32_t, 0, (uintmax_t)0xffffffff, ERANGE);
```
runs into gcc problems:
```
../../util/parsenum.h:71:14: error: comparison is always true due to limited range of data type [-Werror=type-limits]
        (max) >= *(x) ? *(x) : (max))),
```
We could avoid this with an `assert()`, but I don't think that's what you want here.  :)

gperciva: I went ahead and added a `_Pragma` to take care of this, since it's only in the test code.  Could still be an issue for real `PARSENUM()` usage, though.

tests/parsenum/main.c
cperciva: Well, I thought you were trying to make sure that the maximum possible value could be successfully parsed.

tests/parsenum/main.c
cperciva: If someone turns on -Wtype-limits and then says "turn this into a size_t between 0 and 5 billion" on a 32-bit system, they deserve to get a warning.  Heck, never mind *deserve*, they probably *want* that warning.

tests/parsenum/main.c
cperciva: It isn't specified whether `int` is sign-magnitude, 1's complement, or 2's complement; so it could have a `-0` but not a `-2^(N-1)`.

But the point here was to test that we can successfully parse the minimum and maximum values in a signed type, so `int` isn't useful anyway -- it could be a 64-bit value, for example.

tests/parsenum/main.c
```diff
@@ -119,8 +121,13 @@ main(int argc, char * argv[])
...
 	TEST4_SUCCESS("0x77", size_t, 0, 1000, 119);
 	TEST4_SUCCESS("077", size_t, 0, 1000, 63);
-	TEST4_SUCCESS("0xffffffff", size_t, 0, (size_t)0xffffffff, UINT32_MAX);
-	TEST4_FAILURE("0xfffffffff", size_t, 0, (size_t)0xffffffff, ERANGE);
+
+	TEST4_SUCCESS("0xffffffff", uint32_t, 0, (uintmax_t)0xffffffff,
+	    UINT32_MAX);
+	TEST4_FAILURE("0xfffffffff", uint32_t, 0, (uintmax_t)0xffffffff,
+	    ERANGE);
+	TEST4_FAILURE("0xfffffffff", uint32_t, 0, (uintmax_t)0xffffffff,
```
cperciva: This looks like you wrote the same thing twice?

gperciva: right, removed.

gperciva:
Shall I rebase the latest few commits?

tests/parsenum/main.c
```diff
@@ -0,0 +1,155 @@
...
+
+	TEST4_SUCCESS("1234", size_t, -123, 4000, 1234);
+	TEST4_FAILURE("7f", size_t, 0, 1000, EINVAL);
+	TEST4_SUCCESS("0x7f", size_t, 0, 1000, 127);
+	TEST4_SUCCESS("0x77", size_t, 0, 1000, 119);
+	TEST4_SUCCESS("077", size_t, 0, 1000, 63);
+
+	TEST4_SUCCESS("0xffffffff", uint32_t, 0, (uintmax_t)0xffffffff,
+	    UINT32_MAX);
+	TEST4_FAILURE("0xfffffffff", uint32_t, 0, (uintmax_t)0xffffffff,
```
cperciva: It would be good to test the "value is within specified range, but outside of the maximum range of the unsigned type -> ERANGE" case.  So 

`TEST4_FAILURE("0xfffffffff", uint32_t, 0, (uintmax_t)0xfffffffff, ERANGE);`

for example.  (I thought this might have been what you were trying to do with the duplicate line.)

Also, we should probably check capitalization of hex values?

gperciva: ok, done.  I had to play some games with `_Pragma` to avoid tautological comparison warnings, which is why I didn't bother doing that this morning.

cperciva:
Looks good.  Please tidy up the commit history.

gperciva:
Updated PR with rebased history.

PR #59, opened 2017-03-26
Import

gperciva:


gperciva:
These are imported from https://github.com/Tarsnap/spiped/pull/117 https://github.com/Tarsnap/spiped/pull/118

PR #61, opened 2017-03-30
Parsenum double promotion

gperciva:


gperciva:
This is one possible way of avoiding `(double)INFINITY` as the maximum for "a non-zero double".  Note that da689ff looks like a rather questionable cast to me, but seems to be required.

util/parsenum.h
```diff
@@ -61,14 +61,16 @@ _Pragma("clang diagnostic pop")
...
+			((*(x)) = parsenum_float((s), (double)(min),	\
+			    (double)(max))) :				\
 		(((*(x)) = -1) <= 0) ?					\
 			((*(x)) = parsenum_signed((s),			\
 			    (*(x) <= 0) ? (min) : 0,			\
 			    (*(x) <= 0) ? (max) : 0)) :			\
 			(((*(x)) = parsenum_unsigned((s),		\
 			    (min) <= 0 ? 0 : (min),			\
-			    (max) >= *(x) ? *(x) : (max))),		\
+			    (double)(max) >= (double)(*(x)) ?		\
```
cperciva: Did you decide the making this a cast to `uintmax_t` didn't work?

gperciva: See the comment in 13c93b1:

> Since (max) could be INFINITY, we cannot cast this to an int: the C99 standard
says that's undefined behaviour [in 6.3.1.4 Real floating and integer].

gperciva:
Updated and rebased history.

PR #66, opened 2017-08-06
Test getopt

gperciva:


tests/getopt/main.c
```diff
@@ -3,6 +3,14 @@
 
 #include "getopt.h"
 
+/*
+ * The test suite ignores the exit code of this program, so that we can
```
cperciva: Umm, what?  Looks to me like we're checking exit codes...?

gperciva: Kind-of.  If there's an exit code, the test shell script has an early exit, but it's called via:

    @/bin/sh -e test_getopt.sh 2>test_getopt.log || true

https://github.com/Tarsnap/libcperciva/blob/master/tests/getopt/Makefile#L7

The thing that ultimately produces a "FAILED!" from `make test` is:

    @if cmp -s test_getopt.log test_getopt.good; then	\

... but I just noticed that this doesn't actually generate a non-zero exit code.  That seems like it'd be a good idea, particularly with Travis-CI running "make test".

I'll clarify the comment, and get some `exit 1` in there.

gperciva: Just as a sanity check that I wasn't crazy: I changed the final line of `int main()` in `getopt/main.c` to `return (1);`, and still got:

```
$ make test
./test_getopt.sh
Testing command-line options parsing... PASSED!
```

because `test_getopt.log` still matched `test_getopt.good`.

gperciva: Never mind, I see that I misunderstood the effect of `|| true` when running the script within a Makefile, vs. calling a function within a script.

gperciva:
New approach to test_getopt.

tests/getopt/main.c
```diff
@@ -3,6 +3,16 @@
 
 #include "getopt.h"
 
+/*
+ * The test suite does not explicitly check the exit code of this program;
+ * instead, we rely on comparing the output with good output to determine
+ * success or failure.
```
cperciva: This is true, but a bit misleading.  If `test_getopt` fails, `test_getopt_args.sh` will abort early, and so it won't print the final "done".  So unless `test_getopt` somehow prints that by accident, having it exit with the wrong status (whether an incorrect failure or an incorrect success) will result in the test script output not matching and the test being flagged as a failure.

... at least, that's the theory.  Doesn't it work?

gperciva: I agree that's how it works.  I thought I described it acceptably with "does not **explicitly** check the exit code... rely on **comparing** the output", but evidently not.

Fun aside: what do you expect to see here?
```
#!/bin/sh -e
func() {
	echo "begin func"
	false
	echo "after false, ret code: $?"
}
echo "before calling func()"
func || true
echo "after calling func()"
```

cperciva:
```
before calling func()
begin func
after calling func()
```

I think?

gperciva: I'd think so, too.  :)  dash 0.5.8 disagrees.

```
$ sh func-err.sh 
before calling func()
begin func
after false, ret code: 1
after calling func()
$ ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Feb  5  2017 /bin/sh -> dash
$ dpkg -l dash | grep dash
ii  dash           0.5.8-2.1ubuntu2 amd64        POSIX-compliant shell
```

gperciva:
Rebased to remove the comment in `main.c`.  I kept the comment in the shell script.

tests/getopt/main.c
cperciva: Huh, maybe -e doesn't apply inside functions?  News to me...

cperciva: Just confirmed, FreeBSD sh does the same thing.

PR #67, opened 2017-08-23
POSIXFAIL_CLOCK_GETTIME

gperciva:


gperciva:
In response to https://github.com/Tarsnap/tarsnap/issues/286

util/monoclock.c
```diff
@@ -43,6 +44,12 @@ monoclock_get(struct timeval * tv)
 		goto err0;
 	}
 #endif
+#else	/* POSIXFAIL_CLOCK_GETTIME */
```
gperciva: Repeating the `gettimeofday()` lines isn't ideal, but I think it's more readable than doing fancy `#if something && anotherthing`, especially since these variables are long so it would go well over 80 chars.

gperciva:
Yes, the monoclock test uses the current way of writing Makefiles in libcperciva.  Once https://github.com/Tarsnap/spiped/pull/144 is merged, I'll update the monoclock test along with the other libcperciva tests.

POSIX/posix-cflags.sh
```diff
@@ -14,6 +14,11 @@ if ! ${CC} -D_POSIX_C_SOURCE=200809L posix-clock_realtime.c 2>/dev/null; then
 	printf %s "-DPOSIXFAIL_CLOCK_REALTIME"
 	echo "WARNING: POSIX violation: <time.h> not defining CLOCK_REALTIME" 1>&2
 fi
+if ! ${CC} -D_POSIX_C_SOURCE=200809L posix-clock_gettime.c 2>/dev/null; then
```
cperciva: Hmm, isn't this orthogonal to the problem it's supposed to be handling?  This detects if `clock_gettime` is *declared*, but I thought the problem was systems where it wasn't being *exposed in libraries* (and the failure only occurring at run time).

gperciva: Yes, the initial report was about a system which declared `clock_gettime()` but did not provide it for linking.  However, that's after a layer of build system + fixes.

On my native 10.11 machine, `clock_gettime()` is *not* declared:
https://github.com/Tarsnap/tarsnap/issues/286#issuecomment-324072189
and this PR fixes that problem.

The related PR that imports this to Tarsnap (https://github.com/Tarsnap/tarsnap/pull/287) also addresses the case of exposing the function but not providing it for linking:
https://github.com/Tarsnap/tarsnap/pull/287/commits/0af569e18e7cf90562106cb90cd0250fbb29ea8e#diff-67e997bcfdac55191033d57a16d1408aR502
via the `$ac_cv_lib_rt_clock_gettime` variable.

So the only case that's missing is POSIX build with libcperciva monoclock.  I admit that I haven't tried compiling spiped with brew yet, but I'm not certain it needs an extra "function is includable but not linkable" fix because the weirdness that got the system into that situation might only apply to brew + autotools.  I'll test this the next time I boot into OSX.

cperciva: Ok, but your patch doesn't change the behaviour on the 10.11 machine you mention, because CLOCK_REALTIME isn't defined either.  You would only see a change in behaviour if CLOCK_REALTIME was defined in `<time.h>` but clock_gettime wasn't, and AFAIK nobody has ever made a system quite *that* broken.

util/monoclock.c
cperciva: I'd be inclined to rewrite the code by putting some preprocessor logic at the top of the file which sets `USE_MONOTONIC`, `USE_REALTIME`, and `USE_GETTIMEOFDAY` based on whether `CLOCK_MONOTONIC`, `POSIXFAIL_CLOCK_REALTIME`, and `POSIXFAIL_CLOCK_GETTIME` are defined.  I don't like code duplication, and my hunch is that it would be clearer that way.

cperciva:
Something weird happened with the commit which adds the monoclock test.  Wrong email address, I think?

gperciva:
Updated PR with a draft version.
- fixed the email address (I made the commit on the OSX partition and forgot to change the "author" field in addition to the "committer" field)
- redesigned the preprocessor commands in `monoclock.c`
- I still need to test in OSX with all the weird environment variables + Xcode stuff.

POSIX/posix-cflags.sh
```diff
@@ -21,6 +21,23 @@ if ! ${CC} -D_POSIX_C_SOURCE=200809L posix-clock_realtime.c 2>/dev/null; then
 	printf %s "-DPOSIXFAIL_CLOCK_REALTIME"
 	echo "WARNING: POSIX violation: <time.h> not defining CLOCK_REALTIME" 1>&2
 fi
+# We need to run the binary, because OSX 10.11 (and below) uses dynamic linking
+# to a library which doesn't include clock_gettime().
+rm -f posix-clock_gettime
+if ! ${CC} -D_POSIX_C_SOURCE=200809L -o posix-clock_gettime posix-clock_gettime.c 2>/dev/null; then
+	[ ${FIRST} = "NO" ] && printf " "; FIRST=NO
+	printf %s "-DPOSIXFAIL_CLOCK_GETTIME"
+	echo "WARNING: POSIX violation: <time.h> not defining clock_gettime()" 1>&2
```
gperciva: What's the best verb to use here?  "declaring", maybe?

gperciva:
I think this is ready, other than the "redirect" weirdness:

```
Grahams-Mac-mini:POSIX gperciva$ ./posix-clock_gettime
dyld: lazy symbol binding failed: Symbol not found: _clock_gettime
  Referenced from: /Users/gperciva/src/libcperciva/POSIX/./posix-clock_gettime
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _clock_gettime
  Referenced from: /Users/gperciva/src/libcperciva/POSIX/./posix-clock_gettime
  Expected in: /usr/lib/libSystem.B.dylib

Trace/BPT trap: 5
Grahams-Mac-mini:POSIX gperciva$ ./posix-clock_gettime 1>stdout 2>stderr
Trace/BPT trap: 5
Grahams-Mac-mini:POSIX gperciva$ ls -l std*
-rw-r--r--  1 gperciva  staff  350 11 Sep 18:15 stderr
-rw-r--r--  1 gperciva  staff    0 11 Sep 18:15 stdout
Grahams-Mac-mini:POSIX gperciva$ 
```

I'm currently avoiding this by calling a separate shell script, but I just figured out that I can do this instead:
```
Grahams-Mac-mini:POSIX gperciva$ $( ./posix-clock_gettime 2>inner ) 2>outer
Grahams-Mac-mini:POSIX gperciva$ echo $?
133
Grahams-Mac-mini:POSIX gperciva$ ls -l inner outer 
-rw-r--r--  1 gperciva  staff  350 11 Sep 18:17 inner
-rw-r--r--  1 gperciva  staff    0 11 Sep 18:17 outer
Grahams-Mac-mini:POSIX gperciva$ 
```

gperciva:
I think this is ready now (I fixed the "extra shell script" thing).

In order to accommodate the Travis-CI clang 3.9 change, I included #70 in this PR.  I recommend merging #70 first, so that it shows up in the history as a separate merge.

gperciva:
... ok, that was a mess.  I think we'll do a `git revert` in the future.

PR #68, opened 2017-09-01
POSIX: sanity-check posix scripts for ${CC}

gperciva:


gperciva:
Don't know if you'll like this change or not, but I'd find it useful.  :)

POSIX/posix-cflags.sh
```diff
@@ -1,4 +1,10 @@
 # Should be sourced by `command -p sh posix-cflags.sh "$PATH"` from within a Makefile
+# Sanity check env variables
```
cperciva: I'd add a blank line so that it doesn't look like the first two lines are both part of the same comment.

Also, is the comment intended to refer to just the first test (s/env variables/$CC/)?  Or to checking both $CC and $PATH (s/env/environment/, remove the blank line between the two tests, and add a blank line after the second test -- whitespace should reflect the scope of comments)?

gperciva:
Updated PR, including rebase since it's the change is so short.

PR #71, opened 2017-09-16
libcperciva: use mpool for network_*_cookie

gperciva:


gperciva:
This came from https://github.com/Tarsnap/spiped/pull/140

PR #73, opened 2017-09-18
build: use ${CFLAGS*} and ${LDADD*} in buildsingles

gperciva:


gperciva:
I don't think that `${LDADD_POSIX}` matters in this case, but I figure it's good to use it everywhere so I don't forget it elsewhere.

(the important here is passing `${CFLAGS}` to `.inc2` so that we can enable/disable warnings as desired)

PR #74, opened 2017-09-18
Makefiles

gperciva:


gperciva:
As we add more tests (particularly performance tests), the current "manually copy an existing Makefile and edit it a bit" method is sub-optimal.  Let's use the same build system as the other non-autotools Tarsnap projects.

The biggest changes are:
- we can no longer simply cd into the relevant test/*/ directory and run `make` -- we won't have the required ${CFLAGS_POSIX} and ${LDADD_POSIX}.  Instead, we rely on the user running `make` from the top-level directory.  Since this is how all the other Tarsnap software works, I think this is an acceptable change
- I've renamed `${MAKE}` to `${MAKEBSD}` in the top-level `Makefile.BSD`, in order to handle the mess with `tests/buildsingles/`.  (so that I can get the output of `make -n` and replace "make" or "bmake" with "${MAKE}")

(this includes #73 )

gperciva:
Yes, this on the "large" side of PRs.  I could potentially move a few things into separate issues:

- code to import "test:" from `Makefile.BSD` into `Makefile` (applied to spiped, since libcperciva doesn't use the BSD system yet)
- replacing `${MAKE}` with `${MAKEBSD}`

I think the rest belongs together.  I split the changes into
- simple test programs
- not-simple test programs (those with a `Makefile.inc1`)
- icky test program (buidsingles which constructs a `Makefile.inc2`)

Note that as a result of this PR, we can no longer simply `cd` into the relevant `test/*/` directory and run `make` -- we won't have the required `${CFLAGS_POSIX}` and `${LDADD_POSIX}`.  Instead, we rely on the user running `make` from the top-level directory.  Since this is how all the other Tarsnap software works, I think this is an acceptable change.
(Also, a separate change could allow a subdirectory to regenerate the top-level `cpusupport-config.h` and set up `${ _POSIX}` stuff -- but that would be some time in the future.)

Makefile
```diff
@@ -0,0 +1,48 @@
+.POSIX:
+
+PKG=	libcperciva
+PROGS=
+TESTS=	tests/buildall tests/buildsingles tests/crc32 tests/getopt tests/heap \
```
cperciva: Is the `tests/` prefix necessary here?  Can't it just be added in the code under `test:` ?

gperciva: How's that going to work with stuff like

        for D in ${PROGS} ${TESTS}; do                  \
                ( cd $${D} && ${MAKE} all ) || exit 2;  \
        done

?  also, we might add some `perftests/` stuff, like in kivaloo.

cperciva: The easy solution there is to split that into two loops, one for ${PROGS} and one for ${TESTS}.

I agree that it makes the code more complicated, but I think it *might* be better to have the mess in the boilerplate which never gets touched rather than in the list of tests which will hopefully be having new entries added?

Your call.

gperciva: I disagree; I think that being explicit about the directories is a good thing.  We've been explicit everywhere else in the build systems of various projects; I don't see that removing a bit of repetition in `tests/` is worth changing that.

cperciva: Ok, I'll merge this as-is.

PR #75, opened 2017-09-21
POSIXFAIL_CLOCK_GETTIME second try

gperciva:

PR #76, opened 2017-09-21
monoclock: switch to USE_FOO for clock type

gperciva:


gperciva:
I thought it might be helpful to reason about `USE_MONOTONIC` and `USE_REALTIME` if that change were independent from the `POSIXFAIL_CLOCK_GETTIME` stuff in #75.

PR #77, opened 2017-10-05
mpool improvements

cperciva:


cperciva:
@gperciva Can you write some test code for this?  I don't think any of the tests in libcperciva make use of mpool.  Performance testing would be good (maybe compare the time it takes to perform a bunch of mallocs+frees vs. the time it takes using the mpool?) but also making sure that everything is being freed at the end (is it easy to pull the valgrind bits into here?).

gperciva:
I'm getting a segfault with 10000+ allocations when using the autotuning commit, but I'm still tracking it donw.  (it doesn't happen when I run it under `gdb` unless I do `(gdb) set disable-randomization off`)

I just mention this in case you take the silence as "everything is fine" and are thinking about merging it now.  :)

cperciva:
Huh, interesting!

Don't worry, I wasn't planning on this any time soon... and I'm busy until the end of the weekend anyway.

datastruct/mpool.h
```diff
@@ -11,69 +13,114 @@
...
-static struct mpool_##name##_struct {				\
-	size_t stacklen;					\
-	int atexit_set;						\
-	type * allocs[size];					\
-} mpool_##name##_rec = {0, 0};					\
+#define MPOOL(name, type, size)					\
+static void mpool_##name##_atexit(void);			\
+static void * mpool_##name##_static[size];			\
+static struct mpool mpool_##name##_rec =			\
+    {0, size, mpool_##name##_static, 0, 0, 0,			\
```
gperciva: This sets `M->stacklen = 0`.  Shouldn't this be `size`?  Because we created `static void * mpool_##name##_static[size];` ?

datastruct/mpool.h
```diff
@@ -11,69 +13,114 @@
...
+	if (M->allocs != M->allocs_static)
+		free(M->allocs);
+}
+
+static inline void *
+mpool_malloc(struct mpool * M, size_t len)
+{
+
+	M->nallocs++;
+	if (M->stacklen)
```
gperciva: (follow-up to my comment about `M->stacklen`)

... because if `M->stacklen==0`, then I don't think we ever enter this `if`.

datastruct/mpool.h
```diff
@@ -11,69 +13,114 @@
...
+		return (M->allocs[--(M->stacklen)]);
```
gperciva: Also, shouldn't we be returning `&M->allocs[--(M->stacklen)]);` here?  Because the initial contents of `static void * mpool_##name##_static` will be undefined.

datastruct/mpool.h
gperciva: Never mind, after going back to the commit that first introduces the stack, I now see that it shouldn't be non-zero.

datastruct/mpool.h
cperciva: If `stacklen == 0` then we don't have any allocations in the pool.  So we allocate a new one.  But `stacklen` can increase in mpool_free.

datastruct/mpool.h
cperciva: No, because we're only going to return that value after it has been written (and stacklen has been incremented).

Also, M->allocs holds pointers to memory allocations.  Those pointers are the ones we want to return; we don't want to return a pointer to a pointer.

datastruct/mpool.h
cperciva: Right, `stacklen` says how many pointers-to-allocations we have.  We start out with zero such allocations, but space to store some if they arrive via mpool_free.

datastruct/mpool.h
```diff
@@ -11,69 +13,114 @@
...
+	if (M->stacklen < M->allocsize) {
+		M->allocs[M->stacklen++] = p;
+		return;
+	}
+
+	if (M->nempties	> (M->nallocs >> 8)) {
+		allocs_new = malloc(M->allocsize * 2 * sizeof(void *));
+		if (allocs_new) {
+			memcpy(allocs_new, M->allocs,
+			    M->allocsize * 2 * sizeof(void *));
```
gperciva: this looks like undefined behaviour.  We've allocated
    `M->allocsize * sizeof(void *)`
bytes, but you're copying `2*` that value.  Removing the `2*` stops the segfaults I see.  (I haven't yet empirically checked that the mpool content is correct after the adjusted `memcpy`, but I can't see how it wouldn't be)

cperciva: *cough*

Umm, yeah.  When we're expanding a buffer from holding one record to holding two, we need to copy the *one* existing record, not copy two records from the buffer which only has one record in it. :-/

I'm glad I asked you to write some tests for this!

gperciva:
I'd like to tweak the `Change mpool from a linked list to a stack` commit so that it actually compiles, but we can play with that after https://github.com/Tarsnap/libcperciva/pull/80 is in.
(I have the desired changes ready in git, but I'd rather not start doing all the rebasing until we have the test framework in place.)

cperciva:
Note to self, this can be closed once #85 is merged.

PR #78, opened 2017-10-05
Set user group

gperciva:


gperciva:
I expect there's a bunch of nitpicks to happen here (including looking up `setgroups()` for OSX, other BSDs, Cygwin, etc.).  But I wanted to throw it up so you could see how easy it is to add a new test with the new Makefiles system.  :)
(also so that you could take a look at the general API)

tests/setuidgid/Makefile.BSD
```diff
@@ -0,0 +1,22 @@
...
+# Don't install it.
+NOINST	= 	1
+
+# Useful relative directories
+LIBCPERCIVA_DIR	=	../..
+
+# main test code
+SRCS	=	main.c
+
+# Algorithms
```
cperciva: Eh?

util/setuidgid.h
```diff
@@ -0,0 +1,19 @@
+#ifndef SET_UID_GID_H
```
cperciva: `SETUIDGID_H`.  Should match the file name; no need for extra underscores.

util/setuidgid.h
```diff
@@ -0,0 +1,19 @@
+#ifndef SET_UID_GID_H 
+#define SET_UID_GID_H 
+
+enum {
+	LIBCPERCIVA_SGROUP_IGNORE = 0,
```
cperciva: These should all start `SETUIDGID_`.

util/setuidgid.h
```diff
@@ -0,0 +1,19 @@
...
+	LIBCPERCIVA_SGROUP_LEAVE_WARN,
+	LIBCPERCIVA_SGROUP_LEAVE_ERROR
+};
+
+/**
+ * setuidgid(user_group_string, leave_supplementary):
+ * Set the UID and optionally GID to the names given in ${user_group_string}.
+ * If there is a ":" in the string, it is parsed into "username:groupname";
+ * otherwise, the entire string is interpreted as a username.  If
+ * ${leave_supplementary} is non-zero, leave any supplementary groups.
```
cperciva: Doesn't match behaviour.

util/setuidgid.c
```diff
@@ -0,0 +1,366 @@
...
+		goto err0;
+	}
+
+	/* Get group list. */
+	if ((ngroups = getgroups(ngroups_max, grouplist)) < 0) {
+		warnp("getgroups()");
+		goto err0;
+	}
+
+	/*
```
cperciva: `/*-`

(Otherwise the comment can be rewrapped, which you don't want here.)

util/setuidgid.c
```diff
@@ -0,0 +1,366 @@
...
+ */
+static int
+set_user(const char * opt_user)
+{
+	uid_t uid;
+	struct passwd * user_info;
+	uintmax_t parsed;
+
+	/* Calculate numerical UID. */
+	if ((opt_user[0] >= '0') && (opt_user[0] <= '9')) {
```
cperciva: Does POSIX say that you can't have a user name which starts with a digit?  Don't be systemd.

Also, why not use parsenum here?  It will correctly handle `[ug]id_t` regardless of whether they are signed or unsigned.  If parsenum fails, it means the string is not completely numeric, so you can throw it at getpwnam at that point.

util/setuidgid.c
```diff
@@ -0,0 +1,366 @@
...
+
+/* Allocate space for the group list. */
+static int
+allocate_group_list(gid_t **grouplist)
+{
+	long ngroups_max_long;
+	int ngroups_max;
+
+	/* Find system's maximum number of groups.   See [1] about the +1. */
+	ngroups_max_long = sysconf(_SC_NGROUPS_MAX) + 1;
```
cperciva: Better make sure that sysconf didn't return LONG_MAX before you add 1 to it.

util/setuidgid.c
gperciva: Do we actually want to support people specifying a UID directly?  At first I thought we should, but now I'm thinking that it opens a bag of worms.  Let's merely allow people to enter the username.

Apparently on Ubuntu 14.04 at least, an all-digit username of `232` is fine.  And for bonus lols, he even made a username of U+1F600 (😀)!
https://unix.stackexchange.com/questions/287077/why-cant-linux-usernames-begin-with-numbers/287086#287086

gperciva: (the smiley-face username is a POSIX violation, but still awesome.)

The actual POSIX rules allow for digit-only usernames.    The username is composed of characters from the "portable filenames set" (alphanum, plus `.` `-` `_`)  (huh, no other punctuation is deemed to be portable for filenames?).  The only other note given is that the username should not begin with the hyphen character.

gperciva:
New version; needs rebase.

- fixes misc minor things.
- vastly simplifies checking if we've left all supplementary groups.
- stops trying to handle numerica IDs; also makes it much simpler.

gperciva:
Updated PR to rebase onto master (conflicts arose due to the new mpool test).

gperciva:
Updated PR.  Tested on Linux, FreeBSD, OpenBSD, MacOS X.  Also contains untested `#include` statements covering NetBSD, Tru64, Solaris, HP-UX, and AIX.

gperciva:
Example of this in use in https://github.com/Tarsnap/spiped/pull/152.

tests/09-setuidgid.sh
```diff
@@ -0,0 +1,53 @@
...
+	test_setuidgid_username_groupname ":bar" 1
+	test_setuidgid_username_groupname "fake_username" 1
+	test_setuidgid_username_groupname "fake_username:fake_groupname" 1
+
+	# Try as self.
+	test_setuidgid_username_groupname "`id -u -n`:`id -g -n`" 0
+
+	# Try as nobody.
+	test_setuidgid_username_groupname "nobody" 0
+	test_setuidgid_username_groupname "nobody:nogroup" 0
```
gperciva: huh, apparently `nogroup` isn't standard; Redhat-derived distros use "nobody" as the group name.  I guess I should use `id -gn nobody` here.

(I won't change it right now, as I expect there will be plenty of other changes that will need to happen.)

util/setuidgid.c
```diff
@@ -0,0 +1,360 @@
...
+ * There is no setgroups() in the POSIX standard, so if we want to drop
+ * supplementary groups, we need to use platform-specific code.  This must
+ * happen before the regular includes, as in some cases we need to define other
+ * symbols before including the relevant header.
+ */
+#if defined(__linux__)
+/* setgroups() includes for Linux. */
+#define _BSD_SOURCE 1
+#include <grp.h>
+#define HAS_SETGROUPS 1
```
cperciva: Rather than having lots of `#define HAS_SETGROUPS 1` lines, wouldn't it be easier to have a single `#define NO_SETGROUPS` at the end?

util/setuidgid.c
```diff
@@ -0,0 +1,360 @@
...
+static int check_supplementary_groups_none(void);
+
+/* Function prototypes related to uid and gid. */
+static int set_group(const char *);
+static int set_user(const char *);
+static int string_extract_user_group(const char *, char **, char **);
+
+/**
+ * setgroups_none():
+ * Leave all supplementary groups.  If we do not know how to do this on the
```
cperciva: "Attempt to leave supplementary groups. [...]"

util/setuidgid.c
```diff
@@ -0,0 +1,360 @@
...
+			goto nopermission;
+		warnp("Failed to set UID");
+		goto err0;
+	}
+
+	/* Success! */
+	return (0);
+
+nopermission:
+	errno = EPERM;
```
cperciva: Why?

util/setuidgid.c
```diff
@@ -0,0 +1,360 @@
...
+
+	/* Failure! */
+	return (-1);
+
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+/* Parses "username" or "username:groupgroup". */
```
cperciva: We should also accept ":groupname", meaning "change the group but not the user".

Also, s/groupgroup/groupname/ ?

gperciva:
updated PR

tests/09-setuidgid.sh
```diff
@@ -0,0 +1,53 @@
...
+}
+
+### Actual command
+scenario_cmd() {
+	rm -f ${test_stderr}
+
+	# These should all fail.  (Hopefully nobody actually has
+	# "fake_username" or "fake_groupname" on their systems!)
+	test_setuidgid_username_groupname "foo:" 1
+	test_setuidgid_username_groupname ":bar" 1
```
cperciva: This could succeed if there's a "bar" group?

cperciva:
Latest changes look good; please rebase.

tests/09-setuidgid.sh
gperciva: Changed to `:fake_groupname`

gperciva:
Rebased PR, but then added new `REBASE` commits.  Please examine but do not merge.

util/setuidgid.c
```diff
@@ -211,6 +211,13 @@ string_extract_user_group(const char * combined, char ** username_p,
 		warn0("Groupname is blank.");
 		goto err0;
 	}
+	if (pos < len) {
+		/* Check for two (or more) colons. */
+		if (strcspn(combined + pos + 1, ":") < (len - pos - 1)) {
```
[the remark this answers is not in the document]
gperciva: I figured that when it came to setuid stuff, we should err on the side of early warnings.  I'll take it out.

util/setuidgid.c
```diff
@@ -227,16 +227,15 @@ string_extract_user_group(const char * combined, char ** username_p,
...
 			warnp("Failed to allocate memory");
 			goto err0;
 		}
 		strncpy(*username_p, combined, pos);
 		(*username_p)[pos] = '\0';
 
 		/* Extract groupname. */
-		if ((*groupname_p = malloc((len-pos + 1) * sizeof(char)))
-		    == NULL) {
+		if ((*groupname_p = malloc((len-pos + 1))) == NULL) {
```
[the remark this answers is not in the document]
gperciva: done

gperciva:
Another PR update

tests/09-setuidgid.sh
```diff
@@ -0,0 +1,53 @@
...
+	fi
+}
+
+### Actual command
+scenario_cmd() {
+	rm -f ${test_stderr}
+
+	# These should all fail.  (Hopefully nobody actually has
+	# "fake_username" or "fake_groupname" on their systems!)
+	test_setuidgid_username_groupname "foo:" 1
```
cperciva: This is to test that "user name followed by a colon but no group" fails, right?  In that case, how about making it `nobody:` so that it's failing for the right reason?

cperciva:
One minor suggestion, then please rebase -- I think this will be good now.

tests/09-setuidgid.sh
gperciva: well, it errors out from the colon before trying to setuid, so it doesn't matter whether they have a `foo` user or not.  But sure, I've changed that to `nobody:`.

gperciva:
updated PR, ready for merging.

tests/09-setuidgid.sh
```diff
@@ -0,0 +1,54 @@
...
+	cd ${scriptdir}/setuidgid/ && \
+	    ${c_valgrind_cmd} ./test_setuidgid "${username_groupname}"	\
+	    2>> ${test_stderr}
+	cmd_exitcode=$?
+
+	if [ ${cmd_exitcode} -eq "$EX_NOPERM" ]; then
+		# Skip this check, since we can only really test setuidgid
+		# as root (or otherwise have permission to change the username
+		# and group list).
+		echo "-1" > ${c_exitfile}
```
cperciva: Hmm, is this the right test here?  Shouldn't we skip if we're non-root?  I mean, right now we wouldn't treat it as an error if `test_setuidgid` *always* returned EX_NOPERM...?

gperciva: Fair enough.  There's only 1 test that doesn't give `EX_NOPERM`, but I've change it so that it makes sure it doesn't get that for that test.

gperciva:
Updated with different final commit (f857956, which added `tests/09-setuidgid.sh`).

gperciva:
Ok, I still think this version works.  I didn't want to merely check for root, in case somebody wanted to do:

```
sudo setcap cap_setuid,cap_setgid+ep tests/setuidgid/test_setuidgid
make test
```

and see that it succeeded.  BTw, the current version (without `sudo` or permission games) gives:
```
$ make test
...
09-setuidgid... PARTIAL SUCCESS / SKIP!
```

Granted, the `PARTIAL SUCCESS` is all related to the very first test (that `nobody:` fails), as all the other commands need more permission.

tests/09-setuidgid.sh
```diff
@@ -0,0 +1,59 @@
...
+
+	# These should all fail.  (Hopefully nobody actually has
+	# "fake_username" or "fake_groupname" on their systems!)
+	test_setuidgid_username_groupname 1 0 "nobody:"
+	test_setuidgid_username_groupname 1 1 "fake_username"
+	test_setuidgid_username_groupname 1 1 ":fake_groupname"
+	test_setuidgid_username_groupname 1 1 "fake_username:fake_groupname"
+
+	# Try as self.
+	test_setuidgid_username_groupname 0 1 "$(id -u -n):$(id -g -n)"
```
cperciva: Shouldn't this be `0 0`?  IIRC you should always be able to setuid/setgid to yourself?

cperciva:
Hmm.  I agree that it's possible that a non-root user could be given the permission to setuid nobody, but can't we check if we're root and throw an error if we get EX_NOPERM?  I can't imagine anyone running these tests on a system where root *isn't* allowed to setuid nobody...?

util/setuidgid.c
```diff
@@ -0,0 +1,341 @@
...
+	size_t pos;
+	size_t len;
+
+	/* Sanity check. */
+	assert(combined != NULL);
+
+	/* Search for ":". */
+	pos = strcspn(combined, ":");
+
+	/* Extract strings, or duplicate string. */
```
cperciva: This comment seems odd?

util/setuidgid.c
```diff
@@ -0,0 +1,341 @@
...
+		/* Extract username. */
+		if ((*username_p = malloc((pos + 1))) == NULL) {
+			warnp("Failed to allocate memory");
+			goto err0;
+		}
+		strncpy(*username_p, combined, pos);
+		(*username_p)[pos] = '\0';
+
+		/* Extract groupname. */
+		if ((*groupname_p = malloc((len - pos + 1))) == NULL) {
```
cperciva: Why not just `strdup(&combined[pos + 1])` ?

util/setuidgid.c
```diff
@@ -0,0 +1,341 @@
...
+	/* Reject silly strings. */
+	if (pos == len - 1) {
+		warn0("Groupname is blank.");
+		goto err0;
+	}
+
+	/* String ok, proceed. */
+	if (pos == 0) {
+		/* Groupname only; duplicate string. */
+		if ((*groupname_p = strdup(combined + 1)) == NULL)
```
cperciva: I would write `combined + 1` as `&combined[1]`, since it makes the point clearer (to me at least) that we're taking the string starting at offset one in `combined`.

util/setuidgid.c
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	/* Set user. */
+	if (username != NULL) {
+		if (set_user(username) != 0) {
+			/* Quit before warnp() to avoid clearing errno. */
+			if (errno == EPERM)
+				goto err1;
+			warn0("Error changing user");
+			goto err1;
```
cperciva: This idiom (occurring many times) could be written

```
if (errno != EPERM)
        warn0("blah blah blah");
goto err1;
```

Seems a bit cleaner to me?

gperciva:
This is turning out to be a great exercise for "... so you think you know POSIX".  I've gone so far through "frustration" that I'm back at "enjoyment".  :)

Here's 16 (!) new commits.  Do you want to glance at them, or should I just rebase them and you can look at the greatly-changed code?

util/setuidgid.c
```diff
@@ -0,0 +1,329 @@
...
+	if (pos == 0) {
+		/* Groupname only; duplicate string. */
+		if ((*groupname_p = strdup(&combined[1])) == NULL) {
+			warnp("strdup()");
+			goto err0;
+		}
+		*username_p = NULL;
+	} else if (pos != len) {
+		/* Extract username. */
+		if ((*username_p = malloc(pos + 1)) == NULL) {
```
gperciva: It occurred to me that instead of `malloc()`, I could use `strdup()` for everything:

1. make a copy of `const char * combined`
2. if the string looks good, replace `:` with `\0`
3. I can use `strdup()` for both the username and groupname

I think it would end up being more lines of code, so probably not worth it.  But I thought that would be a sneaky way to avoid `malloc`.  :)

cperciva: Yeah, "duplicate string, replace separators with NUL, duplicate individual substrings" is a common idiom -- especially in combination with `strtok`, which does the "replace separators with NUL" part.

I considered mentioning it but in this particular case I don't think it would make the code any clearer.

gperciva:
Yeah, I realized the difference between "not shall fail" and "shall not fail" while I was doing dishes.  I've reinstated that part.

util/setuidgid.c
```diff
@@ -126,7 +126,10 @@ set_group(const char * groupname)
 	/* Find numerical GID. */
 	errno = 0;
 	if ((group_info = getgrnam(groupname)) == NULL) {
-		warnp("getgrnam(\"%s\")", groupname);
+		if (errno)
+			warnp("getgrnam(\"%s\")", groupname);
+		else
+			warn0("No such groupname: %s", groupname);
```
cperciva: I'd say `no such group` (and below, `no such user`) rather than group *name*?

Or at very least add a space before "name".

cperciva:
The changes all look reasonable aside from the setgroups failing and groupname/username vs. group/user vs. group name / user name bit.  Please rebase.

gperciva:
PR updated, including the fix for `getgroups()` potentially failing, and the "No such *foo*name".  I think it's ready for merging.

util/setuidgid.c
```diff
@@ -0,0 +1,332 @@
...
+
+	/* Success! */
+	return (0);
+
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+/* Set the effective GID to the group represented by ${groupname}. */
```
cperciva: s/effective// ?  (And below, for UID.)

util/setuidgid.c
```diff
@@ -0,0 +1,332 @@
...
+	/* Sanity check. */
+	assert(combined != NULL);
+
+	/* Search for ":" and get string length. */
+	pos = strcspn(combined, ":");
+	len = strlen(combined);
+
+	/* Reject silly strings. */
+	if (pos == len - 1) {
+		warn0("Groupname is blank.");
```
cperciva: I'd go with
```
warn0("Empty group name: %s", combined);
```
here (both the spelling of "group name" and printing the failed string).

util/setuidgid.c
```diff
@@ -0,0 +1,332 @@
...
+			goto err0;
+		}
+		*username_p = NULL;
+	} else if (pos != len) {
+		/* Extract username. */
+		if ((*username_p = malloc(pos + 1)) == NULL) {
+			warnp("Failed to allocate memory");
+			goto err0;
+		}
+		strncpy(*username_p, combined, pos);
```
cperciva: Why not just `memcpy`, since we know that we want exactly `pos` bytes?

util/setuidgid.c
```diff
@@ -0,0 +1,332 @@
...
+ *
+ * The behaviour with supplementary groups depends on ${leave_suppgrp}:
+ * - SETUIDGID_SGROUP_IGNORE: do not attempt to leave supplementary groups.
+ * - SETUIDGID_SGROUP_LEAVE_WARN: attempt to leave; if it fails, give a
+ *   warning but continue.
+ * - SETUIDGID_SGROUP_LEAVE_ERROR: attempt to leave; if it fails, return
+ *   an error.
+ */
+int
+setuidgid(const char * opt_usergroup, int leave_suppgrp)
```
cperciva: The parameter names should match those in the comment above.

util/setuidgid.c
```diff
@@ -0,0 +1,332 @@
...
+{
+	char * username;
+	char * groupname;
+	int eperm = 0;
+
+	/* Get the username:groupname from the opt_user string. */
```
cperciva: `opt_user`?

util/setuidgid.c
```diff
@@ -0,0 +1,332 @@
...
+	if (string_extract_user_group(opt_usergroup, &username, &groupname))
+		goto err0;
+
+	/* If requested, leave supplementary groups. */
+	if (leave_suppgrp != SETUIDGID_SGROUP_IGNORE) {
+		/* Leave any supplementary groups. */
```
cperciva: s/Leave/Attempt to leave/ ?

util/setuidgid.c
```diff
@@ -0,0 +1,332 @@
...
+	/* Clean up. */
+	free(groupname);
+	free(username);
+
+	/* Success! */
+	return (0);
+
+err1:
+	/* POSIX Issue 7 does not forbid free() from modifying errno. */
+	if (errno == EPERM)
```
cperciva: `saved_errno = errno;` is more idiomatic... is there a reason we only want to save EPERM and don't want to preserve any other error code from `free`'s intervention?

cperciva:
Just some very minor nits.

util/setuidgid.c
gperciva: oh, also: every error path other than `EPERM` should have already received a `warnp()` or `warn0()`.
(I'll double-check this tomorrow)

gperciva:
Added `REBASE` commits for the latest nits.

cperciva:
Looks good, please rebase.

gperciva:
Updated PR, ready for merge.

PR #79, opened 2017-10-09
Posix updates

gperciva:


POSIX/README
```diff
@@ -2,9 +2,23 @@ POSIX compatibility code
...
+- posix-cflags.sh: Detects if ${CC} supports certain POSIX features, and
+  outputs a POSIXFAIL_ define if it is not supported so that we can work around
+  the problem.
+
+  The potential command-line flags are:
+  - DPOSIXFAIL_MSG_NOSIGNAL: <sys/socket.h> not defining MSG_NOSIGNAL.
+  - DPOSIXFAIL_CLOCK_REALTIME: <time.h> not defining CLOCK_REALTIME.
+  - DPOSIXFAIL_CLOCK_GETTIME: <time.h> not declaring clock_gettime(), or
+    clock_gettime() is not linkable.
+  - std=c99: ${CC} does not accept the `restrict` keyword.
```
cperciva: ${CC} does not accept the C99-mandated `restrict` keyword by default, but does so when this flag "reminds" it to behave like a C99 compiler.

Or maybe something else, but the point is that we only emit -std=c99 if it fixes the `restrict` handling.

POSIX/README
```diff
@@ -2,9 +2,23 @@ POSIX compatibility code
 ------------------------
 
 This code exists to work around some common POSIX compatibility issues.
-
 POSIX specifies that if the first line of a Makefile is ".POSIX:" then the
 Makefile should be processed according to POSIX rules, including with CC=c99.
-Further, c99 is required to understand the -lrt and -lxnet options (and ignore
-them if the routines they specify linkage for are already in the standard C
-library).  Unfortunately some systems fail or one or both of these accounts.
+
```
cperciva: Maybe `s/./; thus we should see behaviour consistent with a standard-compliant C99 compiler/` here, to make the point of why the value of ${CC} matters?

gperciva:
updated PR with change to rebase.

POSIX/README
gperciva: hmm, ok.  The original wordings came directly from the warnings given by `posix-cflags.sh`.  (I spent a few minutes trying to think of new phrasings for them, before I decided the original warnings should be clear enough for the `README`.)

cperciva: Yeah, and in most cases that's fine... this particular one is a special case though since the warning given at build time (hey, your system is broken) doesn't entirely explain the circumstances when the option is emitted.

cperciva:
Looks good; please rebase.

gperciva:
Rebased, ready for merge.

PR #80, opened 2017-10-16
Mpool pre

gperciva:


gperciva:
I somewhat ran of out steam when it came to the timing data; as long as it generates some kind of benchmark data, it accepts that the test passed.  I think that comparing one run to another would be a whole other test framework, quite likely not involving Travis-CI (since timing will depend on the type of system it runs on, and we need to store data between runs).

I also included the comment-reword from the `mpool` branch, because that doesn't depend on the mpool-to-stack commits.

gperciva:
This is not a candidate for merging; I just put it online so you could take a look at it if you wanted to.  (after receiving an upcoming email about timing reliability)

tests/mpool/test_mpool.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# Test with full benefit from malloc pool (the binary begins with an
+# initial pool of 100).
+cmp_methods 10000 100 200
+cmp_methods 10000 100 200
+cmp_methods 10000 100 200
+
+# mpool is not much slower than malloc in this range
+# (it's generally faster, but we can't reliably test for that).
+cmp_methods 1000 1000 90
```
gperciva: Right now these are pretty useless, but after the auto-tuning change, we can bump the min_percent up to 300 or so.

.travis.yml
```diff
@@ -7,7 +7,7 @@ matrix:
     - compiler: gcc
       env: CFLAGS="-std=c99 -O2 -Wall -Wextra -Werror -Wpedantic -pedantic-errors -Wno-clobbered"
     - compiler: clang
-      env: CFLAGS="-std=c99 -O2 -Wall -Wextra -Werror -Weverything -Wno-#warnings -Wno-pedantic -Wno-padded -Wno-format-nonliteral -Wno-disabled-macro-expansion -Wno-missing-noreturn -Wno-reserved-id-macro"
+      env: CFLAGS="-std=c99 -O2 -Wall -Wextra -Werror -Weverything -Wno-#warnings -Wno-pedantic -Wno-padded -Wno-format-nonliteral -Wno-disabled-macro-expansion -Wno-missing-noreturn -Wno-reserved-id-macro -Wno-cast-align"
```
gperciva: just noticed that this change isn't necessary after the mpool change.  Should I make this branch mpool-post intead of pre?

gperciva:
(I removed some misleading github comments of mine which applied to an earlier version)

If this is merged before #81, I can update #81 to include these tests.  I'm not aware of any problems with this PR, other than the below.

Before this PR, `make test` takes 0.3 seconds on my desktop.  After the PR, it's almost 3 seconds.  I orginally ran each `cmp_methods` function 3 times to counteract cpu scaling issues, but that of course made it slower.  Also, I'm currently seeing about 5% failure rate on my desktop and freebsd inside virtualbox, so perhaps I should increase the number of iterations a bit more?

tests/mpool/main.c
```diff
@@ -0,0 +1,154 @@
+#include <sys/resource.h>
```
cperciva: Is this left over from an attempt which used getrusage?

tests/mpool/main.c
```diff
@@ -0,0 +1,154 @@
...
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
```
cperciva: Not sure what needs this, either?

tests/mpool/test_mpool.sh
cperciva: Even after we bring in autotuning, we should have a test that mpool isn't much slower than malloc for cases where mpool isn't useful (i.e., sets = 1).

.travis.yml
cperciva: Can you put this flag into the mpool test Makefile instead of being in here?  (And then take it out after the mpool changes go in?)

gperciva:
Updated.  Rebased onto master, to get the "better indirect tests" improvements, but needs another rebase because I left the fixes as separate commits.

gperciva:
Rebased to clean up history, and fixed a few trivial things.  (for example, the previous version still had `(void)argv; /* UNUSED */` even though it was actually used)

PR #81, opened 2017-10-28
Test framework with valgrind

gperciva:

PR #82, opened 2017-10-31
Tests better indirect

gperciva:


gperciva:
Sorry, I should have caught this mistake before #81 went in.

PR #84, opened 2017-11-12
Mpool visualization of timing

gperciva:


gperciva:
Here's an example of the plots it makes.  You might enjoy downloading them into a directory and using an image program to quickly flip between them, so you can see the benefit of each step.

Note that autotuning significantly slows down the memory management in the worst-case scenario.

## Pi 3, cpufreq disabled

# Original mpool
0-pre txt

# mpool with stack
1-stack txt

# mpool with stack autotuning
2-autotune txt

gperciva:
For comparison, my desktop (with normal cpufreq and music playing) produces this for autotuning:

2-autotune txt

PR #85, opened 2017-11-12
Mpool adjusted

gperciva:


gperciva:
Here's my proposed tweaks to the mpool autotuning PR.  I think it can go ahead indepdently of the box-plot stuff.

... wow, this is annoying.  Github displays these commits in chronological order, instead of the actual order in the branch.  So `REBASE PREV` and `REBASE NEXT` aren't very descriptive.  :/

Basically, I'm just doing

```
-} mpool_##name##_rec = {0, 0};					\
+} mpool_##name##_rec = {0, 0, {NULL}};				\
```
for the stack version of mpool (and then reversing that change before the autotune commit).  If that looks ok, I can handle the rebasing.

gperciva:
Oh yeah: the "potential slowdown" I was concerned about with this change was because I'd compiled it with clang instead of gcc.  With the same compiler, there's no penalty to explicitly initializing it with `{NULL}`.

cperciva:
Ok, this looks reasonable to me... can you rebase it?  First two commits (list->stack + your fix) into one commit, last three (your revert + autotuning + memcpy fix) as another?

gperciva:
Visualization of the changes (still working on finalizing the plots):

### original
0-pre txt

### stack
1-stack txt

### stack + autotune
2-autotune txt

PR #87, opened 2017-12-06
tests: no #pragma for gcc < 4.6

gperciva:


gperciva:
In particular, OpenBSD 6.1's cc is gcc 4.2.1 (?!?), which cannot handle `#pragma` inside a function.

gperciva:
Better solution, including fixing the build.

tests/parsenum/main.c
```diff
@@ -102,7 +102,7 @@ main(int argc, char * argv[])
 
 #if __clang__
 	_Pragma("clang diagnostic ignored \"-Wtautological-constant-out-of-range-compare\"");
-#elif __GNUC__
+#elif __GNUC__ >= 4 && __GNUC_MINOR__ >= 6
```
cperciva: So, err, we don't want to include this code for GCC 5.1?

gperciva:
Yikes, how embarrassing!  Updated PR.

PR #88, opened 2017-12-06
Fix posix seq

gperciva:


gperciva:
Apparently FreeBSD also has the non-POSIX `seq`.  OpenBSD 6.1 doesn't.

tests/mpool/test_mpool.sh
```diff
@@ -6,19 +6,29 @@ OUT=timing.txt
 # Header
 echo "sets\treps\tmalloc\tmpool" > ${OUT}
 
+make_count() {
+	END=$1
+	for i in `awk -v MYEND=$END 'BEGIN { for(i=1;i<=MYEND;i++) print i }'`
```
cperciva: This seems a bit baroque... why not do arithmetic in the shell?

````
END=$1
N=1
while [ $N -le $END ]; do
    echo $N
    N=$((N + 1))
done
```

I think?

gperciva:
Updated PR.

PR #89, opened 2017-12-09
Fix build on MacOS X due to struct timespec

gperciva:


gperciva:
This should solve the build failure in https://github.com/Tarsnap/spiped/pull/152.

PR #90, opened 2017-12-14
Travis-CI: add -Wno-cast-align for clang-5.0

gperciva:


gperciva:
They jumped from clang-3.9 to 5.0 on the trusty (ubuntu 14.04) images.  I wasn't expecting this, since apparently xenial (ubuntu 16.04) are coming "soon".

This is slightly awkward timing for us, since we're still adjusting to the new OSX images... but oh well.

PR #91, opened 2018-01-17
Coverity scan assert

gperciva:


gperciva:
This gets coverity-scan working again, and fixes issues it complained about.

I wish I didn't have to add 7256271, but that was the only way I could find to avoid a high-priority "freeing previously freed memory" error.  I'm 99% certain that the actual C code is good.

I'm also a bit uncertain about `ASSERT_MSG` in 1eaeeba.

gperciva:
I rebased this on top of the mpool changes, and now coverity's unhappy again.  Needs more work.

gperciva:
updated PR.  I'm fairly confident about these small changes to make Coverity happier.

(there's still some remaining Coverity issues, but unrelated to these changes)

util/parsenum.h
```diff
@@ -53,7 +61,7 @@ _Pragma("clang diagnostic pop")
 		(((*(x)) = -1) > 0) ?					\
 			((*(x)) = parsenum_unsigned((s), 0, (*(x)),	\
 			    (*(x)))) :					\
-			(assert(0 && "PARSENUM applied to signed integer without specified bounds"), 1),	\
+			(ASSERT_MSG(0, "PARSENUM applied to signed integer without specified bounds"), 1),	\
```
cperciva: You checked that `assert(("foo", 0))` doesn't work?  (Note double parentheses, since this is all part of the first (and only) argument to `assert`.)

gperciva: Yes, that ran into "expression result unused", related to the string.

util/parsenum.h
```diff
@@ -29,6 +30,13 @@ _Pragma("clang diagnostic pop")
 #define PARSENUM_EPILOGUE /* NOTHING */
 #endif /* !__clang__ */
 
+/* Print a message before assert(). */
+#define ASSERT_MSG(x, s)						\
+	(								\
+		printf("ASSERT: " s "\n"),				\
```
cperciva: Bug #1: This ought to go to stderr, not stdout.
Bug #2: This should only get printed if x is zero.  Which it is in this case, of course... maybe drop that parameter and call this macro ASSERT_FAIL ?

While we're here, can you printf all of the assertion output and then call abort(), rather than printing a string ourselves and having assert() print more?

cperciva:
I think blue is a good colour for bike sheds.

cperciva:
(Aside from changing the assertion output format, I think this is good -- can you rebase it so I can merge?)

gperciva:
Updated PR, ready for merge.

PR #92, opened 2018-01-21
Coverity scan easy

gperciva:


gperciva:
There's a number of Coverity-scan fixes coming up.  This PR has the changes I'm most confident about, as well as the basic Travis-CI fixes (it would be useful to have that in master, so other changes have a coverity-buildable "base" to work from).

PR #94, opened 2018-02-04
Limit # of events processed at once

gperciva:


gperciva:
Without this, during busy periods a server could loop for extended periods without ever exiting this loop; as a result, it would never handle signals which had been marked as received.

This is a somewhat messy way of doing it.  I'm mainly putting it forward so that it won't get forgotten.

cperciva:
Going to leave this here for now to remind me that I need to do something about it; but I don't think this is the solution I want.

Maybe a "interrupt the event loop" function which signal handlers can call?

PR #99, opened 2018-04-25
FIXME ridiculous fix for coverity-scan

gperciva:


gperciva:
Not for merging.

In the normal code, Coverty-scan complains that there's a double free concerning `arr[0]`.  I've managed to avoid this by explicitly allocating `arr[0]` outside of a `for` loop, but this is ridiculous.  Any ideas?

Link to the issue for @cperciva: https://scan9.coverity.com/reports.htm#v26532/p10870/fileInstanceId=33776533&defectInstanceId=5427546&mergedDefectId=181734
(or rather, to a slightly modified version: I tried putting the `mpool_` stuff separately from the `malloc` stuff.)

gperciva:
Exact message in context:
```
 75                /*
 76                 * Free one structure to potentially trigger mpool growth,
 77                 * then test memory access.
 78                 */
   	13. freed_arg: free frees *arr.
   	
CID 181734 (#1 of 1): Double free (USE_AFTER_FREE)
35. double_free: Calling free frees pointer *arr which has already been freed.
 79                free(arr[0]);
```

and yes, I tried adding `assert(arr[0] != NULL);` above the `free(arr[0]);` line, but that didn't fix it.

cperciva:
I'm somewhat curious to see how much this test case can be minimized while still confusing Coverity... but I think this is just something we'll have to mark in Coverity as a false positive.  There are limits to how far we can go to make the code easier for Coverity to understand...

gperciva:
I updated this, showing one function which are AFAIK the same, but Coverity-scan complains about one of them.  It seems that ```for (i = 0; i < 1; i++)``` messes up its alloc/free tracking?

Issue #100, opened 2018-04-27
allow numeric IDs in setuid


gperciva:
There's a desire to support numeric values for usernames and groupnames in setuid.  I'm a bit stuck on implementing this.

Current problem: how can I validate that a digit-only argument is a potentially valid username (or group name)?

- there's no `UID_T_MAX`, unfortunately.  :(

- At first I thought that I could use `strtoumax()`, and ensure that the resulting value was not larger than `(uid_t)-1`.  However, POSIX specifies that `uid_t` and `gid_t` are integer types.  No mention of signed or unsigned.
  http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html
  HP-UX for example uses uid -2 for the "nobody" user.

- for reference, FreeBSD assumes that `uid_t` is `unsigned long`... which I suppose is fair enough since they control the whole stack.
 https://github.com/freebsd/freebsd/blob/18951eee499f30623500f670bc0396817ba3b067/usr.sbin/chown/chown.c#L252

I suppose that I could just parse the values with `strtoimax()`, force-cast them as `uid_t`, and hope for the best?  I mean, if a user has the ability to `su` and wants to be malicious, there's far easier ways than trying to setuid to numeric user 4294967297?  That feels like a cop-out, though.

cperciva:
In the documentation for `<sys/types.h>` (http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html), POSIX says that `uid_t` and `gid_t` "shall be integer types", but in the general information (http://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xsh_chap02.html), POSIX says that there is "no structure imposed on uid_t and gid_t other than that they be positive arithmetic types".

Putting these together, I think we can assume that they are positive integer types, so PARSENUM should do what we need here.  (And the use of `(uid_t)(-1)` and `(uid_t)(-2)` to mean "error" and the "nobody" user respectively doesn't matter -- unsigned arithmetic wraps, that's all.)

PR #101, opened 2018-04-28
Setuid numeric

gperciva:


gperciva:
addresses #100.

util/setuidgid.c
```diff
@@ -133,14 +134,20 @@ set_group(const char * groupname)
...
+	} else {
+		if (errno) {
 			warnp("getgrnam(\"%s\")", groupname);
-		else
-			warn0("No such group: %s", groupname);
-		goto err0;
+			goto err0;
+		} else {
+			/* Treat it as a numeric ID. */
+			if (PARSENUM(&gid, groupname)) {
```
cperciva: Why the deep nesting?  It's perfectly valid to write

```
if (foo) {
    ...
} else if (bar) {
    ...
} else if (baz) {
    ...
}
```

all at one level of indentation.

gperciva: Biggest reason was the comment.  Both of these look wrong:
```
                if (errno) {
                        warnp("getgrnam(\"%s\")", groupname);
                        goto err0;
                } else if (PARSENUM(&gid, groupname)) {
                        /* Treat it as a numeric ID. */
                        warn0("No such group: %s", groupname);
                        goto err0;
                }
```
The above looks wrong because the "treat it as a numerid ID" comes **after** the attempt to do so.

Alternatively,
```
                if (errno) {
                        warnp("getgrnam(\"%s\")", groupname);
                        goto err0;
                        /* Treat it as a numeric ID. */
                } else if (PARSENUM(&gid, groupname)) {
                        warn0("No such group: %s", groupname);
                        goto err0;
                }
```
the above looks wrong because the "treat it as a numeric ID" is part of the code block related to an `errno` from `getgrnam()`.

Maybe it's "obvious" what the code is doing, and you don't think we need a comment?  If you explicitly say so, I'll go ahead and remove the comment, but I think it's useful.

gperciva:
Updated with rebased history -- the new format let me move most of the changes to the "refactoring" commit, minimizing the "parsenum" commit.

Issue #42, opened 2016-11-07
CPU feature: RDRAND


cperciva:
We should detect RDRAND on x86 systems and use it as an additional source of entropy in crypto_entropy's reseed function if available.

PR #103, opened 2018-04-30
add RDRAND as entropy source

gperciva:


gperciva:
Addresses #42.

I'm not happy with the amount of testing I've done -- other than "it compiles and runs ok under the Intel software emulator", I'm not certain how to go about looking for problems in a random number generator.

gperciva:
I'll revisit this once #92 is accepted (in particular, after seeing if there's a better way to do the casting from `uint8_t * buf` to an `int` or `long`).

cpusupport/Build/cpusupport.sh
```diff
@@ -46,3 +46,4 @@ feature() {
 feature X86 CPUID ""
 feature X86 SSE2 "" "-msse2" "-msse2 -Wno-cast-align"
 feature X86 AESNI "" "-maes" "-maes -Wno-cast-align" "-maes -Wno-missing-prototypes -Wno-cast-qual"
+feature X86 RDRAND "" "-mrdrnd" "-mrdrnd"
```
cperciva: Why is `-mrdrnd` listed twice here?

tests/crypto_entropy/main.c
```diff
@@ -0,0 +1,43 @@
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
```
cperciva: I don't think all of these are needed?

tests/crypto_entropy/main.c
```diff
@@ -0,0 +1,43 @@
...
+	(void)argv;
+
+	/* Get entropy enough times to trigger a reseed. */
+	for (i = 0; i < RESEED_INTERVAL + 1; i++) {
+		if (crypto_entropy_read(buf, 32))
+			goto err0;
+	}
+
+	/* Print final random value. */
+	for (i = 0; i < 32; i++) {
```
cperciva: Shouldn't we be printing a pass/fail result?  Not that we have any way to recognize if random numbers are random, but at least we know if there was an internal error.

crypto/crypto_entropy_rdrand.c
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * generate_seed_rdrand(seed_material, len):
+ * Fill the ${seed_material} buffer with values from RDRAND.  The length must
+ * be divisible by 4.  This implementation uses the RDRAND instruction, and
+ * should only be used if CPUSUPPORT_X86_RDRAND is defined and
+ * cpusupport_x86_rdrand() returns nonzero.
+ */
+int
+generate_seed_rdrand(uint8_t * seed_material, size_t len)
```
cperciva: I would replace `uint8_t * seed_material` with `int * buf`.

crypto/crypto_entropy.c
```diff
@@ -58,6 +60,15 @@ instantiate(void)
 	/* Mix the random seed into the state. */
 	update(seed_material, 48);
 
+#ifdef CPUSUPPORT_X86_RDRAND
+	/* Add output of RDRAND into the state. */
+	if (cpusupport_x86_rdrand()) {
+		if (generate_seed_rdrand(seed_material, 48))
+			return (-1);
+		update(seed_material, 48);
```
cperciva: There's no need for this to be 48 bytes.  The initial seeding is 48 bytes, but that's 48 bytes of "guaranteed" entropy, whereas RDRAND only gives us "probably" entropy anyway.

If we make this 32 bytes you can take this and the code you added to reseed() and make them a single `update_from_rdrand` function which has an internal `int buf[8]`.

(This also avoids the alignment issues -- `seed_material` is not guaranteed to be aligned properly for converting into `int *`.)

crypto/crypto_entropy.c
```diff
@@ -112,6 +124,18 @@ update(uint8_t * data, size_t datalen)
 	insecure_memzero(Vx, 33);
 }
 
+#ifdef CPUSUPPORT_X86_RDRAND
+static void
+update_from_rdrand() {
+	unsigned int buf[8];
```
cperciva: Need to zero this buffer before exiting.

crypto/crypto_entropy_rdrand.c
```diff
@@ -0,0 +1,36 @@
+#include "cpusupport.h"
+#ifdef CPUSUPPORT_X86_RDRAND
+
+#include <assert.h>
```
cperciva: What are these used for?

crypto/crypto_entropy_rdrand.h
```diff
@@ -0,0 +1,15 @@
+#ifndef _CRYPTO_ENTROPY_RDRAND_H
+#define _CRYPTO_ENTROPY_RDRAND_H
+
+#include <stddef.h>
+#include <stdint.h>
```
cperciva: Why?

crypto/crypto_entropy.c
```diff
@@ -112,6 +124,18 @@ update(uint8_t * data, size_t datalen)
...
+
+	/* This is only *extra* entropy, so it's ok if it fails. */
+	if (generate_seed_rdrand(buf, 8))
+		return;
+	update((uint8_t *)buf, 32);
```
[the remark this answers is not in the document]
gperciva: right, done.  I was confused by:

> `int _rdrand32_step (unsigned int* val)`
> https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_rdrand32_step&expand=4250

where Intel doubles down on "C variables have a specific number of bits".  *La sighe*.

gperciva:
Updated PR, probably ready for rebasing.

cperciva:
Looks good, please tidy up the history (and adjust `Makefile` for the change in the list of tests).

gperciva:
Updated PR, ready for merge

crypto/crypto_entropy.c
```diff
@@ -37,6 +39,21 @@ static void update(uint8_t *, size_t);
 static int reseed(void);
 static void generate(uint8_t *, size_t);
 
+#ifdef CPUSUPPORT_X86_RDRAND
+static void
+update_from_rdrand() {
```
cperciva: `void` should go in here.  Functions without parameter lists is an obsolescent feature of the C language.

gperciva:
Updated PR, including the rebased fix of adding `(void)`.  I also fixed this in a number of other places, in a separate PR.

PR #105, opened 2018-05-01
tests: dump failing mpool info to stderr

gperciva:


gperciva:
We're seeing a bunch of failures in Travis-CI, but I'm not certain which value is failing.

gperciva:
Looks like it's the worst case.  I dropped that from 80 to 75.

PR #106, opened 2018-05-01
cpusupport: test exactly how AESNI will be used

gperciva:


gperciva:
It always bugged me that we had to manually add `-Wno-cast-align`.  If it's truly required to use the intrinsics (as it appears to be), it should be detected as part of `cpusupport-config.h`.

cpusupport/Build/cpusupport-X86-AESNI.c
```diff
@@ -1,14 +1,16 @@
...
-static char a[16];
+static uint8_t a[16];
 
 int
 main(void)
 {
 	__m128i x, y;
 
-	x = _mm_loadu_si128((__m128i *)a);
+	x = _mm_loadu_si128((const __m128i *)&a[0]);
```
gperciva: This is a better match for what happens in `crypto/crypto_aes_aesni.c`.

PR #96, opened 2018-04-19
CRC32 with SSE4.2

gperciva:

PR #107, opened 2018-05-01
Crc32 sse4.2

gperciva:


gperciva:
FFS, I forgot to reopen before pushing **again**.  :(

Continuation of #96 due to github's stupid UX.  Includes #106 (better cpusupport flag detection).  Needs rebase before merging.

cpusupport/Build/cpusupport.sh
```diff
@@ -45,5 +45,5 @@ feature() {
 
 feature X86 CPUID ""
 feature X86 SSE2 "" "-msse2" "-msse2 -Wno-cast-align"
-feature X86 CRC32 "" "-msse4.2" "-msse4.2 -Wno-cast-align"
+feature X86 CRC32 "" "-msse4.2" "-msse4.2 -Wno-cast-align" "-msse4.2 -Wno-cast-align -fno-strict-aliasing"
```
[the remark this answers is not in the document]
gperciva: (err, given that `uint8_t buf[8]`)

alg/crc32c.c
```diff
@@ -74,6 +79,60 @@ init(void)
 	assert(T0[0x80] == T_0_0x80);
 }
 
+#ifdef CPUSUPPORT_X86_CRC32
+/* Test whether CRC extensions and software code produce the same results. */
+static int
+crctest(const uint8_t buf[4])
```
cperciva: I'm not convinced that testing *4 bytes* is really helpful.  I mean, this test would pass even if `crc32q` isn't implemented.

How about taking the CRC of { 0x11, 0x22, 0x33, 0x44 ... 0xff }?

alg/crc32c.c
```diff
@@ -74,6 +79,60 @@ init(void)
...
+		/* Default to software. */
+		crcgood = 0;
+
+		/* If the CPU doesn't claim to support SSE4.2, stop here. */
+		if (!cpusupport_x86_crc32())
+			break;
+
+		/* Test case: "a". */
+		memset(block, 0, 4);
+		block[0] = 'a';
```
cperciva: Why construct this here?  Just create a static constant array?  It will take less space in `.data` than you would spend in `.text` to construct it...

alg/crc32c.c
```diff
@@ -74,6 +79,60 @@ init(void)
...
+		    T0[((state_software >> 24) & 0xff) ^ buf[3]] ^
+		    T1[((state_software >> 16) & 0xff) ^ buf[2]] ^
+		    T2[((state_software >> 8)  & 0xff) ^ buf[1]] ^
+		    T3[((state_software)       & 0xff) ^ buf[0]];
+
+	/* Test hardware transform function. */
+	CRC32C_Update_SSE42(&state_hardware, buf, 4);
+
+	/* Do the outputs match? */
+	return ((state_software == state_hardware) ? 0 : 1);
```
cperciva: Why compare the two values?  We know what the answer should be...

cperciva: I suggest looking at `selectsmix` in the scrypt code, BTW.

alg/crc32c_sse42.c
```diff
@@ -0,0 +1,38 @@
...
+ * CRC32C_Update_SSE42(state, buf, len):
+ * Feed ${len} bytes from the buffer ${buf} into the CRC32C whose state is
+ * ${state}.  This implentation uses x86 SSE4.2 instructions, and should only
+ * be used if CPUSUPPORT_X86_CRC32 is defined and cpusupport_x86_crc32 returns
+ * nonzero.
+ */
+void
+CRC32C_Update_SSE42(uint32_t * state, const uint8_t * buf, size_t len)
+{
+	const size_t pre = ((uintptr_t)buf & 7) < len ? (uintptr_t)buf & 7 : len;
```
cperciva: If `buf` is `0x10003` and `len` is `16`, what will `pre` be?

alg/crc32c_sse42.c
```diff
@@ -0,0 +1,38 @@
...
+	const size_t remaining = len - pre;
+	const size_t block_len = remaining - remaining % 8;
+	size_t i = 0;
+
+	/* Process bytes before the alignment. */
+	for (; i < pre; i++)
+		*state = _mm_crc32_u8(*state, buf[i]);
+
+	/* Process blocks of 8. */
+	for (; i < block_len; i += 8) {
```
cperciva: ... and what will `buf[i]` be when we enter this loop?

cpusupport/Build/cpusupport-X86-CRC32.c
```diff
@@ -0,0 +1,17 @@
...
+main(void)
+{
+	unsigned int state = 0;
+	unsigned char x = 0;
+	uint8_t buf[8] = {0, 0, 0, 0, 0, 0, 0, 0};
+
+	/* Test both the 8-bit and 64-bit data versions. */
+	state = _mm_crc32_u8(state, x);
+
+	_mm_crc32_u64(state, *(unsigned long *)&buf[0]);
```
cperciva: Maybe worth adding `state =` here and making the next line `return (state);` in order to avoid dead-code warnings?

tests/crc32/main.c
```diff
@@ -23,6 +25,83 @@ static struct testcase {
 	    {0x1b, 0xc4, 0xb4, 0x28}}
 };
 
+/* Largest buffer must be first. */
+static const size_t perfsizes[] = {16384, 8192, 4096, 2048, 1024, 512, 256,
+    128, 64, 32, 16};
+static const size_t num_perf = sizeof(perfsizes) / sizeof(perfsizes[0]);
+static const size_t bytes_to_hash = 1 << 29;	/* approx 500 Mb */
```
cperciva: `MB`

tests/crc32/main.c
```diff
@@ -23,6 +25,83 @@ static struct testcase {
...
+
+	/* Allocate buffer to hold largest message. */
+	if ((largebuf = malloc(perfsizes[0])) == NULL) {
+		warnp("malloc");
+		goto err0;
+	}
+	memset(largebuf, 0, perfsizes[0]);
+
+	/* Inform user. */
+	printf("Hashing %.03g bytes...\n", (double)bytes_to_hash);
```
cperciva: Why not just %zu?

tests/crc32/main.c
```diff
@@ -23,6 +25,83 @@ static struct testcase {
...
+			warnp("monoclock_get()");
+			goto err1;
+		}
+
+		/* Find and print elasped time. */
+		delta = 1000000*((long long)(end.tv_sec - begin.tv_sec)) +
+		    (end.tv_usec - begin.tv_usec);
+		printf("... in %zu blocks of size %zu:\t%.02f s\n",
+		    num_hashes, perfsizes[i], (double)delta / 1000000);
+	}
```
cperciva: Maybe worth adding a value in MB/s?

alg/crc32c_sse42.c
gperciva: Well, `0x10003 & 7` is `3`, and `3 < 16`, so `pre=3` and `block_len = 8`.  So when we get here, `i=3`, and `buf[i] = 0x10006` ?  I'm not seeing the problem.

I've rewritten the logic about this, though.  There's no point playing games with `const` vars at the expense of code clarify, especially since this function is only called once per buffer.  If the new code is wrong, at least it should be easier to see why it's wrong.

gperciva:
Updated PR, maybe worth rebasing now?  (not necessarily for immediate merging, but to clarify what's happening)

In case the comment got buried, I didn't understand the problem with calculating pre-aligned-block-bytes in `crc32c_sse42.c`.  I rewrote that code with a view toward more clarity, so hopefully the problem will be more obvious now.

alg/crc32c_sse42.c
```diff
@@ -15,17 +15,29 @@
...
+	size_t pre_block;
+	size_t remaining_bytes;
+	size_t in_block;
 	size_t i = 0;
 
+	/*
+	 * Calculate how many bytes are before an aligned block.  Assume that
+	 * the (uintptr_t) value indicates the memory layout.
+	 */
+	pre_block = (uintptr_t)buf & 7;
```
cperciva: If `buf` is `0x10001`, what is `pre_block`?

And how many bytes do you need to process before you reach a multiple of `0x8`?

tests/crc32/main.c
```diff
@@ -82,11 +83,13 @@ perftest()
...
-		delta = 1000000*((long long)(end.tv_sec - begin.tv_sec)) +
+		/* Find and print elasped time and speed. */
+		delta_us = 1000000*((long long)(end.tv_sec - begin.tv_sec)) +
 		    (end.tv_usec - begin.tv_usec);
-		printf("... in %zu blocks of size %zu:\t%.02f s\n",
-		    num_hashes, perfsizes[i], (double)delta / 1000000);
+		delta_s = (double)delta_us / 1000000.0;
+		printf("... in %zu blocks of size %zu:\t%.02f s\t%.01f MB/s\n",
+		    num_hashes, perfsizes[i], delta_s,
+		    ((double)bytes_to_hash / 1000000.0) / delta_s);
```
cperciva: I would have written this `(double)bytes_to_hash / delta_us` personally... MB/s is the same as B/us. :-)

gperciva: > MB/s is the same as B/us. :-)

Oh, I knew that.  However, when I experimented with showing GB/s, I lost track of all the `0`s.  I thought it was safer to treat each value separately.

... that said, now that I've defined `delta_us` instead of the previous version which was just `delta`, it's more clear.  Ok, I'll go with the assumption that any reader knows the connection between `M` and `u`.

gperciva:
Updated PR.  This version uses `assert()`s unconditionally inside `CRC32C_Update_SSE42()`.  I couldn't find any real speed difference based on the perftests.  I mean, if somebody really wanted to eke out the speed, they'd use `NDEBUG`?

(`clang 6` didn't optimize away the `assert()` statements.)

### With `assert()`

```
$ for a in `seq 10`; do ./tests/crc32/test_crc32 -t | grep Overall ; done;
Overall time and speed:	1.49 s	3951.6 MB/s
Overall time and speed:	1.48 s	3993.8 MB/s
Overall time and speed:	1.48 s	3999.7 MB/s
Overall time and speed:	1.50 s	3939.6 MB/s
Overall time and speed:	1.47 s	4020.6 MB/s
Overall time and speed:	1.48 s	4002.6 MB/s
Overall time and speed:	1.46 s	4056.9 MB/s
Overall time and speed:	1.46 s	4038.6 MB/s
Overall time and speed:	1.47 s	4008.9 MB/s
Overall time and speed:	1.50 s	3940.6 MB/s
```

### without `assert()`

```
$ for a in `seq 10`; do ./tests/crc32/test_crc32 -t | grep Overall ; done;
Overall time and speed:	1.50 s	3924.4 MB/s
Overall time and speed:	1.46 s	4047.4 MB/s
Overall time and speed:	1.45 s	4073.6 MB/s
Overall time and speed:	1.50 s	3949.2 MB/s
Overall time and speed:	1.45 s	4061.6 MB/s
Overall time and speed:	1.46 s	4038.7 MB/s
Overall time and speed:	1.47 s	4026.1 MB/s
Overall time and speed:	1.53 s	3865.9 MB/s
Overall time and speed:	1.46 s	4044.9 MB/s
Overall time and speed:	1.50 s	3931.6 MB/s
```

gperciva:
Hmm, do you want to use a cutoff of `8` or `16`?  Performance drops slightly when using `assert()`, which is expected because I added a third one.

I'm not convinced that it's worth special-casing small values, at least not without a bigger statistical test about the effect.  At a very rough eyeball, it looks like without asserts, using software or lengths below 8 or 16 gives about a 2.5% speed improvement.  But that complicates the code, and slows things down when using asserts.

(unless you want `CRC32C_Update_SSE42()` to absolutely assume that `len >= 8`, i.e. no sanity check.  I suppose that since it's only called when protected by `if (usecrc() && (len >= 8))`, it should be safe to do so.)

### With `assert()` and software for `< 8`:

```
$ for a in `seq 10`; do ./tests/crc32/test_crc32 -t | grep Overall ; done;
Overall time and speed:	1.55 s	3821.9 MB/s
Overall time and speed:	1.53 s	3858.3 MB/s
Overall time and speed:	1.54 s	3837.2 MB/s
Overall time and speed:	1.52 s	3880.1 MB/s
Overall time and speed:	1.54 s	3836.0 MB/s
Overall time and speed:	1.54 s	3823.4 MB/s
Overall time and speed:	1.51 s	3909.0 MB/s
Overall time and speed:	1.51 s	3907.9 MB/s
Overall time and speed:	1.51 s	3923.0 MB/s
Overall time and speed:	1.51 s	3906.5 MB/s
```

### without `assert()` and software for `< 8`:

```
$ for a in `seq 10`; do ./tests/crc32/test_crc32 -t | grep Overall ; done;
Overall time and speed:	1.46 s	4031.5 MB/s
Overall time and speed:	1.44 s	4113.7 MB/s
Overall time and speed:	1.42 s	4155.2 MB/s
Overall time and speed:	1.46 s	4053.4 MB/s
Overall time and speed:	1.46 s	4038.1 MB/s
Overall time and speed:	1.45 s	4071.7 MB/s
Overall time and speed:	1.46 s	4058.0 MB/s
Overall time and speed:	1.44 s	4100.4 MB/s
Overall time and speed:	1.42 s	4159.6 MB/s
Overall time and speed:	1.42 s	4155.7 MB/s
```

### without `assert()` and software for `< 16`:

```
$ for a in `seq 10`; do ./tests/crc32/test_crc32 -t | grep Overall ; done;
Overall time and speed:	1.45 s	4065.6 MB/s
Overall time and speed:	1.41 s	4175.0 MB/s
Overall time and speed:	1.46 s	4043.2 MB/s
Overall time and speed:	1.44 s	4111.2 MB/s
Overall time and speed:	1.45 s	4063.4 MB/s
Overall time and speed:	1.43 s	4141.6 MB/s
Overall time and speed:	1.43 s	4126.7 MB/s
Overall time and speed:	1.43 s	4119.7 MB/s
Overall time and speed:	1.46 s	4046.0 MB/s
Overall time and speed:	1.47 s	4030.6 MB/s
```

gperciva:
BTW, to test without `assert()`, I manually deleted them from `alg/crc32c_sse42.c`.  Leaving them there but compiling with `-DNDEBUG` gave slightly worse results than using `assert()` in the first place.  That's intriguing, so I'll investigate further.

Code uploaded with premature rebase.  Due to a git mistake on my desktop, I rebased onto master and had difficulty trying to get back to rebasing onto the remote `crc32-sse4.2`.  I figured that I might as well clean up the history.

I left the `len < 8` as a separate commit, though.

gperciva:
Preliminary follow-up on `-DNDEBUG` being slower:
- using `clang-6.0.0`
- I timed it with `test_sha256 -t` (the original one, no CPU extensions)
- no special cpu frequency games (so yes, frequency changes could be an issue)

### without `-DNDEBUG`:

```
$ time ./tests/sha256/test_sha256 -t
...
user	0m7.848s
user	0m7.920s
user	0m7.928s
user	0m7.884s
```
### with `-DNDEBUG`:

```
$ time ./tests/sha256/test_sha256 -t
...
user	0m7.904s
user	0m7.924s
user	0m7.972s
user	0m7.936s
```

By chance, I ran the test twice with NDEBUG, then recompiled and did it twice without, then recompiled again without and did another two tests, then finally recompiled with NDEBUG again and did the final two tests.

alg/crc32c_sse42.c
```diff
@@ -0,0 +1,58 @@
...
+
+	/* Process bytes before the alignment. */
+	for (; i < pre_block; i++)
+		*state = _mm_crc32_u8(*state, buf[i]);
+
+	/* If we would process a block, ensure that it's aligned. */
+	assert(!((i < in_block) && !((((uintptr_t)&buf[i]) & 7) == 0)));
+
+	/* Process blocks of 8. */
+	for (; i < in_block; i += 8) {
```
cperciva: This works, but perhaps only by accident?  `in_block` is the number of bytes within blocks of 8, not the position where said blocks end.

alg/crc32c_sse42.h
```diff
@@ -0,0 +1,16 @@
...
+#include <stdint.h>
+
+/**
+ * CRC32C_Update_SSE42(state, buf, len):
+ * Feed ${len} bytes from the buffer ${buf} into the CRC32C whose state is
+ * ${state}.  This implentation uses x86 SSE4.2 instructions, and should only
+ * be used if CPUSUPPORT_X86_CRC32 is defined and cpusupport_x86_crc32 returns
+ * nonzero.  ${len} must be greater than, or equal to, 8.
+ */
+void CRC32C_Update_SSE42(uint32_t *, const uint8_t *, size_t);
```
cperciva: In my testing, there's a nontrivial speedup if we make this

```
uint32_t CRC32C_Update_SSE42(uint32_t, const uint8_t *, size_t);
```

i.e., take the state as an input and output variable rather than passing it through as a pointer.  I suspect the speedup has to do with calling conventions and being able to keep things in registers.

I know I said that normally the _CPUFEATURE variants should be drop-in replacements using the same API, but when there's a performance delta I think it's worth making it slightly different.

cperciva:
On the question of when to cut across from the software version to the hardware version -- look at the speeds for the individual sizes!  What matters is whether hashing 8 bytes is faster with the new implementation; any difference in the 64 kB hashing is just going to be statistical noise.

cperciva:
But make the API for the _SSE42 variant explicitly "requires that len >= 8" because that's what we need in order to simplify the how-many-bytes-to-process-individually calculation.

tests/crc32/main.c
```diff
@@ -50,19 +143,53 @@ selftest(void)
...
 		}
+
+		/* Prepare the buffer in non-aligned memory. */
+		printf("Computing CRC32C of \"%s\" with non-aligned "
+		    "memory...", tests[i].s);
+		if ((unaligned = malloc(strlen(tests[i].s) + 3)) == NULL)
+			goto err0;
+		memcpy(unaligned + 3, tests[i].s, strlen(tests[i].s));
+
+		/* Try again with unaligned memory. */
```
cperciva: I think a better test here would be to construct a large buffer "helloworldhelloworldhelloworldhelloworld..." and then (a) hash it as a single buffer, and (b) feed in `random() % 256` bytes at a time.

That should ensure that we get every combination of lengths mod 8 and offsets mod 8, as well as testing that a series of Update calls does what they should -- right now I think we're only ever testing with a single Update call before we invoke Final.

alg/crc32c_sse42.c
gperciva: That did occur to me, and if I was using `i++`, then of course it would be an accident if it worked at all.  But with `i+=8` I think it's ok?

On the other hand, if both of us wondered about a "bad code smell", then it should probably be reworked anyway.

alg/crc32c_sse42.h
gperciva: Hmm, I'm seeing a slight slow-down from this.  I'll re-evaluate when we have the new cpu timing code.

cperciva: I guess it's possible that different compilers have differently quirky optimization behaviours here...

gperciva:
Delayed until #135 "updated CRC32 test" is in, so that we can better judge the timings.

gperciva:
Ok, update with timing (below).  Based on the timing data, I think it make sense to use the "store state in separate var".  Would you like that rebased into the commit that introduces CRC32, or should we keep it as a separate commit (i.e. if we want to revisit this with later CPUs / compilers)?

Unfortunately github doesn't display the commits in a nice order.  :/
- "basic" means not including the `MAYBE` commits or the `crc32_c_sse42.c: add asserts`
- "assert" means not including the `MAYBE` commits
- "nopointerstate" is everything.


### basic vs. asserts

(all timing done with 1 round of "warm up", then 30 timed runs)

It's a bit slower on GCC with 16, 32, 64, 128, and 256.  No change in higher hash sizes, and no change at all with clang.

basic-gcc-asserts-gcc-basic-clang-asserts-clang txt

### asserts vs. nopointerstate

nopointerstate is faster or equal in all cases.  Huge benefit to gcc; not much change with clang.  Clear win.
(no change with clang and hash size 256+)

asserts-gcc-nopointerstate-gcc-asserts-clang-nopointerstate-clang txt


### basic vs. nopointerstate (including assert)

Almost the same as assert vs. nopointerstate.  GCC with hash size 16 is equal (the penalty of assert is cancelled by the benefit of nopointerstate); everything else shows an improvement.


basic-gcc-nopointerstate-gcc-basic-clang-nopointerstate-clang txt

gperciva:
Rebased onto master, and rebased the non-pointer `state` into the history.  I left the asserts as a separate commit, though.

alg/crc32c_sse42.c
```diff
@@ -0,0 +1,54 @@
...
+		state = (uint32_t)_mm_crc32_u64(state,
+		    *(const unsigned long *)&buf[i]);
+	}
+
+	/* Process any remaining bytes. */
+	for (; i < len; i++)
+		state = _mm_crc32_u8(state, buf[i]);
+
+	return state;
```
cperciva: Style: Values in return statements should be enclosed in parentheses.

alg/crc32c_sse42.c
```diff
@@ -0,0 +1,54 @@
...
+		state = _mm_crc32_u8(state, buf[i]);
+
+	/*
+	 * Process blocks of 8.  It's ok if (i % 8 != 0), because we're
+	 * checking that (i < in_block), not (i != in_block).  For example,
+	 * if we start with i = 4 and want 1 block, we'll end up at i == 12.
+	 */
+	for (; i < in_block; i += 8) {
+		state = (uint32_t)_mm_crc32_u64(state,
+		    *(const unsigned long *)&buf[i]);
```
cperciva: Can we change this to `*(const uint64_t *)&buf[i]` ?  It seems that different headers spell this differently (clang has `unsigned long long` on my FreeBSD/amd64 system, the intel intrinsics guide says `unsigned __int64`, etc.) but it seems that `uint64_t` *should* work everywhere.

As it is right now, I worry that there will be some weird OS/compiler where `long` isn't 64 bits...

gperciva: Look ok on my system.  I agree that `long != 64` bits is a concern, which is why I find the Intel instrinsics often mindboggling.  (for example, `_mm_crc32_u32` takes `unsigned int` instead of `__uint32`)

cperciva:
Looks good, please rebase.

cpusupport/Build/cpusupport-X86-CRC32.c
```diff
@@ -0,0 +1,17 @@
...
+main(void)
+{
+	unsigned int state = 0;
+	unsigned char x = 0;
+	uint8_t buf[8] = {0, 0, 0, 0, 0, 0, 0, 0};
+
+	/* Test both the 8-bit and 64-bit data versions. */
+	state = _mm_crc32_u8(state, x);
+
+	state = (uint32_t)_mm_crc32_u64(state, *(const uint64_t *)&buf[0]);
```
gperciva: I changed this line from `unsigned long` to `const uint64_t` to match the other recent change.

gperciva:
Rebased, also making the `*(const uint64_t *)&buf[i]` change in the `cpusupport/Build/` file.  (I added a github comment on that line so you can find it easily).

PR #108, opened 2018-05-01
Build unify

gperciva:


gperciva:
Commits d2b1a05 and 6efaf13 are imported from spiped; the others are simplifications for keeping the Tarsnap project build files in sync.

I deliberately did not include the results of `make Makefiles`, because that changes every `Makefile` due to the addition of `$CPPFLAGS` and `$LDFLAGS`.  I'll do that in a separate PR after this is merged.

Issue #104, opened 2018-04-30
Duplicated tests?


cperciva:
It seems to me that `test_libcperciva.sh` runs the same tests as `make test` in the individual test directories.  Is there a reason for having two different ways to run the tests?

gperciva:
We could remove the individual-dir `make test` if you want.  That's the original way of running tests.  I figured there was no harm in leaving them around, though.

In the case of libcperciva, the tests are all (so far) simple and stick to one binary.  The only advantage (for libcperciva) of the whole `shared_test_functions.sh` framework that I can think of right now is that it handles valgrind.  However, for spiped, we definitely need the whole extra framework -- we can't launch `nc`, `spiped -d`, and `spiped -e` from a single binary.  (well, we *could*, but that would get really ugly)

At the moment, if you're doing a lot of work on CRC32 (for example), you could
```
cd tests/crc32/
make && make test
... edit code
make && make test
```

The alternative would be
```
make && ./test_crc32 -x
```
which is slightly more finicky.

cperciva:
Got it.  I guess it could be useful if the `make test` target ran the appropriate script, so that when the test changes it only needs to be changed once?

... and, now that I look at it, `make test` in `tests/setuidgid` doesn't actually work.

gperciva:
Hmm.  I have a minimal fix for `tests/setuidgid`.

We might want to discuss the overall testing strategy offline.  I'm confident that for projects other than libcperciva, it makes sense to have all the test logic in `tests/??-*.sh`, since the test binaries are used in multiple tests.  In those cases, there's no point in supporting `cd tests/foo && make test` at all.

It's muddier for libcperciva, because of the history of single-binary tests.  At the moment, `test/09-setuidgid.sh` uses `setup_check_variables`, which brings in a whole bunch of boilerplate variables defined in `shared_test_function.sh`.  It wouldn't be impossible to separate that out so that we can have a more independent test script in the `tests/setuidgid/` directory, of course.

Maybe come up 15 minutes before board games tonight?

gperciva:
From offline discussion: we'll keep `cd tests/foo && make test` for now, but might revisit it after kivaloo no longer requires such tests.

PR #109, opened 2018-05-02
Fix make test

gperciva:


gperciva:
Minimal fix for the setuidgid problem you mentioned in #104, but we might prefer a different solution in this case.  See the discussion there.

tests/setuidgid/Makefile.BSD
```diff
@@ -17,6 +17,6 @@ SRCS	+=	warnp.c
 IDIRS	+=	-I${LIBCPERCIVA_DIR}/util
 
 test:	test_setuidgid
-	@./test_setuidgid.sh
+	@./test_setuidgid $$$$(id -u):$$$$(id -g)
```
cperciva: You know that this breaks `make -f Makefile.BSD test`, right?

gperciva: ok, I reverted back to using a shell script.  I'll update this PR once #112 is merged (no point in generating more `Makefile` conflicts).

gperciva:
Updated PR

tests/setuidgid/Makefile.BSD
```diff
@@ -17,6 +17,6 @@ SRCS	+=	warnp.c
 IDIRS	+=	-I${LIBCPERCIVA_DIR}/util
 
 test:	test_setuidgid
-	./test_setuidgid.sh
+	./test-setuidgid-minimal.sh
```
cperciva: Why not just run the single command here?  Now that we're using awk to extract the test: target there's no issue with multiply escaping `$` characters.

PR #110, opened 2018-05-08
build: allow for no cpusupport dir

gperciva:


gperciva:
Most of our projects contain `${LIBCPERCIVA_DIR}/cpusupport/` so there's no problem calling `.../Build/cpusupport.sh`.  But kivaloo currently does not, and other new projects might not need cpusupport.

Another option is to always include `cpusupport/Build/cpusupport.sh` even if there's no other files in that directory.  I personally prefer to not include it and patch the `Makefile` as proposed, but if you'd rather always have `cpusupport.sh` then I'm content with that.

Makefile
```diff
@@ -21,7 +21,14 @@ all: cpusupport-config.h
...
 
 cpusupport-config.h:
-	( export CC="${CC}"; command -p sh ${LIBCPERCIVA_DIR}/cpusupport/Build/cpusupport.sh "$$PATH" ) > cpusupport-config.h
+	if [ -e ${LIBCPERCIVA_DIR}/cpusupport/Build/cpusupport.sh ];	\
+	then								\
+		( export CC="${CC}"; command -p sh 			\
+		    ${LIBCPERCIVA_DIR}/cpusupport/Build/cpusupport.sh	\
+		    "$$PATH" ) > cpusupport-config.h;			\
+	else								\
+		echo "" > cpusupport-config.h;				\
```
cperciva: Is the newline necessary?  If not, I would change `echo ""` to `:`.

cperciva:
Why `printf ""` rather than just `:`?  It's supported by POSIX...

gperciva:
> New applications are encouraged to use printf instead of echo.
> http://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html

cperciva:
Yes, if you want to actually print something.  But we don't...

gperciva:
oh, I get it now!  When you wrote
> change `echo ""` to `:`

, you meant that I should, err, change

    echo ""

into

    :

!  PR updated.

PR #114, opened 2018-05-10
build: list TESTS on separate lines

gperciva:


gperciva:
It's getting annoying to re-wrap this list whenever I add a new test.  Or rather: it's annoying to double-check that I didn't accidentally a test when adding a new one.

PR #115, opened 2018-05-11
Sha256 prep

gperciva:


gperciva:
This is extracted from #98, in case it's easier to handle the CPU detection (and new test vector) as a "bite-sized chunk".

cperciva:
When Intel first announced these, they were "AES New Instructions" and "SHA New Instructions", from which people obtained "AESNI" and "SHANI".  Intel likes to rename features after releasing CPUs which include them, because... well, actually, I don't know.  But they do it a lot.

I think using the original AESNI/SHANI names makes more sense than chasing Intel's rebranding.

gperciva:
Replaced by #127.

PR #117, opened 2018-05-12
Noreturn

gperciva:


gperciva:
Oops, my draft PDF was for C11, not C99.  Sorry.

PR #120, opened 2018-05-15
Fix warning with -DNDEBUG

gperciva:


cperciva:
I handled this in #128 where I spelled it slightly differently.

PR #121, opened 2018-05-15
Interrupt event loop

gperciva:


gperciva:
Alternate version of #94 limit number of events processed at once.

events/events.h
```diff
@@ -5,10 +5,11 @@
 
 /**
  * events_immediate_register(func, cookie, prio):
- * Register ${func}(${cookie}) to be run the next time events_run is invoked,
- * after immediate events with smaller ${prio} values and before events with
- * larger ${prio} values.  The value ${prio} must be in the range [0, 31].
- * Return a cookie which can be passed to events_immediate_cancel.
+ * Register ${func}(${cookie}) to be run the next time events_run or
+ * events_spin is invoked, after immediate events with smaller ${prio} values
```
gperciva: I'm no longer confident that we need to explicitly mention events_spin here.  A few weeks ago, I was 90% certain this was a good change; now I'm hovering around 60%.

events/events.c
```diff
@@ -79,7 +83,8 @@ doevent(struct eventrec * r)
  * associated with expired timers registered via events_timer_register will
  * be run.  If any event function returns a non-zero result, no further
  * events will be run and said non-zero result will be returned; on error,
- * -1 will be returned.
+ * -1 will be returned.  May be interrupted by events_interrupt, in which
+ * case -2 will be returned.
```
gperciva: Is -2 a decent value to return?  Returning 0 doesn't fly with `events_spin()`, and returning `>0` seems to indicate an error in the event handler.

events/events.h
cperciva: I don't think it really helps.  If you want to clarify things, I'd add a note in `events_spin` saying that it uses `events_run`.

events/events.c
```diff
@@ -19,6 +20,9 @@ MPOOL(eventrec, struct eventrec, 4096);
 /* Zero timeval, for use with non-blocking event runs. */
 static const struct timeval tv_zero = {0, 0};
 
+/* Interrupt a running event loop. */
+static volatile sig_atomic_t interrupt = 0;
```
cperciva: Maybe `interrupted` would make more grammatical sense?

events/events.c
cperciva: `0` is the right value to return -- exiting the loop because we asked to exit the loop is not an error.  Just have `events_spin` check the flag.

events/events.h
```diff
@@ -85,18 +85,26 @@ int events_timer_reset(void *);
...
+ * code from the callback).  May be interrupted by events_interrupt, in which
+ * case -2 will be returned.
  */
 int events_spin(int *);
 
+/**
+ * events_interrupt(signal):
+ * Halt the event loop after finishing the current event.
+ */
+void events_interrupt(int);
```
cperciva: The parameter here is weird.  I'd say that it makes sense if you want to register this as a signal handler... but I can't imagine what that would accomplish since if you do that you would exit the event loop and then not know what happened.

But this should be documented as being something which can be called safely from within a signal handler.

tests/events/main.c
```diff
@@ -0,0 +1,80 @@
...
+	sa.sa_handler = events_interrupt;
+	sigemptyset(&sa.sa_mask);
+	sa.sa_flags = 0;
+	if (sigaction(SIGUSR1, &sa, NULL))
+		goto err0;
+
+	/* Queue some events. */
+	for (i = 0; i < NUM_EVENTS; i++) {
+		I[i].n = i;
+		event_cookie[i] = events_immediate_register(&event, &I[i], 10);
```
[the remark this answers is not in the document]
cperciva: I would register a single immediate event, and have that event register a new event.  That would effectively create an infinite loop since `events_run` would always decide that it has more immediate events to run.

events/events.c
```diff
@@ -170,17 +182,17 @@ events_run(void)
 
 /**
  * events_spin(done):
- * Run events until ${done} is non-zero (and return 0), an error occurs (and
+ * Call events_run until ${done} is non-zero (and return 0), an error occurs (and
  * return -1), or a callback returns a non-zero status (and return the status
- * code from the callback).
+ * code from the callback).  May be interrupted by events_interrupt.
```
gperciva: should this be listed before the `(and return 0)` clause?  I tried moving it there, but it made the sentence seem a bit awkward.

events/events.c
```diff
@@ -189,6 +201,33 @@ events_spin(int * done)
...
+}
+
+/**
+ * events_interrupt_cancel():
+ * Cancel a requested interrupt from events_interrupt.  Must be used before
+ * running another event loop.  This function can be safely called from within
+ * a signal handler.
+ */
+void
+events_interrupt_cancel(void)
```
gperciva: API question: I considered automatically resetting `interrupt_requested = 0` once it was used, but then I figured that explicit was better than implicit, so I added this function instead.  Do you want application code to call `events_interrupt_cancel()` before running `events_run()` again?

(I also considered automatically setting `interrupt_requested=0` at the top of `events_run()` and `events_spin()`, but since the latter calls the former, that could end up with an interrupt being ignored.)

cperciva: Good question.  It should be reset automatically before `events_spin` and `events_run` return; otherwise we could "lose a wakeup" if a signal arrives at the wrong time.

Let's have the existing `events_run` become an internal function and have wrappers around it.

gperciva:
Updated PR, rebased on master to clear up the conflict and cleaned up history.  Added a wrapper for `event_run`, and automatically resets `interrupt_request` when beginning an event loop.

cperciva:
On 06/01/18 18:41, Graham Percival wrote:
> Updated PR, rebased on master to clear up the conflict and cleaned up history.
> Added a wrapper for |event_run|, and automatically resets |interrupt_request|
> when beginning an event loop.

What happens if a signal arrives and we call events_interrupt at exactly the
wrong time?

-- 
Colin Percival

gperciva:
Oops.  Reading is hard.

cperciva:
Looks good, please tidy up the commit history.

gperciva:
Updated PR with rebased history.

PR #123, opened 2018-05-16
Test sha256 cleanup

gperciva:


gperciva:
Let's not rely on the non-POSIX `CLOCK_VIRTUAL`.
http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_gettime.html

Incidentally,
```
$ uname -a
FreeBSD tarsnap-pkg-builder 11.1-RELEASE-p8 FreeBSD 11.1-RELEASE-p8 #0: Tue Mar 13 17:07:05 UTC 2018     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64
$ man clock_gettime
```

fails to mention that `CLOCK_VIRTUAL` is a FreeBSD extension, although it specifically lists other extensions to POSIX.

gperciva:
Oops, this might not work with `tests/sha256/perftest.sh`.  Untested so far because `lam` is BSD-specific.  I'll take a look at replacing it with `paste`, which is POSIX.

cperciva:
> Switching from CLOCK_VIRTUAL to CLOCK_MONOTONIC seems to add more variability to the timing.

No surprise there, since CLOCK_VIRTUAL is counting how much time the process is running while CLOCK_MONOTONIC is measuring wall-clock time.  How much variability will depend on how heavily loaded your system is.  But yeah, this is good for now.

> Should I look at supporting CLOCK_PROCESS_CPUTIME_ID, which is available in the optional [CPT] functionality in POSIX? There's a bunch of places where we want to benchmark code.

Hmm, good idea.  How about adding `monoclock_get_cpu` which uses `CLOCK_PROCESS_CPUTIME_ID` if available but falls back to the clocks used by `monoclock_get` if that's not available (not defined in the header or non-functional)?

PR #124, opened 2018-05-16
Compile ubuntu 1804

gperciva:


gperciva:
I'm not fond of either of these patches.  However,
- it looks like there's no avoiding `#define _DEFAULT_SOURCE 1`.  Ugly, but I figure that if this is the worst that glibc throws at us in the latest two-year Ubuntu update, we should consider ourselves fortunate!
- at least the horror show with `isnan()` only happens in test code.

gperciva:
And yes, of *course* there's an "Unused macro" that occurs on clang-5 and Ubuntu 14.04.  Not 16.04, and not 18.04.

I'm *really* not fond of 54fed72, but I didn't want to leave a broken PR before dinner.

util/setuidgid.c
```diff
@@ -13,6 +13,10 @@
 #define _BSD_SOURCE 1
 #define _DEFAULT_SOURCE 1
 
+/* Otherwise clang-5.0 + Ubuntu 14.-04 complains about an unused macro. */
```
cperciva: I think this is a case where it makes more sense to just turn off that warning in our tests?

cperciva:
Can you remove `If you're thinking WTFBBQ, you're not alone.` from the commit message.  It doesn't really add any information and I don't think it's the tone we want in permanent commit history...

gperciva:
Like you haven't written worse things in commit messages... ok, removed.  Also made the `_BSD_SOURCE` commit messages more polite.

PR #126, opened 2018-05-19
Build toplevel cpusupport

gperciva:


gperciva:
Includes #125.

cperciva:
Ick.  I think it might be cleaner to have something like

```
buildsubdir: cpusupport-config.h posix-flags.sh
    . ./cpusupport-config.h; \
    . ./posix-flags.sh; \
    export HAVE_BUILD_FLAGS=1; \
    cd ${BUILD_SUBDIR} && ${MAKE} ${BUILD_TARGET}
```

in the top-level Makefile...

cperciva:
My thought was that the code in tests/sha256 would say e.g.,

```
all:
if [ -z "$HAVE_BUILD_FLAGS" ]; then
    cd ../..
    make BUILD_SUBDIR=tests/sha256 BUILD_TARGET=all
    exit 0
fi
```

so that running `make all` in tests/sha256 would run the top-level makefile which would then run the lower-level makefile with the right environment set.

gperciva:
Updated PR to "loop back" to the top-level Makefile if necessary.

Makefile.BSD
```diff
@@ -34,8 +34,10 @@ ${D}/Makefile::
 	( cd ${D} && printf 'SRCS=' && ${MAKEBSD} -V SRCS | sed -e 's| cpusupport-config.h||' ) >> $@
 	( cd ${D} && printf 'IDIRS=' && ${MAKEBSD} -V IDIRS ) >> $@
 	( cd ${D} && printf 'LDADD_REQ=' && ${MAKEBSD} -V LDADD_REQ ) >> $@
+	( cd ${D} && printf 'SUBDIR_DEPTH=' && ${MAKEBSD} -V SUBDIR_DEPTH ) >> $@
 	if [ `cd ${D} && ${MAKEBSD} -V NOINST` ]; then		\
 		cat Makefile.prog |				\
+		    sed -e 's|<THISDIR>|${D}|' |		\
```
cperciva: Why not just make THISDIR a variable which is set like all the others?

cperciva: Also, I don't think what you have here even works since in the non-NOINST case you're not substituting?

gperciva:
Updated PR

PR #127, opened 2018-05-22
Sha256 shani prep

gperciva:


gperciva:
This replaces #115 by renaming `SHA` to `SHANI`.  (I didn't just force-push the previous PR because then we'd still have that branch if you decide you want `SHA` after all.)

cpusupport/Build/cpusupport-X86-SHANI.c
```diff
@@ -0,0 +1,17 @@
+#include <stdint.h>
+#include <immintrin.h>
+
+static uint8_t block[16] = {0};
+
+int
+main(void)
+{
+	uint32_t state[8] = {0};
```
cperciva: s/8/16/ ?

cpusupport/Build/cpusupport.sh
```diff
@@ -47,3 +47,4 @@ feature X86 CPUID ""
 feature X86 SSE2 "" "-msse2" "-msse2 -Wno-cast-align"
 feature X86 AESNI "" "-maes" "-maes -Wno-cast-align" "-maes -Wno-missing-prototypes -Wno-cast-qual"
 feature X86 RDRAND "" "-mrdrnd"
```
cperciva: Hmm, now that this list is getting nontrivial, maybe this and the list in `cpusupport.h` should be sorted?

cpusupport/cpusupport_x86_shani.c
```diff
@@ -0,0 +1,29 @@
...
+	unsigned int eax, ebx, ecx, edx;
+
+	/* Check if CPUID supports the level we need. */
+	if (!__get_cpuid(0, &eax, &ebx, &ecx, &edx))
+		goto unsupported;
+	if (eax < 7)
+		goto unsupported;
+
+	/* Ask about CPU features.  This macro takes e?x as pointers! */
+	__cpuid_count(7, 0, eax, ebx, ecx, edx);
```
cperciva: Why not `__get_cpuid` like we do for the other cpusupport functions?

cperciva: Oh, just saw the commit about this, never mind.

cperciva: We should add `__cpuid_count` to the test in `cpusupport-X86-CPUID.c` in case there are platforms which have `__get_cpuid` but not `__cpuid_count`.

cpusupport/Build/cpusupport-X86-SHANI.c
gperciva: Hmm, don't you mean 4?  It's `unit32_t`, and getting loaded into `__m128i`.

gperciva: We have
```
static void
SHA256_Transform(uint32_t state[static restrict 8],
    const uint8_t block[static restrict 64],
    uint32_t W[static restrict 64], uint32_t S[static restrict 8])
{
```

because the state is split up into two separate `__m128i` variables.  But when checking for the available of the instructions, we don't need to bother with that.

cperciva: Err, yeah.  I was confused because I think of XMM registers as being 16 bytes, not 4 doublewords.

cperciva:
Obsoleted by #140

PR #129, opened 2018-05-26
Valgrind setuidgid NSS

gperciva:


gperciva:
Annoying memory leak in FreeBSD and Ubuntu 18.04.  It occurs in NSS, through some linking magic, so I can't simply use `getpwnam()` and `getgrnam()` in the `potential-memleaks.c` file.  Instead, those function calls need to occur in separate binaries.  :(

(because if I only check for `getpwnam()` in `potential-memleaks.c`, then it doesn't generate a suppression for the NSS leak being caused by `getgrnam()`.  And if I use both functions in `potential-memleaks.c`, then the leak is deemed to have been caused by the first function used, with no reference to the second one.)

cperciva:
Rather than having separate binaries here, why not have a single binary which takes a parameter telling it which routine to run?  It could have e.g.,

```
#define MEMLEAKTEST(x) { ##x, x }
struct memleaktest {
    char * name;
    void (* func)(void);
} tests[] = {
    MEMLEAKTEST(pl_freebsd_link_lrt),
    MEMLEAKTEST(pl_freebsd_printf),
    MEMLEAKTEST(pl_freebsd_getdelim)
};
```

and if invoked without any parameters have it emit the list of test names, so that you could run

```
potential-memleaks |
    while read t; do
        valgrind potential-memleaks ${t}
done
```

and then every test will be running independently of the rest and all you have to do is add a new line to that array when there's a new test.

gperciva:
That's a deliciously twisted, or twistedly delicious, solution!

gperciva:
Revised PR, now includes #133 (run the valgrind binary multiple times, as you suggested).

PR #130, opened 2018-05-28
Store POSIX flags in a build-file

gperciva:


gperciva:
Split off from #126.

gperciva:
(the build failure was due to mpool being 66% < 75% slower than malloc instead on one particular Travis-CI run.  I could lower that bound to 60%, but even 75% is silly -- I mean, if it's normally that slow, we have a serious problem that needs investigating.  I think we need to move that test to our own CI system on dedicated hardware.  We can discuss it offline, though.)

Makefile
```diff
@@ -21,15 +21,33 @@ TEST_CMD=	tests/test_libcperciva.sh
...
-	export "LDADD_POSIX=`export CC=\"${CC}\"; cd ${LIBCPERCIVA_DIR}/POSIX && command -p sh posix-l.sh \"$$PATH\"`";	\
-	export "CFLAGS_POSIX=`export CC=\"${CC}\"; cd ${LIBCPERCIVA_DIR}/POSIX && command -p sh posix-cflags.sh \"$$PATH\"`";	\
+	. ./posix-flags.sh;				\
 	. ./cpusupport-config.h;			\
 	for D in ${PROGS} ${TESTS}; do			\
 		( cd $${D} && ${MAKE} all ) || exit 2;	\
 	done
 
+# To avoid escape characters becoming overwhelming, this generates the _POSIX
+# environment variables first, then writes them to the file.
```
cperciva: I'm not sure I understand this.  Isn't it enough to have

```
printf "export \"LDADD_POSIX="
cd ${LIBCPERCIVA_DIR}/POSIX && command -p sh posix-l.sh "$$PATH"
printf "\"\nexport \"CFLAGS_POSIX="
cd ${LIBCPERCIVA_DIR}/POSIX && command -p sh posix-cflags.sh "$$PATH"
printf "\"\n"
```
?

Makefile
```diff
@@ -21,15 +21,33 @@ TEST_CMD=	tests/test_libcperciva.sh
...
+posix-flags.sh:
+	if [ -d ${LIBCPERCIVA_DIR}/POSIX/ ];				\
+	then								\
```
cperciva: I would normally put `then` on the same line as `if`, unless there's a reason to do otherwise.

gperciva:
Updated PR with revised shell escaping.  Not exactly what you suggested, but I think we'll both be happy with it.  Since it's short enough, I rewrote the history as well.

cperciva:
The code seems much simpler with these two commits, and I can't see any situation where they wouldn't work... do you see something wrong with them?

gperciva:
huh, I could have sworn that `export "CC=${CC}"` only worked if it was inside a subshell.  But the Travis-CI logs clearly show that it *does* work.

Do you want to merge it as-is, or rebase into the two initial commits?

gperciva:
PR updated, including a few more tweaks to the cpusupport-config creation (putting the `export CC` line on its own, to match posix-flags creation.

Ready for merge.

PR #131, opened 2018-05-29
Monoclock get cputime

gperciva:


util/monoclock.c
```diff
@@ -61,3 +61,36 @@ monoclock_get(struct timeval * tv)
...
+monoclock_get_cputime(struct timeval * tv)
+{
+	struct timespec tp;
+
+	/* Use CLOCK_PROCESS_CPUTIME_ID if available. */
+#ifdef CLOCK_PROCESS_CPUTIME_ID
+	if (clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &tp) == 0) {
+		tv->tv_sec = tp.tv_sec;
+		tv->tv_usec = (suseconds_t)(tp.tv_nsec / 1000);
+	} else if ((errno != ENOSYS) && (errno != EINVAL)) {
```
cperciva: Umm, what happens if clock_gettime returns `ENOSYS`?

util/monoclock.h
```diff
@@ -11,4 +11,12 @@
  */
 int monoclock_get(struct timeval *);
 
+/**
+ * monoclock_get_cputime(tv):
+ * Store how much time the process has been running in ${tv} if
```
cperciva: "how much time the process has been running in ${tv}" seems dangerously easy to misparse.  Maybe "Store in ${tv} the duration the process has been running..."?

gperciva: That's easier to parse, but it looks really silly for the simple `monoclock_get()`: "Store in ${tv} the current time.".  And it also looks weird to have different grammar in such simple functions.

I've uploaded this change so you can see how it looks in reality.  Despite the awkwardness, I actually *do* prefer the new version, since it's easy to parse.

util/monoclock.c
```diff
@@ -19,7 +19,7 @@
 
 /**
  * monoclock_get(tv):
- * Store the current time in ${tv}.  If CLOCK_MONOTONIC is available, use
+ * Store in ${tv} the current time.  If CLOCK_MONOTONIC is available, use
```
cperciva: Nah, it just feels clumsy here.  It's a bit clumsy in the other function too, but there we need it for clarity.

util/monoclock.c
cperciva: ?

gperciva: ??

I just copied those lines from `monoclock_get()`.  `ENOSYS was removed in Issue 6:
http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_gettime.html
and there's no mention of it in the Linux or FreeBSD man pages.

Are you saying that if we get `ENOSYS` we should silently fall back to `monoclock_get()`?

gperciva:
Updated PR with rebased history.  Still not certain about `ENOSYS`.

util/monoclock.c
cperciva: You failed at copying.  Right now if we get ENOSYS or EINVAL, this code silently returns 0.

But yes, we should fall back to monoclock_get if we get ENOSYS or EINVAL.  It's perfectly legal for the kernel to return EINVAL (we might e.g., have compiled using headers which are newer than the kernel) and ENOSYS is also sometimes returned (I can't remember exactly when, but I think it was Linux -- pretty sure there was an issue reported about that a couple years ago).

We should fall back to monoclock_get if CLOCK_PROCESS_CPUTIME_ID is not defined *or* if it fails with ENOSYS/EINVAL.

gperciva:
I think this update fixes the `#ifdef` and `if ... else` logic to do the fallback you wanted.  Needs a rebase.

gperciva:
Whoops.  Another fix.  I'm not certain what the desired indentation is of the `return monoclock_get`.

I'm tempted to ignore the example of clever code in `monoclock_get()` and write this in a more verbose, but easier to understand, fashion.  But I know that matching the existing style is good, so here we are.

util/monoclock.c
```diff
@@ -61,3 +61,35 @@ monoclock_get(struct timeval * tv)
...
+/**
+ * monoclock_get_cputime(tv):
+ * Store in ${tv} the duration the process has been running if
+ * CLOCK_PROCESS_CPUTIME_ID is available; fall back to monoclock_get()
+ * otherwise.
+ */
+int
+monoclock_get_cputime(struct timeval * tv)
+{
+	struct timespec tp;
```
cperciva: I think this needs to be wrapped in an `#ifdef` too?

gperciva:
Updated PR.  I put the variable declaration inside the `#ifdef` because it looked silly otherwise.

PR #132, opened 2018-06-01
tests: use monoclock_get_cputime()

gperciva:


gperciva:
Unsurprisingly, this makes the timing much more reliable.  On my desktop, running `stress --cpu 4` in another window, and testing with
```
for a in `seq 100` ; do ./tests/mpool/test_mpool 1 1000000 0 &>> gettime.txt ; done;
```

I get the following plot:
output

(I made no attempt to avoid cpu frequency scaling in this quick&dirty test)

PR #133, opened 2018-06-05
Valgrind run binary multiple times

gperciva:


gperciva:
This is split off from #129.  It doesn't fix the leaks arising from setuidgid; that will come in a later PR.

gperciva:
(and yes, once this PR is in, catching the leaks in NSS is as simple as you would expect)

PR #135, opened 2018-06-08
Crc32 test

gperciva:


gperciva:
Split off from #107.  It'll be easier to compare CRC32C timings if `master` has the timing test as well.

PR #137, opened 2018-07-07
cpusupport: sort lists alphabetically

gperciva:


cpusupport/Build/cpusupport.sh
```diff
@@ -43,7 +43,10 @@ feature() {
 	esac
 }
 
+# Detect CPU-detection features
```
gperciva: It's not actually required to detect CPUID before trying the other features (because the compile-time test doesn't depend on CPUID; only the run-time tests do).  But I think it's nice to sort the lists separately, because otherwise I'll stare at `cpusupport.h` and wonder why the number of CPUSUPPORT_FEATURES isn't the same as the list in `cpusupport.sh`.

PR #140, opened 2018-07-11
Use SHANI instructions for SHA256 computations

cperciva:


cperciva:
@gperciva Leaving aside the silly compiler warnings for the moment, can you performance test this on real hardware?

gperciva:
FWIW, this branch up to 7f054bad026f1d66d46b6ed274df682608b0897e compiles just fine.  (that's the last `make Makefiles` before you play with SHA256 code).

gperciva:
What do you think about printing whether or not it's using hardware instructions?  i.e.
https://github.com/Tarsnap/libcperciva/pull/107/commits/3f6f0669415c91416aca98b4e60d5f2dfaa814b9

That might be helpful to double-check that the compiler / CPU is doing what we think it should be doing, and also so that when there's timing log files floating around it's clear what they refer to.

gperciva:
Well, it's faster:

```
::::::::::::::
master.txt
::::::::::::::
SHA256 time trial. Digesting 100000 10000-byte blocks ... done
Digest = 3f7daf2b36997b6eb269f63176e09fee8f6daa13202d87e020da63d262c6a4ca
Time = 3.278087 seconds
Speed = 305055967.093003 bytes/second
::::::::::::::
shani.txt
::::::::::::::
SHA256 time trial. Digesting 100000 10000-byte blocks ... done
Digest = 3f7daf2b36997b6eb269f63176e09fee8f6daa13202d87e020da63d262c6a4ca
Time = 0.595845 seconds
Speed = 1678288816.722470 bytes/second
```

Should I break that down by hash size, like we're doing for CRC32?

gperciva:
BTW,
```
model name	: AMD Ryzen 7 2700 Eight-Core Processor
stepping	: 2
microcode	: 0x8008204
cpu MHz		: 1377.200
cache size	: 512 KB
```

PR #144, opened 2018-07-13
alg/sha256_shani: convert binary constants to hex

gperciva:


gperciva:
Sorry, I forgot to check the SHA code in gcc.  Binary constants are a GCC extension, which it warned about.  The C99 rationale document V5.10 explicitly rejects this idea in 6.4.4.1 line 30.

cperciva:
Huh, why did I think binary constants were part of the standard?  Thanks for catching this.

Issue #83, opened 2017-11-05
Teach util/entropy.c about other APIs


cperciva:
Linux 3.17+ has `getrandom(2)`.  Linux also has `getentropy(3)`, which theoretically behaves the same way as the `getentropy(2)` in OpenBSD 5.6+.  FreeBSD has the `kern.arandom` sysctl.  iOS has `SecRandom`.  Windows has `CryptGenRandom`.

gperciva:
Is the idea that we use those other APIs instead of `/dev/urandom` and only fall back to that if they fail?

gperciva:
I'm guessing that you'll want a
```
entropy_linux_getrandom.c
```

(like `alg/sha256_shani.c` or `crypto/crypto_aes_aesni.c`).  Do you think that should stay in the `util/` directory, or should we move it to a new `entropy/` directory?  i.e. the first step would be moving `util/entropy.{ch}` to `entropy/entropy.{ch}`.

(I've had this code ready to go for a while, but I thought that we should get the current round of releases done before adding it.)

cperciva:
I'd prefer to keep it in `/util/`.  In general, that's where non-POSIX bits belong.

PR #145, opened 2018-07-19
Apisupport

gperciva:


gperciva:
In order to address the API-specific #83 (teach entropy.c about non-`/dev/urandom` APIs), we need to have optional compilation.

My first thought was to include this in cpusupport directly, but then I realized that was silly, since all I needed to do was a bunch of search&replace operations.  You can see this in operation in #146.

gperciva:
sigh.  Build failure due to mpool.  (#146 has more serious problems, but I was shocked to see this one fail)

gperciva:
This likely needs a change to cpusupport before it's suitable for merging.

gperciva:
The above-mentioned change has been made, so I think this is good to go if you like this `apisupport` in general.

PR #146, opened 2018-07-19
Apisupport linux getrandom

gperciva:


gperciva:
This includes #145 Apisupport.  I'm less certain about some parts of this, but I figured it was ok as a starting place for discussion.

apisupport/Build/apisupport-LINUX-GETRANDOM.c
```diff
@@ -0,0 +1,16 @@
...
+#include <stdlib.h>
+
+#define BUFLEN 2
+
+int
+main(void)
+{
+	char buf[BUFLEN];
+
+	if (getrandom(buf, BUFLEN, 0) != BUFLEN)
```
gperciva: I don't know if we want to actually check that we can get bytes from `getrandom()`, or just let this auto-succeed and handle the possible failure in the calling function.

gperciva: Oops, I meant to add this comment to `apisupport/apisupport_linux_getrandom.c`, not this file.

util/entropy.c
```diff
@@ -34,6 +39,16 @@ entropy_read(uint8_t * buf, size_t buflen)
 		goto err0;
 	}
 
+#ifdef APISUPPORT_LINUX_GETRANDOM
+	if (apisupport_linux_getrandom()) {
+		if (getrandom(buf, buflen, 0) != (ssize_t)buflen) {
```
gperciva: 1. Do we want to move this to a separate `.c` file?  Since this doesn't require any additional compiler flags and it's so short, that seemed excessive.  Or I could move it to a separate function in the same file?

2. The `entropy_read()` function is general, although we're only calling it for 32 or 48 bytes.  In theory, `getrandom()` on Linux should never return fewer bytes than `buflen` as long as it's less than 256.  Not that I'm suggesting that we don't check the value, of course.

I was more wondering if it would be useful to have an `entropy_read_byte_buflen()` [yes, silly name] to indicate that it's intended for small buffers.

PR #149, opened 2018-09-01
Document & test getopt hyphen

gperciva:


gperciva:
My finger missed adding a one-letter argument to scrypt or spiped, and I was surprised to find the program continue as if nothing was weird.

cperciva:
I'm confused.  Can you give me an example of what you (accidentally) ran and what it did / should have done?

gperciva:
The difference is "do we quit the getopt loop silently, or go to `GETOPT_DEFAULT`".  I want to go to `GETOPT_DEFAULT`.

libcperciva master:

```
$ ./tests/getopt/test_getopt -
Option being processed: (dummy)
bflag = 0
Remaining arguments: -
Option being processed: (dummy)
$
```

libcperciva after patch:

```
$ ./tests/getopt/test_getopt -
Option being processed: (dummy)
test_getopt: unknown option: -
Option being processed: -
usage: blah blah blah
bflag = 0
Remaining arguments:
Option being processed: (dummy)
Option being processed: -
usage: test_getopt [-b] [-f bar] [--foo bar] ...
$
```

gperciva:
... just occurred to me that this PR would mess up using stdin / stdout in scrypt.  Never mind.

gperciva:
I removed the code changes and changed the tests to include

```
$ test_getopt -
```

because otherwise I'm likely to try to change this again in a few months.

I also kept the "document using ` -- ` to indicate no more arg parsing" line in `getopt.h`.

cperciva:
On 9/1/18 2:06 PM, Graham Percival wrote:
> ... just occurred to me that this PR would mess up using stdin / stdout in
> scrypt. Never mind.

Right, in traditional UNIX command-line parsing, "-" isn't an unknown option,
it's simply an argument no different from "foo".

-- 
Colin Percival

util/getopt.h
```diff
@@ -12,6 +12,7 @@
  * -abcfoo		(packed single-character options and an argument)
  * --foo bar		(long option)
  * --foo=bar		(long option and argument separated by '=')
+ * -b -- ignore these   (some arguments followed by a "--" and then anything)
```
cperciva: "This getopt implementation parses options of the following forms... stuff which isn't options and doesn't get parsed"?

I think your intent here is better served by adding a new paragraph:

```
Upon encountering '--', it consumes that argument (by incrementing optind) and
returns NULL to signal the end of option-processing.  Upon encountering any
argument not starting with '-' (except as an argument to an option) it returns NULL
to signal the end of option-processing (without consuming the argument).
```

This is all standard UNIX getopt behaviour but it's probably worth documenting...

gperciva:
Updated PR with rewritten history.  I added an extra sentence to your proposed paragraph:

```
Upon encountering '-' by itself (not immediately followed by a
character) it returns NULL to signal the end of option-processing (without
consuming the argument).
```

because that was precisely the case that prompted this PR.

cperciva:
I think my commit here clarifies things further.  What do you think?

gperciva:
LGTM.  Should I play with git rebase history, or leave it as-is?

cperciva:
No need to rewrite history, I've merged as is.  Thanks.

PR #150, opened 2018-10-25
tests/valgrind: forbid inlining of pl_*()

gperciva:
We want the pl_*() names to be visible in the valgrind output, otherwise we
can't properly generalize the suppressions.

Unfortunately there doesn't seem to be a non-compiler-specific way to mark
functions as "not inline".  Fortunately, using a volatile function pointer
forces the compiler to not inline it.

### DIFF
--- a/tests/valgrind/potential-memleaks.c
+++ b/tests/valgrind/potential-memleaks.c
@@ -54,8 +54,8 @@ pl_nss_getpwnam(void)
 
 #define MEMLEAKTEST(x) { #x, x }
 static const struct memleaktest {
-	const char * name;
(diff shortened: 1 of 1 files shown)


gperciva:
I haven't yet seen this needed in any libcperciva software, but clang "helpfully" inlining the valgrind test functions *is* a problem in the GUI.  Does this solution make sense?

cperciva:
This isn't guaranteed to work -- C compilers are allowed to figure out that objects aren't "really" volatile despite having the volatile keyword -- but I think it's a reasonable workaround, yes.

PR #152, opened 2018-11-23
Safely append 0 items to an empty elasticarray

gperciva:


gperciva:
I've been annoyed by clang's scan-build not quite liking libcperciva, but so far I've ignored the 3 copies of this warning (the only one).  However, when I started looking at the recent kivaloo changes, I see 43 warnings, of which 27 are copies of this issue.  So I figured I'd do something about it.

I recall that you didn't want `assert()` in elasticarray for speed reasons, but surely if speed is that important, they can compile it with `-DNDEBUG`.  Also, we have a bunch of non-assert sanity checks for overflows in there already.

I'm content to drop the `assert(nrec > 0)` checks that I added in the second commit.  That's a "developer friendliness" thing, not a "silence clang" thing.

cperciva:
I'm inclined to say that we should support appending zero records to an empty array.  I mean, it's a silly thing to do, but if we don't support it then future code might need to do things like

```
if (nrecs > 0)
    foo_append(array, buf, nrecs)
```

which seems unnecessarily complicated.

How about just adding a `if (nrec > 0)` in front of the `memcpy`?  Arguably scan-build is correct in warning, since right now we could end up calling `memcpy(NULL, buf, 0)` which is undefined behaviour.

gperciva:
Ok.  Even with the `if (nrec > 0)` check, clang still isn't happy without an `assert()`.  
9e4dff8 works.

I've tried adding `assert(reclen > 0)` since we have an implicit assumption of that (given that `elasticarray_append()` is used via `ELASTICARRAY_DECL` and `sizeof()` won't return 0), but that didn't help it figure out that it's safe.

cperciva:
This seems to be a recently-added case of scan-build stupidity; it's not present in llvm 6.0.  I've filed a bug report with LLVM about this, and until I hear back about that I'm not inclined to sprinkle assertions just to keep the scanner quiet.

gperciva:
Ok.  The latest version of this PR just adds `if (nrec > 0)` since that makes sense to us as humans, and I'll avoid scan-build in clang 7.

PR #154, opened 2018-11-25
AWS

gperciva:


cperciva:
Good catch on the `<stddef.h>`.  I don't think we need any more tests than what you've got here.

Please rebase and then I'll merge.

gperciva:
Rebased, ready for merge.

PR #156, opened 2018-11-25
Json

gperciva:


gperciva:
Needs rebasing, but I kept the changes separate so you could see them easily.

cperciva:
Doesn't ab104b1a break finding non-7-bit-clean keys?

gperciva:
I'm not certain that I understand.  Should I add
```
	/* Non-7-bit-clean. */
	{ "{\"tête\":\"oui\"}", "tête", "oui"},
```

to `tests/json/main.c` and make sure it works?  Because that doesn't work right now, even if I take out the clang strict sign conversion commits.

gperciva:
Wait, I misread the test cases.  (err, do you *really* want to find `"bar"}` instead of just `bar` ?!?)

Your code works with extended ascii.

util/json.c
```diff
@@ -43,7 +43,7 @@ skip_literal(const uint8_t * buf, const uint8_t * end)
 static const uint8_t *
 skip_string(const uint8_t * buf, const uint8_t * end)
 {
-	char ch;
+	uint8_t ch;
```
gperciva: I'm changing this variable because on lines 53 and 59, we assign it a value from `buf`, which is `uint8_t`.

util/json.c
```diff
@@ -220,7 +220,7 @@ match_str(const uint8_t * buf, const uint8_t * end, const char * s,
 	do {
 		if (buf == end)
 			return (end);
-		ch = *buf++;
+		ch = (char)(*buf++);
```
gperciva: I'm not changing the `char ch` on line 214 of this function because it's compared to `char s[0]` on line 275.

However, to make clang happy, I'm casting the value it gets from `buf` here.

gperciva:
Tentative update to PR, needs rebase.

More test cases might be good, especially if the json files might include non-extended-ascii characters.

cperciva:
Looks good, please clean up the commit history.

gperciva:
Updated with rebased history, ready for merge.

PR #157, opened 2018-11-27
Fix publish

gperciva:


gperciva:
The real reason for updating `make publish` here is so that I can fix it in kivaloo.

cperciva:
Speaking of fixing things in kivaloo, I noticed that `make Makefiles` isn't generating `test` targets.  I think that was working in a different repo -- maybe that just never got merged across?

gperciva:
Yes, I'm aware of the `make test` thing.  There's a bunch of build-system changes which I've been holding off on -- `SUBDIR-DEPTH`, top-level flag files, `-DCPUSUPPORT_CONFIGFILE`, etc.  Those involve a lot of changes to the generated `Makefile`s, so I didn't want to start that until dynamodb was finished.

(That's also why your dynamodb isn't using `sha256-shani.c`)

PR #159, opened 2018-12-07
timeval_diff

gperciva:


gperciva:
Every time I write some benchmarking code, I have to look up the POSIX definition of `struct timespec` to see whether it contains `.usec` or `.nsec`.  I know that you said you didn't like this idea, but can we please abstract this away?

util/monoclock.c
```diff
@@ -94,3 +94,17 @@ monoclock_get_cputime(struct timeval * tv)
 	/* Failure! */
 	return (-1);
 }
+
+/**
+ * monoclock_diff_seconds(start, stop):
+ * Calculate the difference (in seconds) between a start time and an end time.
+ */
+long double
```
cperciva: Why on earth would you want to use `long double` for this?

util/monoclock.h
```diff
@@ -19,4 +19,10 @@ int monoclock_get(struct timeval *);
  */
 int monoclock_get_cputime(struct timeval *);
 
+/**
+ * monoclock_diff_seconds(start, stop):
+ * Calculate the difference (in seconds) between a start time and an end time.
+ */
+long double monoclock_diff_seconds(struct timeval, struct timeval);
```
cperciva: How about making this a macro?
```
#define timeval_diff(x, y) ((y.tv_sec - x.tv_sec) + (y.tv_usex - x.tv_usec) * 0.000001)
```

I'd be inclined to include it in `monoclock.h` but call it `timeval_diff` since it applies to timevals in general.

gperciva:
Updated PR.

PR #160, opened 2018-12-24
 tests: use grep -F for has_pid()

gperciva:


cperciva:
Hmm, this may be a dumb question, but why not just use fgrep?

gperciva:
Well, `fgrep` isn't POSIX.  But `grep -F` is, and that's the proper solution.  I've put that in https://github.com/Tarsnap/spiped/pull/186.

gperciva:
Import from spiped

PR #163, opened 2019-04-07
Add monoclock_getres()

gperciva:


gperciva:
This is split off from https://github.com/Tarsnap/scrypt/pull/156 (use monoclock in `scryptenc_cpuperf.h`).

util/monoclock.c
```diff
@@ -94,3 +94,35 @@ monoclock_get_cputime(struct timeval * tv)
...
 	return (-1);
 }
+
+/**
+ * monoclock_getres(resd):
+ * Store the timer resolution in ${resd}. If CLOCK_MONOTONIC is available, use
+ * that clock; if CLOCK_MONOTONIC is unavailable, use CLOCK_REALTIME (if
+ * available) or gettimeofday(2).
+ */
+int
```
cperciva: Why?  The way you've written this function, it can't fail.

... perhaps it should be able to fail, e.g., if `clock_getres(CLOCK_MONOTONIC)` fails with something other than `ENOSYS` or `EINVAL`, mind you.

util/monoclock.c
```diff
@@ -94,3 +94,35 @@ monoclock_get_cputime(struct timeval * tv)
 	/* Failure! */
 	return (-1);
 }
+
+/**
+ * monoclock_getres(resd):
+ * Store the timer resolution in ${resd}. If CLOCK_MONOTONIC is available, use
```
cperciva: I'd say "Store an upper limit on timer granularity" here -- on many systems gettimeofday has microsecond accuracy.

You should probably have a comment somewhere saying that we're assuming that `clock_getres` works iff `clock_gettime` works.  It's a reasonable assumption to make, but it should still be documented... I could imagine embedded systems without RTC clocks which don't implement `clock_getres(CLOCK_REALTIME)` but nonetheless return a timer granularity for `clock_getres(CLOCK_*)`.

gperciva:
Yuck, I don't like having to assume that `clock_getres()` works iff `clock_gettime()` does.  I totally agree that since we're making that assumption, it should be documented (in the header file, no less).  I also really don't like the hanging `else`.  I know that it works, but I'm not fond of that type of `#ifdef` games.

Still, it's a step forward (as it will allow us to clean up scrypt).  And the scrypt code which did timing stuff also had the implicit assumption about `clock_getres` iff `clock_gettime`.

util/monoclock.c
```diff
@@ -94,3 +94,47 @@ monoclock_get_cputime(struct timeval * tv)
 	/* Failure! */
 	return (-1);
 }
+
+/**
+ * monoclock_getres(resd):
+ * Store an upper limit on timer granularity in ${resd}. If CLOCK_MONOTONIC is
+ * available, use that clock; if CLOCK_MONOTONIC is unavailable, use
+ * CLOCK_REALTIME (if available) or gettimeofday(2).  For this value to be
+ * meaningful, we assume that clock_getres(x) works iff clock_gettime(x) works.
```
cperciva: I would s/works/succeeds/g

util/monoclock.c
```diff
@@ -94,3 +94,47 @@ monoclock_get_cputime(struct timeval * tv)
...
+#ifdef USE_REALTIME
+	if (clock_getres(CLOCK_REALTIME, &res) == 0) {
+		/* Convert clock resolution to a double. */
+		*resd = res.tv_sec + res.tv_nsec * 0.000000001;
+	} else {
+		warnp("clock_getres(CLOCK_REALTIME)");
+		goto err0;
+	}
+#else
+	/* Will be using gettimeofday(). */
```
cperciva: I'd expand this somewhat:
```
/*
 * We'll be using gettimeofday().  There is no standard way of getting the
 * resolution of this clock, but it seems safe to assume that it ticks at
 * a minimum rate of CLOCKS_PER_SEC Hz (even though that is defined in
 * relation to the measurement of processor time usage, not wallclock time);
 * on non-broken systems we'll be relying on clock_gettime and clock_getres
 * anyway.
 */
```

cperciva:
Looks good aside from the comment changes I suggested.

gperciva:
Updated PR, ready for merge.  Since the latest changes were merely to comments, I took the liberty of rebasing them.

PR #166, opened 2019-04-17
monoclock: strict usage of #if define

gperciva:


gperciva:
This wasn't discovered by Travis-CI because they currently use 10.13.3, which is evidently late enough.  IIRC Apple fixed the `CLOCK_GETTIME` POSIX violation in 10.12 (but don't quote me on that).

cperciva:
IIRC some versions of OS X define CLOCK_REALTIME but don't implement it (so we end up calling gettimeofday).  I think the latest versions finally get it right though...

gperciva:
With OSX 10.11 and an older Xcode (pre-2016, I think), there's no CLOCK_REALTIME at all.  With OSX and a recent-ish Xcode (i.e. my desktop, after a reboot) CLOCK_REALTIME is defined but not implemented.  That's why `posix-cflags.sh` tries to run it.  With modern OSX (such as in Travis-CI), it has CLOCK_REALTIME and all is well.

cperciva:
Right, I meant in the kernel -- I think there's some versions where it's defined in header files and it links, but the kernel returns ENOSYS.

Or maybe that was on Linux, come to think of it.  *shrug*

PR #167, opened 2019-04-25
Test elasticarray

gperciva:


tests/elasticarray/main.c
```diff
@@ -0,0 +1,167 @@
...
+			goto err0;
+	}
+
+	/* Use the list. */
+	for (i = 0; i < dynlist_getsize(list); i++) {
+		p = *dynlist_get(list, i);
+		printf("dyn:\t%i\n", p->arr[0]);
+	}
+
+	/* Free memory. */
```
gperciva: This part is the real reason I'm adding this test.  Whenever I try to use elasticarray for structs, I end up in a mess of memory leaks and/or invalid `free()`.  (and yes, one of my branches currently has both problems)

Having a simple, reliable, reference "use case" of elasticarray will be helpful for me.

tests/elasticarray/main.c
```diff
@@ -0,0 +1,167 @@
...
+	/* Append some values. */
+	for (i = 0; i < NUM_ELEM; i++) {
+		/* Dynamically allocate item to append. */
+		if ((p = malloc(sizeof(struct dyn))) == NULL)
+			goto err0;
+		if ((p->arr = malloc(sizeof(int))) == NULL)
+			goto err0;
+		p->arr[0] = (int)i;
+
+		/* Append item. */
```
cperciva: I'd say "Append pointer to structure" here -- since `list` is an array of *pointers* rather than an array of structures

gperciva: Right, that leads me back to where I started getting into trouble: I wanted an array of structs, but dynamically allocated.

I think the missing part was to `free(p)` after appending, not at clean-up -- because the contents of `p` had been copied to the list.  See the `REBASE` commit.

BTW, would you recommend the current syntax of
```
struct dyn * p;
...
p = dynlist_get(list, i);
```
or do you think it's cleaner to do
```
struct dyn d;
struct dyn * p = &d;
...
d = *dynlist_get(list, i);
```

cperciva: Sometimes you want an elastic array of pointers.  Sometimes you want an elastic array of structures.  In this particular case, `struct dyn` has a fixed size so it's hard to see why you would want to malloc one, append (a copy of) it, and then free it.

As for `struct dyn * p = dynlist_get(list, i);` vs. `struct dyn d = *dynlist_get(list, i);` -- those do different things.  The former makes `p` point at the structure inside the elastic array; the latter copies the structure out of the array.

gperciva:
Updated PR

cperciva:
Code looks good.  Did you want to create a shell script which runs this test binary?

gperciva:
Oops, I forgot to add it to git.

gperciva:
updated with script included.

PR #168, opened 2019-04-30
sha256: fix pre-processor #if

gperciva:


gperciva:
I'm glad that we caught this before a release!

cperciva:
Not sure I see the issue here.  The C standard says that undefined identifiers in an `#if` preprocessor directive get converted to `0`.

gperciva:
Sure, it's the best type of correct.  But it pops up with a compiler warning, which of course we can disable or ignore... but given that all the other cases use `#ifdef CPUSUPPORT_`, why not do the same here, and avoid the warnings?

cperciva:
Oh, I see.  Silly compilers...

PR #169, opened 2019-05-19
Metabuild: refactor from Makefile.BSD to shell script

gperciva:


gperciva:
Our "creating POSIX Makefiles" has grown beyond what can conveniently be done in a Makefile -- it's
awkward dealing with the special escape characters, changing directory all the time, and being unable to add comments in the middle of the processing.

This PR does not change any functionality; that will be added in other PR(s).

cperciva:
This seems to have a lot of unnecessary subshells.  `foo | bar | baz >> blah` doesn't need to be in a subshell.

I would also abstract `print FOO= && ${BSDMAKE} -V FOO` into a `copyvariable` function.

cperciva:
I absolutely agree in principle with turning this into a shell script, though.  We should have done this a long time ago.

gperciva:
`copyvar` is kind-of done in https://github.com/Tarsnap/libcperciva/pull/170/commits/388bb539821c56b7b59575b5eda71141e71e6d96#diff-26187cc46bdcdfca14c49ecc7e89395cR17
but I suppose that we could start off with `copyvar`, then changing it to `copyvar_if_nonempty` in that PR.

I'll clear up those subshells (inherited from the Makefile version).

gperciva:
Updated PR

release-tools/metabuild.sh
```diff
@@ -0,0 +1,100 @@
...
+cd ${D}
+SUBDIR_DEPTH=`${MAKEBSD} -V SUBDIR_DEPTH`
+
+# Set up *-config.h so that we don't have missing headers
+rm -f ${SUBDIR_DEPTH}/cpusupport-config.h
+touch ${SUBDIR_DEPTH}/cpusupport-config.h
+
+copyvar() {
+	var=$1
+	printf "$var=$(${MAKEBSD} -V $var)\n" >> $OUT
```
cperciva: I would spell this as

```
printf "%s=" "$var" >> $OUT
${MAKEBSD} -V "$var" >> $OUT
```

to avoid ugly subshells (and also the risk of multiple-shell-expansion issues, although those are *hopefully* not likely to arise in this context).

release-tools/metabuild.sh
```diff
@@ -0,0 +1,100 @@
...
+	CPP_ARGS_FIXED="-std=c99 -DCPUSUPPORT_CONFIG_FILE=\"cpusupport-config.h\" -I${SUBDIR_DEPTH} -MM"
+	OUT_CC_BEGIN="\${CC} \${CFLAGS_POSIX} ${CFLAGS_HARDCODED}"
+	OUT_CC_MID="-I${SUBDIR_DEPTH} \${IDIRS} \${CPPFLAGS} \${CFLAGS}"
+
+	# Generate build instructions for each object
+	for F in $OBJ; do
+		S=`${MAKEBSD} source-${F}`
+		CF=`${MAKEBSD} cflags-${F}`
+		IDIRS=`${MAKEBSD} -V IDIRS`
+		echo `${CPP} ${S} ${CPP_ARGS_FIXED} ${IDIRS} -MT ${F}` | \
```
cperciva: I don't understand why this is
```
echo `foo` | bar
```
rather than just
```
foo | bar
```
?

gperciva: That's left-over from what you wrote in `Makefile.BSD`, so don't look at me!  :P

gperciva: hmm.  Adding the `echo` combines all the lines together -- the `${CPP}` helpfully wraps lines.  Granted, there's other ways of combining multiples lines in a posix sh.

cperciva: Looks to me like you introduced that here:
https://github.com/Tarsnap/spiped/commit/b99a843e40a27afd20cf36638a8f533554d8dbf7#diff-e541a5d9bd0f63955b28217432f983eb

;-)

I don't know, maybe using `echo` is cleaning up the whitespace?  Try changing it and see if things break.

release-tools/metabuild.sh
gperciva: fixed in a REBASE commit.

release-tools/metabuild.sh
gperciva: Ok.  In a REBASE commit, I've replaced:
- 1 `echo` and 1 `sed`

with:
- 2 `sed`s and 2 `tr`s.

Umm, enjoy?

cperciva: Yeah, I guess it's a bit ugly.  But I still think this is better than "subshell inside echo command"

cperciva:
LGTM.  I assume you've tested that it produces identical Makefiles?

gperciva:
Rebased, ready for merge.  No change to `Makefile`s on Linux or FreeBSD.

release-tools/metabuild.sh
```diff
@@ -30,18 +30,31 @@ add_makefile_prog() {
...
-	    done ) >> $OUT
+		# Get the build instructions, then remove newlines, condense
+		# multiple spaces, remove line continuations, and replace the
+		# final space with a newline.
+		${CPP} ${S} ${CPP_ARGS_FIXED} ${IDIRS} -MT ${F} |	\
+		    tr '\n' ' ' |					\
+		    tr -s ' '	|					\
+		    sed -e 's| \\ | |g' |				\
+		    sed -e 's| $||g'
+		printf "\n"
```
gperciva: Last-minute change here: on linux, `sed -e 's| $|\n|g'` does what you'd expect.  But FreeBSD is stricter about `sed` being a line editor:
```
$ man sed
...
Sed Regular Expressions
...
     2.   The escape sequence \n matches a newline character embedded in the
          pattern space.  You cannot, however, use a literal newline character
          in an address or in the substitute command.
```

Simple enough to add a `printf "\n"` instead.

gperciva: And yes, I double-checked the identical-ness of `Makefile`s after making this change.

PR #170, opened 2019-05-19
Metabuild: support non-code directories

gperciva:


gperciva:
This allows us to have a directory which doesn't compile any program.  Primarily intended for test code, such as kivaloo's `tests/kvlds-dump`.

This is based on #169, and I'd like to get that merged before having a final copy of this PR.  But you can see that it produces a nice `tests/buildnothing/Makefile`.

gperciva:
Update PR, candidate for merge.

tests/buildnothing/Makefile.BSD
```diff
@@ -0,0 +1,4 @@
+test:
+	echo "Exciting Makefile, huh?  Probably run a script here?"
```
cperciva: This should be `echo "SUCCESS"`, I think?

Anything which is hooked into the tests should behave like a test...

gperciva: well, it's not being run by `make test` (which is now `tests/test_libcperciva.sh`, so I figured it didn't matter.  But ok.

gperciva:
Updated PR

PR #171, opened 2019-05-20
Includes

gperciva:


events/events_network.c
```diff
@@ -1,4 +1,5 @@
 #include <sys/select.h>
+#include <sys/time.h>
```
cperciva: Not needed; POSIX says `The <sys/select.h> header shall define the timeval structure`

gperciva: Oops.  I always forget about that one.  I did double-check that it's not in `<time.h>`.

gperciva:
Rebased to remove the `<sys/time.h>`.

PR #175, opened 2019-06-05
tests/elasticarray: free memory upon error

gperciva:

PR #176, opened 2019-06-07
Elasticarray freeitems

gperciva:


gperciva:
This PR depends on #175.

I found myself writing the same functionality over and over, so I thought it would be nice to add it to the library.  (Actually, in a not-yet-uploaded PR, I added a `stringlist.{c,h}` just so I could write `stringlist_extra_free(list)` and be done with it.)

I'm only 60% happy with the name of this function, but I think the intent is clear.

cperciva:
I don't like this.  If you want an `elasticarray_iter` function, that would make sense; but I don't see any reason to combine that with the `free` special case.

gperciva:
> I don't see any reason

```
elasticarray_iter(list, freefunc);
elasticarray_free(list);
```
vs.
```
elasticarray_free_items(list, freefunc);
```

?  That's a reason!

But ok, your call.  I'll rework this into `_iter`.

PR #178, opened 2019-06-07
Elasticarray iter

gperciva:


datastruct/elasticarray.h
```diff
@@ -159,6 +166,13 @@ int elasticarray_exportdup(struct elasticarray *, void **, size_t *, size_t);
 		return (rec);						\
 	}								\
 	static inline void						\
+	prefix##_iter(struct prefix##_struct * EA,			\
+	    void(free_fp)(void *))					\
```
cperciva: This should be a `void(free_fp)(rectype *)` function pointer, with appropriate casting so that `elasticarray_iter` gets the `void (*)(void *)` function pointer type it expects.

gperciva:
Updated with REBASE commits.  I think this is what you wanted?

cperciva:
Yes, that's right.  This way users of ELASTICARRAY_DECL don't have to cast to/from `void *` anywhere, and they'll get errors if they accidentally have a `foo *` when they need a `foo **` or vice versa.

gperciva:
Yes, I noticed that when I had to update the test.  :)

Updated PR and ready for merge.

gperciva:
If #183 ("The value ${reclen} must be positive") is merged, this PR should be modified accordingly.

datastruct/elasticarray.c
```diff
@@ -255,6 +255,20 @@ elasticarray_get(struct elasticarray * EA, size_t pos, size_t reclen)
 	return ((uint8_t *)(EA->buf) + pos * reclen);
 }
 
+/**
+ * elasticarray_iter(EA, reclen, fp):
+ * Run the ${fp} function on every member of the array.
+ */
+void
+elasticarray_iter(struct elasticarray * EA, size_t reclen, void(fp)(void *))
```
cperciva: I think I normally spell the last parameter `void (* fp)(void *)`.  This might be a case like arrays where the C language infers a pointer even when not explicit... but let's be consistent anyway.

datastruct/elasticarray.h
```diff
@@ -77,6 +77,12 @@ int elasticarray_truncate(struct elasticarray *);
  */
 void * elasticarray_get(struct elasticarray *, size_t, size_t);
 
+/**
+ * elasticarray_iter(EA, reclen, fp):
+ * Run the ${fp} function on every member of the array.
+ */
+void elasticarray_iter(struct elasticarray *, size_t, void(void *));
```
cperciva: And I think this would be `void (*)(void *)`.

gperciva:
PR updated, ready for merge.

datastruct/elasticarray.h
```diff
@@ -160,6 +168,13 @@ int elasticarray_exportdup(struct elasticarray *, void **, size_t *, size_t);
 		return (rec);						\
 	}								\
 	static inline void						\
+	prefix##_iter(struct prefix##_struct * EA,			\
+	    void(free_fp)(rectype *))					\
```
cperciva: `void (* free_fp)(rectype *)`, yes?

gperciva:
sigh, oops again.  Updated?

PR #177, opened 2019-06-07
elasticarray: remove obsolete comment

gperciva:

PR #179, opened 2019-06-07
elasticarray: sanity checks to avoid div by 0

gperciva:


gperciva:
You commented https://github.com/Tarsnap/libcperciva/pull/177#issuecomment-499741074 that I might as well add an `assert`.  This is one instance that I'm not at all certain we want to bother -- reclen is normally taken care of by `ELASTICARRAY_DECL`, so things like `SIZE_MAX / reclen` could be checked at compile-time, not run-time.

Even if you like (some) of these, I'd suggest that we don't bother with the checks for files that called often, such as `_resize` and `_append`.  And maybe not even bothering for `_getsize`.

... right, I'd need to `#include <assert.h>`.  I vaguely recall asking about that before, and you said that you didn't want it in elasticarray because of the speed penality.

gperciva:
I can't find any instance of `elasticarray_init()` in our projects (other than inside `elasticarray.c`), so I'm even more convinced that we should check `reclen > 0` in the `_init()`, but not all the other functions (which actually do need `reclen > 0` in order to avoid a divide-by-zero).

I've removed the commit that added extra sanity checks; the one in `_init()` should be sufficient.

gperciva:
Also, if this PR is merged: should I add "The value of ${reclen} must be positive" to every function which contains `SIZE_MAX / reclen`?

cperciva:
> We don't actually need (reclen > 0) in this function

We do, because it calls elasticarray_resize.

gperciva:
Updated PR with corrected git comment.

PR #180, opened 2019-06-07
Match function comments between h and c files

gperciva:


gperciva:
Finally wrote that script to automatically compare function documentation.  Some debugging on that still ongoing.

Almost all of these changes are adding some text to match the more verbose option.  Generally adding some extra guarantee to the function documentation.

If you want, I could try to dig up the exact commits where the function docs diverged.  The changes look simple enough to me that they could just be accepted, though.

gperciva:
This PR is blocked on the crypto_entropy 2^16 question.

gperciva:
I moved the change to `crypto_entropy` to a separate PR (#182).  This one is now a merge candidate again.

PR #181, opened 2019-06-09
Tests: expected exitcode

gperciva:


gperciva:
This isn't a complicated thing, but it keeps on cropping up, so no point repeating the same 7 lines every time.  (Also, in some places I've forgotten to check for the valgrind exit code.)

gperciva:
There's more substantial uses of `expected_exitcode()` in https://github.com/Tarsnap/scrypt/pull/181.

PR #183, opened 2019-06-12
elasticarray: add comments about reclen being positive

gperciva:


gperciva:
Do you want these everywhere?  Or should we only specify this for `_init()` and remove it from other functions?

cperciva:
Might as well include the comment anywhere that we need `reclen > 0`.  So I think everywhere except `elasticarray_get`.

gperciva:
Everywhere that includes `reclen`, other than `elasticarray_get()`.  This PR is ready for merging.

PR #187, opened 2019-08-09
Improve events testing

gperciva:


cperciva:
LGTM, but as I was looking at this I wondered why `test_interrupt_run` has `while ((ret = events_run()) { ...`.

The function `events_run` is not required to handle events forever -- in fact at one point I had an implementation which would return after a hard-coded number of events were processed -- and will return `0` if nothing breaks.

This should probably be looping while `events_run` returns `0`, with a `if (done) break;` to escape the loop at the appropriate time.

I'll leave this PR unmerged for now in case you want to add another commit to it...

gperciva:
Right, at first I thought that adding an `events_run_done` would jeopardize actually testing `events_interrupt`, but when I deliberately break that function I see that the test breaks appopriately.

gperciva:
... shall I go ahead and rebase this?

cperciva:
Still not right.  `events_run` can return zero whenever it wants.  I think the code needs to look like

```
while ((ret = events_run()) == 0) {
    if (events_run_done)
        break;
}
if (ret) {
    /* handle error case here */
}
```

gperciva:
Judging from how `events_run()` is used in kivaloo, I think it's more idiomatic to wrap it in a `do ... while` and have the error-handing inside the loop.

cperciva:
Ok, we could do it that way.  But anything nonzero from `events_run` is an error, not just `-1`.  (The events code itself will return `-1` if it has an internal error, but we don't specify what error codes callbacks might return.)

cperciva:
LGTM

gperciva:
Rebased, ready for merge.

PR #188, opened 2019-08-09
Run events_shutdown() automatically

gperciva:


gperciva:
This includes PR #187.

events/events.c
```diff
@@ -187,6 +198,13 @@ events_run(void)
 {
 	int rc;
 
+	/* Register the atexit function (if necessary). */
+	if (!registered_atexit && atexit(events_atexit)) {
+		warn0("atexit(events_atexit)");
```
gperciva: I'm not 100% certain that we want to add a dependency on `warnp.h` to the event layer.

gperciva:
I know that we discussed running `events_shutdown()` automatically at some point, but I can't recall where.

Personally, I'm not convinced that it's worth the extra complexity... it's simple enough to call `events_shutdown()` explicitly.  But at least we have something to examine now.

cperciva:
You can't do this from `events_run` and `events_spin` since that wouldn't clean up if a program e.g., created and then cancelled a timer event without ever waiting for it.

Looks to me like we need two atexits, one for events_network and one for events_timer.  But they can be registered from the respective initialization code.

cperciva:
We need to keep a stub function for `events_shutdown` for compatibility, BTW.

cperciva:
Oh, and no need to `warnp` if atexit fails; the only reason it should fail is ENOMEM.

gperciva:
What if there's already 32 [*] other registered `atexit()` functions?

[*] or more, depending on the implementation.

cperciva:
That's the case where `atexit` might fail with ENOMEM.

gperciva:
ENOMEM is a non-standard FreeBSD addition.  The C spec and POSIX don't define any errors.
https://pubs.opengroup.org/onlinepubs/9699919799/functions/atexit.html

But yes, it will be nice to remove the `warnp`.

cperciva:
FreeBSD documents ENOMEM, but POSIX explicitly allows functions to return errors not listed -- the requirement is only that if a listed error occurs, errno will be set to the specified value (and not something else).

But in terms of implementation: The only plausible way for atexit to fail is if it can't allocate memory; hence ENOMEM.

gperciva:
New approach, but I'm not confident about how you want to approach deprecating functions, so I've marked it `[draft]`.  AFAIK the code is ok.

events/events_network.c
```diff
@@ -86,6 +95,10 @@ init(void)
 	fds = NULL;
 	fds_alloc = nfds = fdscanpos = 0;
 
+	/* Clean up the socket list at exit. */
+	if (atexit(events_network_atexit))
```
gperciva: I just realized that we could make this `atexit(events_network_shutdown)` instead of having the 1-line function `events_network_atexit`.  Would that be preferable?  I think my original thought was that `_atexit` functions would obviously be run via `atexit`, but in retrospect `_shutdown` has almost the same implication.

PR #189, opened 2019-08-13
style: fix include guard

gperciva:


gperciva:
"Hmm, I wonder if it's worth automating checking for style guideline X?"

"Yes.  The answer is always yes."

PR #191, opened 2019-08-13
Add PARSENUM_BASE

gperciva:


gperciva:
Checking for `ULLONG_MAX` in https://github.com/Tarsnap/kivaloo/pull/149 seemed too complicated, so I figured that I should play with the c preprocessor by adding PARSENUM_BASE().

gperciva:
I didn't add many tests; I figured that since it re-used the existing infrastructure for PARSENUM (where possible), they weren't as necessary.

More detail about why I thought this would be useful: the lengths in chunked http don't include `0x`, so we can't use the regular `PARSENUM` for it.  I considered copying those strings into `"0x%s"` so that kivaloo could still use the existing `PARSENUM`, but I didn't think it would be hard to add support for different bases, so I went with that solution.  Also, I'm not certain that we won't find base-16-parsenum useful in other situations.

tests/parsenum/main.c
```diff
@@ -176,6 +212,17 @@ main(int argc, char * argv[])
 	TEST2_FAILURE("0XfffffFFFF", uint32_t, ERANGE);
 	TEST2_FAILURE("-1", uint32_t, ERANGE);
 
+	/* Handle alternate bases */
+	TEST3_SUCCESS("11", size_t, 17, 16);
+	TEST3_SUCCESS("11", size_t, 11, 0);
+	TEST3_FAILURE("122223333", uint32_t, ERANGE, 16);
+	TEST3_SUCCESS("122223333", uint32_t, 122223333, 0);
+	TEST3_FAILURE("ga", size_t, EINVAL, 16);
```
cperciva: Maybe add a base-17 success for "ga"?

cperciva:
I'm a bit confused by the order of the commits.  It looks like the first commit came second for some reason?

I'll merge this if you fix the ordering.  And maybe add a test where "ga" gets successfully parsed in base 17 because hey, why not.

gperciva:
Added the "ga"_17 test.

I definitely have the rebase before adding PARSENUM.  But github is definitely showing them in the wrong order.  *sigh*.  I'll try a bunch of pushes and investigations and will let you know.

gperciva:
... I see.  It was displaying them chronologically, rather than in the actual order of commits in the tree.

https://github.com/isaacs/github/issues/386

LOL from that page:
>> I think it is git default so nobody just shouted loud enough for somebody to notice.
>
> No, this is not Git’s fault. There are no options you can pass to git log or similar to get an ordering as broken as the one GitHub uses on the PR page

cperciva:
Wow, that's pretty impressively stupid of github.

PR #194, opened 2019-08-31
events: events_run() may return 0 at any time

gperciva:


gperciva:
I don't think that this is clear from the existing API description.

events/events.c
```diff
@@ -84,7 +84,8 @@ doevent(struct eventrec * r)
  * be run.  If any event function returns a non-zero result, no further
  * events will be run and said non-zero result will be returned; on error,
  * -1 will be returned.  May be interrupted by events_interrupt, in which case
- * 0 will be returned.
+ * 0 will be returned.  May return 0 at any time; this does not mean that
```
cperciva: I think a better way to phrase this is "If there are runnable events, events_run is guaranteed to run at least one; but it may return while there are still more runnable events."

gperciva:
Updated as suggested.

PR #193, opened 2019-08-29
Switch to PARSENUM_EX

gperciva:

PR #195, opened 2019-09-03
Prep for PARSENUM_EX

gperciva:


gperciva:
This prepares the `PARSENUM_BASE` macro for the upcoming rename to `PARSENUM_EX`.  There's a few bits I'm uncertain about, on which I'll comment.

util/parsenum.h
```diff
@@ -107,26 +108,38 @@ _Pragma("clang diagnostic pop")
  * For an unsigned integer variable ${x}, this can also be invoked as
  * PARSENUM_BASE(x, s, base), in which case the minimum and maximum values are
  * set to the limits of the unsigned integer type.
+ *
+ * For a floating-point variable ${x}, the ${base} must be 0.
```
gperciva: I don't know if this comment should be worked into the previous paragraph, or if it can stay here by itself.

util/parsenum.h
```diff
@@ -107,26 +108,38 @@ _Pragma("clang diagnostic pop")
...
  */
-#define PARSENUM_BASE3(x, s, b)						\
+#define PARSENUM_BASE3(x, s, base, _define_str)				\
```
gperciva: If we want the eventual `PARSENUM_EX` to handle both `PARSENUM` and `PARSENUM_EX`, and if we want the `ASSERT_FAIL()` line to give an accurate error message, then this macro needs to know whether it's been called as `PARSENUM` or `PARSENUM_EX`.

gperciva: PS: I used the prepended `_` to indicate that this is not a variable which the user supplies.

util/parsenum.h
```diff
@@ -141,7 +154,7 @@ _Pragma("clang diagnostic pop")
 	)
 
 /* Magic to select which version of PARSENUM_BASE to use. */
-#define PARSENUM_BASE(...)	PARSENUM_BASE_(PARSENUM_BASE_COUNT(__VA_ARGS__))(__VA_ARGS__)
+#define PARSENUM_BASE(...)	PARSENUM_BASE_(PARSENUM_BASE_COUNT(__VA_ARGS__))(__VA_ARGS__, "PARSENUM_BASE")
```
gperciva: The `_define_str` is added here.  When `PARSENUM` is switched to call `PARSNEUM_EX{3,5}`, those lines will set it to the relevant string.

gperciva:
I've updated the "full" PR in #193 if you want to see how these changes will be used in the whole thing.

util/parsenum.h
cperciva: This is fine.

cperciva:
Looks ok.  Maybe `define_str` would make more sense as `define_name`?

cperciva:
But I'm happy to merge this as it is if you don't want to change that.

gperciva:
I'll make the `_define_str` -> `_define_name` change.

I just noticed that sometimes I use parentheses around `base`:
```
                        ((*(x)) = parsenum_unsigned((s), 0, (*(x)),     \
                            (*(x)), (base))) :                          \
```

and other times I don't:

```
                        ((*(x)) = parsenum_signed((s),                  \
                            (*(x) <= 0) ? (min) : 0,                    \
                            (*(x) <= 0) ? (max) : 0, base)) :           \
```

I'm guessing that I should always use parentheses, for consistency?

cperciva:
You should always use parentheses in macros, yes.  Not for consistency, but for correctness -- without parentheses, bad things can happen (imagine what happens if `base` has a comma in it, for example).

gperciva:
Fixed the missing parentheses around `(base)`, and changed the internal command name variable to `_define_name`.

PR #196, opened 2019-09-11
tests/elasticarray: append 0 items to an empty list

gperciva:


gperciva:
I'm trying to get clang to stop warning about `elasticarray_append`, so I figured that I should ensure that I didn't accidentally break this corner case.

PR #198, opened 2019-09-14
elasticarray: avoid false warning about copying to NULL

gperciva:


gperciva:
I'm not terribly happy about this fix, but:
- it's the last one that clang's `scan-build` complains about, and
- since it's in elasticarray, I see this warning every time I run clang-scan on any Tarsnap project (other than scrypt).

I think the assert in `resize()` is fair, but I'm annoyed that it needs the `assert` in `elasticarray_append()`.

I tried adding a separate `ANALYZER_ASSERT()` macro which was a no-op unless `__clang_analyzer__` was defined.  It worked, but in retrospect I decided that it was an unduly clunky approach for just one false positive.

PR #200, opened 2019-09-17
Minor fixes for tests

gperciva:


tests/parsenum/test_parsenum_improper_usage.sh
```diff
@@ -5,16 +5,16 @@ rm -f $log
 
 # Check assertion failures, separating lines so that we automatically check
 # that test_parsenum is giving a non-zero exit code.
-( ! ./test_parsenum 1 ) 2>$log
+( set +e; ! ./test_parsenum 1 ) >$log 2>&1
```
gperciva: I'm reconsidering this; it's probably best to replace it with

    ./test_parsenum 1 || true

or even have a separate function to deal with it.  I'll revisit this a bit later.

PR #202, opened 2019-09-20
Misc: events_counter, parsenum_ex, test_json

gperciva:


gperciva:
These commits all involved a change of a `Makefile`, but it seemed silly to put them as separate PRs.  It's a bit more iffy than the previous "misc" PR (which was trivial).

util/sock.c
```diff
@@ -266,12 +267,8 @@ sock_resolve(const char * addr)
...
 	ips[strlen(ips) - 1] = '\0';
 
-	/*
-	 * Parse the port number.  If strtol fails to parse the port number,
-	 * it will return 0; but that's fine since port 0 is invalid anyway.
-	 */
-	p = strtol(ports, NULL, 10);
-	if ((p <= 0) || (p >= 65536)) {
+	/* Parse the port number, allowing for trailing characters. */
+	if (PARSENUM_EX(&p, ports, 1, 65535, 10, 1)) {
```
gperciva: Since we've got this code now, we might as well use it here.  I tested it in spiped (since libcperciva doesn't have any test code that calls `sock_resolve`).

tests/events/Makefile.BSD
```diff
@@ -9,6 +9,7 @@ LIBCPERCIVA_DIR	=	../..
 
 # Main test code
 SRCS	=	main.c
+SRCS	+=	events_counter.c
```
gperciva: I'm working on an "event counter network server" to test `event_network.c`, so it makes sense to share the "event counter" functionality between the two binaries.

util/sock.c
cperciva: Hmm, I don't think we need to support trailing junk.  Are you aware of any case where that support is needed?

gperciva: You're right, we don't need it here.  I was thinking about parsing chunked HTTP headers.

gperciva:
Rebased to not allow trailing characters in sock's port number.

tests/events/main.c
```diff
@@ -56,6 +43,9 @@ static int
 test_interrupt_run()
 {
 
+	/* Reset coutner. */
```
cperciva: "coutner"

tests/events/main.c
```diff
@@ -84,6 +74,9 @@ test_interrupt_spin()
 	int done = 0;
 	int ret;
 
+	/* Reset coutner. */
```
cperciva: "coutner"

tests/events/main.c
```diff
@@ -139,6 +121,9 @@ test_timer()
...
```
cperciva: Wow you really like this typo.

gperciva:
Rebased with typo fixes.  Now working on including my comment spell-checker with the "find style problems" script.

PR #206, opened 2019-09-27
Visualize mpool timing benchmark

gperciva:


gperciva:
This has almost no effect on the normal test.  The only changes are:
- runs & discards 5 tests to "warm up" the cache (instead of 10 times)
- runs the test 11 times (instead of 10 times)
- exits if the benchmark produces a time of 0 micro-seconds.  (This can occur on systems with a relatively poor granularity of `CLOCK_PROCESS_CPUTIME_ID`.)

Everything else in this PR only applies if you run the test manually, with an extra parameter asking for logfiles.  Once this is merged, I have another PR which contains significant changes to how this benchmark is performed (median instead of lowest time, and interleave malloc and mpool runs).  I'm putting that in a separate PR since you'll probably want to look at that more carefully.

FYI, here's the graphical output of this test.

### Ubuntu desktop

log-all

### FreeBSD server

freebsd

cperciva:
I'm confused by "amortized microseconds"  on the Y axis.  Surely it doesn't take up to 60000 microseconds (1/16th of a second!) per malloc?

gperciva:
Err, right.  I forgot to amortize it.

gperciva:
The test now outputs the amortized time to logfiles.  The automatic test suite still calculates the ratios using the non-amortized values, because `sh` is happier with integers than floating point.

median-all

cperciva:
Personally I would make the Y axis nanoseconds rather than microseconds.  But up to you; please tidy up the git history.

gperciva:
Done.  I left it with microseconds because that's what monoclock gives us.  It's easy enough to change later if we want.

PR #207, opened 2019-09-27
Improve reliability of mpool benchmark

gperciva:


gperciva:
This includes #206, which should be merged first.  I think these commits are pretty self-explanatory, but I'll add two graphs of timing a run on Ubuntu to show how useful interleaving / alternating is.

I've manually tweaked the display of the mpool data so that it lines up (Y-axis) with the malloc data.  I want to highlight the (relatively) good correlation between the timing data.  I plotted with lines rather than points so that the spikes would be easier to see.

### Ubuntu, "full benefit" of mpool

interleave-full


### Ubuntu, "no benefit" of mpool

interleave-none

PR #208, opened 2019-09-30
Style nitpicks

gperciva:


gperciva:
These are slightly "iffy" changes.  I wrote a script to be as nitpicky as possible with `.h` files, including a requirement that every top-level "paragraph" begin with a comment [1].  Running it on libcperciva, scrypt, spipe, and kivaloo didn't highlight any important defects, but the result is slightly more pleasing to my eye.

[1] exceptions: `#include` statements and `static inline` functions.

cperciva:
The comments are good, but I don't see any reason to move the blank line from `Makefile.prog` into a printf.

If the issue here is that your style checker doesn't like files ending with blank lines: `Makefile.prog` should be exempt from style checking; to the extent that style is checked, it should be checked in the Makefiles which are generated using it.

gperciva:
Fair enough; rebased accordingly.

PR #209, opened 2019-10-04
Fix events_network interrupt (1)

gperciva:

PR #210, opened 2019-10-04
Fix event_network interrupt (2)

gperciva:


gperciva:
This depends on #209, which should be merged first.

tests/events/network_client/main.c
```diff
@@ -0,0 +1,200 @@
...
+
+static int done;
+
+static int
+poke(struct sock_addr ** sas)
+{
+	int client_sock;
+
+	/* Attempt to connect, but don't complain about any failures. */
+	if ((client_sock = sock_connect(sas)) == -1)
```
gperciva: For this test, it would be ideal if `sock_connect()` could return the reason in errno.  For example, if we get `ECONNREFUSED` from `connect()` then we know that the server shut itself down (as the test is designed to do), whereas anything else would indicate an actual error.

However, I don't think it's worth messing with `sock_connect()` just for this test.

tests/events/network_client/main.c
```diff
@@ -0,0 +1,200 @@
...
+	return (0);
+
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+/* Parse a pid_t; it's a signed integer and there's no PID_MAX. */
+static pid_t
+str_to_pid(const char * str)
```
gperciva: This is yucky, but it's the cleanest solution I could find for reading a pid from the command-line.

Yes, POSIX 2008 [1] includes
> The implementation shall support one or more programming environments in which the widths of blksize_t, pid_t, size_t, ssize_t, and suseconds_t are no greater than the width of type long.

but I don't think that means that we can assume that our current programming environment qualifies.  The paragraph goes on to say:

> The names of these programming environments can be obtained using the confstr() function or the getconf utility.

and we're not checking either of those.

[1] https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html

tests/events/network_client/main.c
cperciva: "Return the reason in errno" is a bit ill-defined.  If we have 2 IP addresses, and one returns ECONNREFUSED and the other returns EHOSTUNREACH, which should we return?

gperciva: Yes, that's another good reason to avoid messing with `sock_connect`.

cperciva:
I feel like I'm missing some context here.  What are these for?  What are you trying to test?

gperciva:
At some point, you commented that network events couldn't be interrupted.  All of this is to support a fix for that, namely this commit:
https://github.com/Tarsnap/libcperciva/pull/211/commits/ae9fa975d397c9da4fedd795577d0ab025cd7ab8
from #211.

This particular PR adds 2 network tests, but only 1 of them works.  The other one is enabled in #211, thereby demonstrating that the fix works.  (Whether or not you like that particular fix is a separate question -- the first stage was to convince myself that it worked.)

cperciva:
Ok, this seems like it's wildly overcomplicated.  What's wrong with having a program which

1. Catches SIGQUIT and calls events_interrupt from the handler;
2. Calls events_spin; and
3. Prints "Exited successfully",

then a shell script which runs
```
./testcode &
sleep 1
kill -QUIT $!
```

gperciva:
I'm not seeing the network part of the program you described?  We already have a number of tests for `event_interrupt()` in `tests/events/main.c`.

My goal in this PR was to produce:
1. a test which failed on current master
2. but which worked after the fix in `events_network.c`.

I'll admit that I was surprised that `repeated_poker()` to worked -- my original plan was that this would be the only test.  Once I understood the problem better, I came up with `long_poke()`, which satisfied those two conditions.  I considered deleteting `repeated_poker()`, but I figured there was no harm keeping it around.  But if that's confusing the issue, I could nuke it.

cperciva:
If we have no events registered, I think we end up in an infinite wait in poll?  So testing with nothing registered should be the same as testing waiting for the network.

Your current tests for interrupt IIRC all have immediate events being handled.

gperciva:
Better solution in #213.

PR #211, opened 2019-10-04
Fix events-network interrupt (3)

gperciva:


gperciva:
This includes #209 and #210, which should be merged first.  The whole chain is designed to isolate the non-test-code changes to a single short PR (this one), which demonstrates the fix.

PR #212, opened 2019-10-07
Improve style of sh scripts

gperciva:


gperciva:
First batch of fixes arising from shellcheck.  Nothing special; just slight improvements to portability and clarity of the scripts.  Oh, and fixing an embarrassing mistake involving `-eq` vs. `=`.

I was *very* tempted to write "am I groot?" instead of "am I root?" in the final commit message.

PR #213, opened 2019-10-11
Events interrupt network

gperciva:


gperciva:
As per your comment https://github.com/Tarsnap/libcperciva/pull/210#issuecomment-539772224, this is a much simpler method of verifying that we can now interrupt the `poll()`.

events/events_internal.h
```diff
@@ -27,12 +29,13 @@ void events_freerec(struct eventrec *);
 struct eventrec * events_immediate_get(void);
 
 /**
- * events_network_select(tv):
+ * events_network_select(tv, interrupt_requested):
  * Check for socket readiness events, waiting up to ${tv} time if there are
  * no sockets immediately ready, or indefinitely if ${tv} is NULL.  The value
- * stored in ${tv} may be modified.
+ * stored in ${tv} may be modified.  If ${interrupt_requested} is non-zero
+ * and a signal is sent, exit.
```
cperciva: s/sent/received/

gperciva: I also assume s/is/was

events/events_internal.h
```diff
@@ -32,8 +32,8 @@ struct eventrec * events_immediate_get(void);
  * events_network_select(tv, interrupt_requested):
  * Check for socket readiness events, waiting up to ${tv} time if there are
  * no sockets immediately ready, or indefinitely if ${tv} is NULL.  The value
- * stored in ${tv} may be modified.  If ${interrupt_requested} is non-zero
- * and a signal is sent, exit.
+ * stored in ${tv} may be modified.  If ${*interrupt_requested} is non-zero
```
gperciva: not certain about this -- should we "dereference" the pointer variable in this comment?  I mean, technically any non-NULL `volatile sig_atomic_t *` will be non-zero (unless we're on a really weird platform).

gperciva: Answering my own question: yes, `${*arg}` is the style used on the tarsnap client.

cperciva: Right, the contents of `${}` should be expression-like, so if we want to dereference something we should say that.

events/events_internal.h
cperciva: hmm, I think "is" makes more sense -- we're saying if a signal arrives *while we're waiting*, right?

gperciva: ok, switched back to "is".  Technically, we'd also exit if the signal arrives while that function is blocked on `poll`, but I'm guessing that's too detailed to mention (and/or can be considered an implementation detail).

cperciva:
LGTM

gperciva:
Rebased, ready for merge.

PR #214, opened 2019-10-11
Misc fixes to tests

gperciva:


gperciva:
Enabling valgrind on the AWS test shows uninitialized memory on FreeBSD, but I'm not convinced that this isn't simply due to the old / unsupported port (last commit in 2016).  There's no memory problems on Linux, and the problem is reported in libc rather than our own code, so I think we can move forwards.

Since you'll ask, the FreeBSD issue is:
```
	char buf[64];
	FILE * f;
	volatile size_t len;	/* Force the compiler to record this value. */

	if ((f = fopen("Makefile", "r")) == NULL) {
		fprintf(stderr, "Can't read Makefile to test fgets\n");
		return;
	}
	fgets(buf, sizeof(buf), f);
	len = strlen(buf);
```

It also occurs if the last line is `printf("%s", buf);`.  Warning goes away if we `memset(buf, 0, 64);` before the `fgets()`.

PR #215, opened 2019-10-12
Style doc

gperciva:


gperciva:
"78 chars" was an old commit that was in Tarsnap, but didn't make its way to libcperciva.

PR #216, opened 2019-10-13
style: whitespace in parsenum.h

gperciva:


util/parsenum.h
```diff
@@ -121,8 +121,8 @@ _Pragma("clang diagnostic pop")
 			    (min) <= 0 ? 0 : (min),			\
 			    (uintmax_t)(max), *(x), (base),		\
 			    (trailing))),				\
-			((((max) < 0) && (errno == 0)) ?		\
-			    (errno = ERANGE) : 0)),			\
+				((((max) < 0) && (errno == 0)) ?	\
```
gperciva: Based on the parentheses, these two lines belong to the `parsenum_unsigned()` branch, so I think they should be indented.  But maybe you thought it was easier to read in the original format?  This indentation was present in the very first parsenum commit: https://github.com/Tarsnap/libcperciva/commit/0461d4159cb701700568e937402f104d5a309fd9#diff-6c2e648c26c72487e2702c8daeea8629R30

util/parsenum.h
```diff
@@ -117,12 +117,10 @@ _Pragma("clang diagnostic pop")
 			    (*(x) <= 0) ? (min) : 0,			\
 			    (*(x) <= 0) ? (max) : 0, (base),		\
 			    (trailing))) :				\
-			(((*(x)) = parsenum_unsigned((s),		\
+			((*(x)) = parsenum_unsigned((s),		\
 			    (min) <= 0 ? 0 : (min),			\
 			    (uintmax_t)(max), *(x), (base),		\
 			    (trailing))),				\
-				((((max) < 0) && (errno == 0)) ?	\
```
gperciva: Removing these lines has no effect on the tests at all.  I'm guessing that the idea was to warn about the misuse of arguments in
```
    if (PARSENUM(&uint, str, 0, -100)) {
```

but I wonder if it's really necessary.  It complicates things for compatibility with gcc due to their bug of not allowing `_Pragma` everywhere that it should be [1], which means that we can't disable this specific warning for `PARSENUM` in the way that we've done for clang.

[1] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78657 .

cperciva: Not just that case; we also need this for `PARSENUM(&uint, "0", -200, -100)`.  Which isn't quite as silly as it looks, since the range bounds might not be compile-time constants.

I'm inclined to say that we need to keep this -- and should probably add a test for it, too.

PR #217, opened 2019-10-14
tests: add crypto-entropy

gperciva:


tests/14-crypto-entropy.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+### Constants
+
+# The default valgrind in Travis-CI (currently Ubuntu xenial, which has
```
gperciva: This is why all we had all those RDRAND failures in the past few days -- the "Illegal instruction" was from old valgrind failing to deal with RDRAND, not from the instruction itself.

gperciva:
Added for discussion; I'm not certain about the best solution yet.

History: I noticed that the scrypt repository still used Ubuntu trusty 14.04 for testing.  I tried updating that to xenial, but at least one of the tests failed (when we fail a test, we don't bother running the other tests).  I tracked it down to rdrand, then worked on adding this test to the libcperciva code, because IMO any problem in libcperciva should have a failing test here, not a failing test in a different repository.
(We already had a `tests/crypto_entropy`, but I'd failed to add that to the `make test` suite.)

Having reproduced the failure in libcperciva, it took me a while to figure out that the "Illegal instruction" arose due to an old version of valgrind failing to understand RDRAND, as opposed to that cpu instruction itself failing.  But at least I know what's up now.

At this point I see three possibilities:
1) use xenial 16.04 (the current default for Travis-CI), but disable any memory test for code which uses RDRAND.
    Unfortunately, that includes any `scrypt enc` command, which would cut out a lot of the scrypt tests.

2) use bionic 18.04 (which is officially supported by Travis-CI: https://docs.travis-ci.com/user/reference/bionic/ ).

3) postpone any decision until we find out what github actions are like (general availability is scheduled for Nov 13; I could sign up for the beta if we're particularly curious).

gperciva:
Update: I've decided that it's not worth gimping the test suite to work around bugs in old versions of valgrind.  Let's go with option 2 from the above list: switch to Ubuntu bionic 18.04.  I've added a comment about this to `.travis.yml` so that I won't forget.

Once this is merged, I'll do the same for the scrypt repository.

gperciva:
It would be nice to rebase this on top of master after you've merged #218.

PR #218, opened 2019-10-15
Travis-CI: switch to Ubuntu bionic 18.04

gperciva:


gperciva:
Ubuntu xenial 16.04 comes with valgrind-3.11, which can't handle the
RDRAND instructions and aborts with "Illegal instruction".  This is not a
problem with Ubuntu bionic 18.04, which has valgrind-3.13.

PR #219, opened 2019-10-15
style: fix punctuation of ${arg} in comments

gperciva:


crypto/crypto_verify_bytes.c
```diff
@@ -5,8 +5,8 @@
 
 /**
  * crypto_verify_bytes(buf0, buf1, len):
- * Return zero if and only if buf0[0 .. len - 1] and buf1[0 .. len - 1] are
- * identical.  Do not leak any information via timing side channels.
+ * Return zero if and only if ${buf0[0 .. len - 1]} and ${buf1[0 .. len - 1]}
```
gperciva: I'm not certain if `_len_` is acceptable here.  Unfortunately adding `${}` around both `len` makes the line exceed 80 chars, but it would be annoying to add a linebreak in `${buf1[0 .. len - 1]}`.

At the moment I've hard-coded an exception to the style script for this case, particularly because I'm not certain about the guideline for argument names appearing within an `${ }` context.

crypto/crypto_aesctr.c
```diff
@@ -44,7 +44,7 @@ crypto_aesctr_init(const struct crypto_aes_key * key, uint64_t nonce)
 
 /**
  * crypto_aesctr_stream(stream, inbuf, outbuf, buflen):
- * Generate the next ${buflen} bytes of the AES-CTR stream and xor them with
+ * Generate the next ${buflen} bytes of the AES-CTR ${stream} and xor them with
```
cperciva: In this case, "stream" was being used as a noun rather than as a name.  Perhaps it should be "the AES-CTR stream ${stream}" and in _free "Free the AES-CTR stream ${stream}." though.

crypto/crypto_verify_bytes.c
cperciva: It's ok.  Basically ${foo} means "stuff in here has pseudo-C syntax"; it applies to expressions (or things resembling them, like the above array bounds) rather than to individual variables.

util/humansize.c
```diff
@@ -7,7 +7,7 @@
 
 /**
  * humansize(size):
- * Given a size in bytes, allocate and return a string of the form "<N> B"
+ * Given a ${size} in bytes, allocate and return a string of the form "<N> B"
```
cperciva: As with "stream", the word "size" here is a noun, not a name.

PR #221, opened 2019-10-26
Import from spiped

gperciva:


gperciva:
Minor updates from spiped.

Issue #222, opened 2019-10-29
our CRC32C only support 64-bit


gperciva:
We're using `_mm_crc32_u64()`, which of course can't be used on a 32-bit platform.

I see two solutions:
1) Document it as-is.  We only use the CRC32C instructions in kivaloo, which I'm pretty certain we're only using on 64-bit platforms.  So we could just document that our implementation of CRC32C is only aimed at 64-bit systems, and possibly even change

       Checking if compiler supports X86 CRC32 feature...

   into

       Checking if compiler supports X86 CRC32 64-bit feature...

   to avoid confusion.

2) change the code (and `cpusupport/`) to support `_mm_crc32_u32()`.

I'm leaning towards the first option, but it's your call.

cperciva:
For now, just document it.  I wouldn't change the "Checking if..." message since the format there is currently consistent with the preprocessor macro and compiler flag variable names.

At some point we might want to support using `_mm_crc32_u32` -- possibly even using a single _mm_crc32_u32 after a series of _mm_crc32_u64 calls if we're processing 4-7 mod 8 bytes -- but let's leave that until we actually have a use for that functionality.

gperciva:
Right, manually changing the "Checking if..." message would be silly.  But what about changing the preprocessor macro?  I started trying to add a "64-bit warning" to the function docs, but then I thought it would be a lot more clear just to rename the macro to `CPUSUPPORT_X86_CRC32_64`.  And assorted follow-through renamings, like `cpusupport_x86_crc32_64()`.

And yes, I've double-checked that `tests/crc32` still results in using hardware CRC32C.  (After fixing a few typos along the way.)

PR #223, opened 2019-10-29
CRC32 only 64bit

gperciva:


gperciva:
This is now my favourite solution for #222.

PR #224, opened 2019-10-30
build: allow `make -f Makefile.BSD`

gperciva:


gperciva:
This is the last step in unifying the build system among all non-autotools projects; merging these `Makefile.inc` bits from kivaloo.  See also
https://github.com/Tarsnap/kivaloo/pull/64#issuecomment-449625403
https://github.com/Tarsnap/kivaloo/pull/166

PR #225, opened 2019-10-31
tests/valgrind: false positive with FreeBSD's strlen

gperciva:


gperciva:
Same as https://github.com/Tarsnap/spiped/pull/217.

PR #227, opened 2019-11-12
readpass_file

gperciva:


gperciva:
Draft PR to discuss exactly what this function should do.  Things like:
- is the maximum length of 256 chars ok?
- should we strip a final `\n` and `\r\n`?
- is it ok to bail if there's any other `\n` or `\r` characters in the line?
- should we support an empty file / file only containing a newline, or not?

I found it useful to construct a whole bunch of `good-*.pass` and `bad-*.pass` files to clarify a bunch of edge cases.

gperciva:
TODOs:
- add a test containing exactly the maximum number of chars.
- add a test containing exactly 1 more than the maximum number of chars.
- check on more OSes to see if the `fopen: No such file or directory" message is consistent, or whether I should change that test to a `grep` instead.

util/readpass_file.c
```diff
@@ -0,0 +1,106 @@
...
+readpass_file(char ** passwd, const char * filename)
+{
+	struct stat sb;
+	FILE * fp;
+	char passbuf[MAXPASSLEN];
+	size_t len;
+	size_t eol_pos;
+
+	/* Stat the file. */
+	if (stat(filename, &sb)) {
```
cperciva, line 33: There's a race condition here if the file is replaced between the stat call and the fopen.  I would fopen first and then use fstat(fileno(f)).

cperciva: ... or for that matter just fopen and fgets, without needing to stat the file at all.

gperciva:
Switched to fgets, rebased the whole thing.

util/readpass_file.c
```diff
@@ -0,0 +1,78 @@
...
+ * Read a passphrase from ${filename} and return it as a malloced
+ * NUL-terminated string via ${passwd}.  Produce an error if the file is 2048
+ * characters or more, or if it contains any newline \n or \r\n characters
+ * other than at the end of the file.  Do not include the \n or \r\n
+ * characters in the passphrase.
+ */
+int
+readpass_file(char ** passwd, const char * filename)
+{
+	FILE * fp;
```
cperciva, line 23: Why `fp`?  Normally we just call a file `f`...

util/readpass_file.c
```diff
@@ -0,0 +1,78 @@
...
+			warnp("fread(\"%s\")", filename);
+			goto err2;
+		} else {
+			/* We have a 0-byte password. */
+			passbuf[0] = '\0';
+		}
+	}
+
+	/* Bail if there's a second line. */
+	if ((fgets(secondline, 2, fp)) != NULL) {
```
cperciva, line 45: Why not `fgetc`?  That way you don't need a second buffer.

util/readpass_file.c
gperciva: file pointer?  I hate one-character variables other than `i` and `j`.

... that said, I admit that you like `f` so I should follow that.

gperciva: Also, easier for searching -- not many words contain `fp`, but tons contain `f`.

util/readpass_file.c
gperciva: done

util/readpass.h
```diff
@@ -20,4 +20,14 @@
  */
 int readpass(char **, const char *, const char *, int);
 
+/**
+ * readpass_file(passwd, filename):
+ * Read a passphrase from ${filename} and return it as a malloced
+ * NUL-terminated string via ${passwd}.  Produce an error if the file is 2048
```
cperciva, line 26: "Produce an error" seems like it should be "Print an error and fail".

util/readpass_file.c
```diff
@@ -0,0 +1,75 @@
...
+			goto err2;
+		} else {
+			/* We have a 0-byte password. */
+			passbuf[0] = '\0';
+		}
+	}
+
+	/* Bail if there's a second line. */
+	if (fgetc(f) != EOF) {
+		warn0("more than 1 line in %s", filename);
```
cperciva, line 45: We could get here if there's a single overly long line, I think?

util/readpass_file.c
```diff
@@ -0,0 +1,75 @@
...
+ */
+int
+readpass_file(char ** passwd, const char * filename)
+{
+	FILE * f;
+	char passbuf[MAXPASSLEN];
+
+	/* Open the file. */
+	if ((f = fopen(filename, "r")) == NULL) {
+		warnp("fopen(\"%s\")", filename);
```
cperciva, line 28: I think I usually omit the `\"` in warning strings like this.

util/readpass_file.c
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * readpass_file(passwd, filename):
+ * Read a passphrase from ${filename} and return it as a malloced
+ * NUL-terminated string via ${passwd}.  Produce an error if the file is 2048
+ * characters or more, or if it contains any newline \n or \r\n characters
+ * other than at the end of the file.  Do not include the \n or \r\n
+ * characters in the passphrase.
+ */
+int
```
cperciva, line 20: Why not return a `char *`?

util/readpass_file.c
gperciva: yuck, yes.  That makes for a more confusing warning, but I suppose that the code clarity from switching to `fgetc` is still worth it.

util/readpass_file.c
gperciva: to match:
```
int readpass(char **, const char *, const char *, int);
```

cperciva: Hmm, ok.

cperciva:
Looks ok, please clean up the history.

gperciva:
Rebased, ready for merge.

PR #228, opened 2019-11-17
Readpass with no stdin fallback

gperciva:


gperciva:
This will be useful for `--passphrase dev:tty` in https://github.com/Tarsnap/scrypt/pull/216

util/readpass.c
```diff
@@ -82,11 +82,14 @@ readpass(char ** passwd, const char * prompt,
...
 		readfrom = stdin;
 		break;
-	default:
+	case 1:
 		/* Try to open /dev/tty; if that fails, read from stdin. */
 		if ((readfrom = fopen("/dev/tty", "r")) == NULL)
 			readfrom = stdin;
 		break;
+	default:
+		warn0("readpass does not support devtty=%i", devtty);
```
cperciva, line 99: It means the same thing, but normally we use `%d` rather than `%i` in printf strings.

gperciva: oh right, that's a hold-over from when I was teaching C to first-year students... I recommended `%i` for ints and `%f` for floats, because when you say "decimal" to a first-year engineer, they think you mean a number which contains a decimal point (e.g., 2.13).

cperciva:
The rest looks good.  Please fix the `%i` then I'll merge.

gperciva:
Updated

PR #231, opened 2019-12-09
build: avoid empty top-level -config.sh files

gperciva:


gperciva:
I'm sometimes confused when I try compiling on a raspberry pi because
there's a 0-byte `cpusupport-config.h`.  We don't support any special ARM
features, so that makes sense... but my first thought is always that
something went wrong in the build system.

We'll only ever end up with a blank `posix-flags.sh` if we don't include
`libcperciva/POSIX`, and I doubt that we'll ever have a project where we
use this build system but don't want to check POSIX compatibility.  But
I figured that we might as well use the same boilerplate for both config
files.

Makefile
```diff
@@ -52,19 +52,21 @@ posix-flags.sh:
 		printf "export \"CFLAGS_POSIX=";			\
 		command -p sh posix-cflags.sh "$$PATH";			\
 		printf "\"\n";						\
-	else								\
-		:;							\
 	fi > $@
+	if [ ! -s $@ ]; then						\
+		printf "#define POSIX_COMPATIBILITY_NOT_CHECKED 1\n";	\
```
cperciva: Why not just put this into the `if` statement?

gperciva: For `cpusupport-config.h`, the `CPUSUPPORT_NONE` will trigger if
1) we didn't include `libcperciva/cpusupport`.
2) if those scripts didn't detect any support (i.e. ARM).

I just used the same format for `posix-flags.sh`, even though `#2` isn't possible (since it always creates a non-zero file).

Looking in the future, this will apply to `cflags-filter.sh`, and we **do** expect that to be empty most of the time.  (Well, empty or a "`#define CFLAGS_OK 1` message or some message like that.)  With that in mind, I thought it would be less confusing if we used the same `Makefile` construct for generating all 3 "config" files, instead of having 2 different constructs.

cperciva: Hmm, I guess using the same structure has its merits.

PR #238, opened 2020-05-01
tests/parsenum: exfiltrate $? from shell-subshell

gperciva:


tests/parsenum/test_parsenum_improper_usage.sh
```diff
@@ -16,9 +17,9 @@ run_test() {
 	# from both () and $() was necessary to cover Linux, FreeBSD, and
 	# MacOS X.
```
cperciva, line 18: Should this say " and Solaris"?

gperciva: Technically, the previous discovery, regarding the non-pollution of output, made on linux, freebsd, and macosx, was sufficient.

I'll add a blurb about exfiltrating `$?` to that comment, which would therefore increase the number of OSes.

gperciva:
rebased

PR #239, opened 2020-05-01
POSIXFAIL for INET_ADDRSTRLEN

gperciva:


gperciva:
I'm not certain what level of documentation you want in the commit messages.  This is what I wrote originally, but then I decided that the first message didn't need the speculation about how this bug arose and I'd leave it as a github comment, but then I decided to leave it there (with this disclaimer) since it's easy to remove it later (before merging).

Why haven't other people noticed this?  Based on casual googling and skimming other project's bug trackers and PRs, it looks like most open-source software compiling on Solaris just defines `-D__EXTENSIONS__` and doesn't worry about strict standards compliance.

cperciva:
I think having details in the commit logs is good.  This information belongs in the repository -- otherwise 20 years from now when GitHub no longer exists and this is hosted on AWS Elastic Git nobody will be able to see why we made this change (or what to check before removing this POSIXFAIL).

A couple notes:
1. I believe this was added in Issue 7; Solaris probably had the #if for many years and they forgot to update it for the new release.  At least, "forgot to update for the new standard" seems more plausible to me than "put the #endif in the wrong place".
2. Other people probably haven't tripped over this since most people don't define _POSIX_C_SOURCE or _XOPEN_SOURCE.

gperciva:
Ok, feel free to merge (unless you want me to add the speculation about `__EXTENSIONS__` as well).

As for 2, python does contain `_XOPEN_SOURCE`, and the solaris port stripped it out, and then they decided they should have `__EXTENSIONS__`, and generally had a very muddled conversation in https://bugs.python.org/issue1759169.

PR #240, opened 2020-05-02
Pass $CC to POSIX/*.sh as an argument

gperciva:


gperciva:
This is more useful in conjunction with autoconf.  Compare:
```
CFLAGS_POSIX=`export CC="${CC}"; command -p sh "${POSIX_CFLAGS_SH}" "$PATH" "$CC"`
CFLAGS_POSIX=`env CC="${CC}" command -p sh "${POSIX_CFLAGS_SH}" "$PATH" "$CC"`
CFLAGS_POSIX=`command -p sh "${POSIX_CFLAGS_SH}" "$PATH" "$CC"`
```

In terms of "how many things can go wrong?"[1], I think the latter case is a clear winner.

[1] for example, on an operating system whose default shell utilities are not guaranteed to be POSIX-compliant and which requires a special $PATH to get POSIX behaviour.  I'm still not certain how that interacts with `command -p`.

cperciva:
I'd prefer to pass neither -- remember, we only pass $PATH on the command line because there was a shell (bash I think?) which didn't keep the environment properly.

I'm not absolutely opposed to doing this, but I don't buy the "we should handle CC the way we handle PATH" argument when PATH is getting special POSIX-violation-workaround handling.

PR #242, opened 2020-05-09
Test getopt longjmp

gperciva:


gperciva:
This consists of minimal reproductions of problems I've seen with getopt.  If we see any other problems (such as due to https://github.com/Tarsnap/scrypt/pull/234), I'll add it to this test.

PR #243, opened 2020-05-09
Alternate fix for getopt

gperciva:


gperciva:
We have a fix for the clang optimizer breaking getopt in `util/getopt.h`.  Unfortunately, it appears that gcc-7.3.0 on Solaris 11.4 requires a fix as well.  The same fix works (i.e. replacing the `'#ifdef __clang__` with `#if 1`), but then `c99` fails.

One option would be to add a `CFAIL_LONGJMP`, similar to our `POSIXFAIL` checks. 
 Then we could intelligently decide which set of preprocessor hacks to use.

Alternatively, we could rely on a `volatile int` to force the optimizers not to break getopt.  This PR allows the new https://github.com/Tarsnap/libcperciva/pull/242 test to succeed on all compilers on bsd, solaris, and linux-i386.  Once #242 is merged, I'll test this fix on more platforms, but so far it looks promising.

cperciva:
I'm not sure how this is supposed to work.  Having a `volatile` variable isn't an automatic "don't optimize anything"; compilers are allowed to do any of the optimizations they would normally do as long as the sequence of accesses to the variable doesn't change.

cperciva:
> We have a fix for the clang optimizer breaking getopt in `util/getopt.h`. Unfortunately, it appears that gcc-7.3.0 on Solaris 11.4 requires a fix as well. The same fix works (i.e. replacing the `'#ifdef __clang__` with `#if 1`), but then `c99` fails.

Have you tested gcc 7.3.0 on other platforms?  It seems unlikely that Solaris would have a uniquely broken version of gcc...

gperciva:
I've tested gcc 7.4 on buster-i386, and gcc 7.3.0 on Freebsd 10 i386 and amd64; no reproduction of the bug.  :(

Just as a quick refresh, given this code in `main.c`:
```
                GETOPT_OPT("-a"):
                        a = 1;
                        break;
```

the assembly code for gcc 7.3.0 on Solaris with `-O0` looks good:

```
.L22:
/ main.c:24:            GETOPT_OPT("-a"):
        nop
.L10:   
/ main.c:25:                    a = 1;
        movl    $1, -4(%rbp)    /, a
/ main.c:26:                    break;  
        jmp     .L2     /     
```

while with `-O1` we get:

```
/ main.c:27:            GETOPT_DEFAULT:
        cmpl    $0, getopt_initialized(%rip)    /, getopt_initialized
        jne     .L2     /,
        jmp     .L9     /
.L7:
/ main.c:25:                    a = 1;
        movl    $1, -1064(%rbp) /, %sfp
/ main.c:24:            GETOPT_OPT("-a"):
        cmpl    $0, getopt_initialized(%rip)    /, getopt_initialized
        jne     .L2     /,
/ main.c:24:            GETOPT_OPT("-a"):
        movq    -1040(%rbp), %rax       / getopt_ln_min, getopt_ln_min.8_9
        movl    $24, %esi       /, tmp114
        subq    %rax, %rsi      / getopt_ln_min.8_9, tmp113
        movl    $0, %edx        /,
        movl    $.LC0, %edi     /,
        call    getopt_register_opt     /
        movl    $1, %esi        /,
        leaq    -1024(%rbp), %rdi       /, tmp124
        call    siglongjmp      /
/ main.c:25:                    a = 1;
        movl    $1, -1064(%rbp) /, %sfp
        jmp     .L2     /
```

(I did discover that this "alternative fix for getopt" doesn't work on all platforms, though.)

PR #244, opened 2020-05-09
Test cpusupport

gperciva:


gperciva:
This is a test of whether the compiler can handle missing symbols (or rather, whether our preprocessor tricks work with that compiler).  See https://github.com/Tarsnap/libcperciva/blob/master/cpusupport/cpusupport.h#L24 for the description of the tricks.

This test works on gcc and clang, but on Solaris with `cc` or `c99`, fails with:

```
cc -o test_cpusupport main.o cpusupport_x86_sse2.o    -lrt -lxnet
Undefined                       first referenced
 symbol                             in file
cpusupport_x86_aesni_init_1         main.o
cpusupport_x86_ssse3_init_1         main.o
cpusupport_x86_aesni_present_1      main.o
cpusupport_x86_aesni_detect_1       main.o
cpusupport_x86_ssse3_detect_1       main.o
cpusupport_x86_ssse3_present_1      main.o
ld: fatal: symbol referencing errors
*** Error code 2
```

gperciva:
... wait, this should be unnecessary, as the existing cpusupport-related tests should show this error.

PR #245, opened 2020-05-13
Fix readpass_file newlines

gperciva:


gperciva:
Whoops.  I forgot that I needed to manually inspect the test output; when we switched to `fgets()` I just checked that the exit code was correct.  Test suite now upgraded to automatically check for newlines.

gperciva:
Having all those almost-duplicate lines in the test suite is a bit clumsy, but it seemed like over-engineering just to make a function just to run the command and then check the output.  After this PR, we shouldn't need to tweak the test script (knock on wood).

PR #247, opened 2020-05-23
Print descriptions of failed tests

gperciva:


gperciva:
Most of these are obvious (i.e. if "humansize" fails, there's only one binary to check), but some aren't so clear (i.e. if something in setuidgid fails, there's 10 different tests to check).  I've occasionally had to resort to manually counting `setup_check_variables` lines to find the 8th one in order to figure out which command is failing.

It would be much nicer if the test suite gave me a unique string so that I could instantly find the piece of code which fails.

tests/shared_test_functions.sh
```diff
@@ -181,15 +181,22 @@ ensure_valgrind_suppression() {
...
 }
 
-## setup_check_variables ():
+## setup_check_variables (description):
 # Set up the "check" variables ${c_exitfile} and ${c_valgrind_cmd}, the
 # latter depending on the previously-defined ${c_valgrind_min}.
 # Advances the number of checks ${s_count} so that the next call to this
-# function will set up new filenames.
+# function will set up new filenames.  Write ${description} into a
+# file.
```
cperciva, line 189: I feel like there might be an intended "and print it if a test fails" or something like that?

gperciva: Technically, the "print it" doesn't happen in the `setup_check_variables` function.  But yes, I should add such a comment to the description of `notify_success_or_fail`.

gperciva:
Rebased with that extra comment.

PR #250, opened 2020-05-27
add POSIXFAIL_ABSTRACT_DECLARATOR

gperciva:


gperciva:
I'm not 100% certain about the `-DPOSIXFAIL_ABSTRACT_DECLARATOR` name, but I'm not coming up with anything snappier.

POSIX/posix-abstract-declarator.c
```diff
@@ -0,0 +1,17 @@
+/* To trigger the problem, we must not have an argument name. */
+static int func(int [static restrict 1]);
```
cperciva, line 2: How about making this `static int func(int ARGNAME[static restrict 1]);` and testing not just whether it fails to compile with ARGNAME defined to "" but also checking that it compiled with ARGNAME *not* defined to ""?

Otherwise this test could fail for the wrong reason (e.g. old compilers which don't understand "restrict").

gperciva: hmm, ok.  I put this test after the `restrict` test in `posix-cflags.sh`, but your method is safer.

gperciva:
Rebased as suggested.  Also added a verbose comment, which might be too verbose but I figured better more than less.

gperciva:
See https://github.com/Tarsnap/libcperciva/pull/252 for an example of this in practice.

POSIX/posix-abstract-declarator.c
```diff
@@ -0,0 +1,25 @@
+/* To trigger the problem, we must not have an argument name.
```
cperciva, line 1: `/**`

Also, do you want to require `-DARGNAME=""` vs. `-DARGNAME="var"`?  I was thinking that we could `#define ARGNAME` based on whether POSIXFAIL_ABSTRACT_DECLARATOR was defined.

gperciva: Hmm... none of the other `POSIX/*.c` files make use of `POSIXFAIL_*`, but that _would_ simplify `posix-cflags.sh`.  Sure.

gperciva:
I rebased the change (and tested on openbsd), so it's potentially ready for merge now.

PR #251, opened 2020-06-02
POSIX: fail if we can't work around a problem

gperciva:


gperciva:
Most of the POSIX tests have a workaround (i.e. if a header doesn't include a symbol, we can #define it ourselves, or else omit compiling that part of the code).

However, in the case of 'restrict', if a compiler can't handle it (even with `-std=c99`), then they won't be able to compile `alg/sha256.c`, so there's no point continuing with the compilation.

cperciva:
What if we're not trying to compile `sha256.c`?  There might be code using libcperciva which doesn't use SHA256 some day.

Also, I'd rather let the compilation continue and have it fail with whatever error the compiler throws -- that's more likely to be googlable than anything we come up with.

gperciva:
Well, I was thinking that if somebody was trying to use a compiler which didn't support `restrict`, they'd get in touch with us.  But ok.

PR #252, opened 2020-06-03
Use POSIXFAIL_ABSTRACT_DECLARATOR

gperciva:


gperciva:
This allows libcperciva to compile on OpenBSD with gcc 4.2.1.  Depends on https://github.com/Tarsnap/libcperciva/pull/250.

alg/sha256_shani.h
```diff
@@ -10,7 +10,12 @@
  * instructions, and should only be used if CPUSUPPORT_X86_SHANI and _SSSE3
  * are defined and cpusupport_x86_shani() and _ssse3() return nonzero.
  */
+#ifdef POSIXFAIL_ABSTRACT_DECLARATOR
```
gperciva, line 13: I'm not certain whether you'd prefer `#ifdef` or `#ifndef`.

... of course you'd prefer none at all, because this looks disgusting (especially in a header file!).  But if we want to support the default compiler on `configure` on OpenBSD, we need something like this.  :(

cperciva: `#ifdef` is fine here.

gperciva:
(this little travis-ci "failure" was a random mpool timing test, not a real problem)

PR #253, opened 2020-06-08
Clarify getopt

gperciva:


gperciva:
These should be relatively simple clarifications.  I don't anticipate any problems, other than perhaps some quibbles over exact terminology used in the comments.

Other changes to getopt (such as running it twice in the same function) will happen in later PR(s), once I think I have a clean way of doing them.

Issue #254, opened 2020-06-10
PARSENUM, pragmas, and perfection


gperciva:
(where "perfection" means "a synonym of 'diagnostics' or 'warnings' that begins with a p")

I'm investigating this code:
```
#include "parsenum.h"
int main()
{
        unsigned int val;
        // ok
        if(PARSENUM(&val, "4", 2, SIZE_MAX/2))
                return 1;
        // gcc fail
        if(PARSENUM(&val, "4", 2, SIZE_MAX/2+1))
                return 1;

        return 0;
}
```

It's totally fine with
```
clang -Weverything -Wno-reserved-id-macro -I ../../util problem.c
```
However, the "fail" line gives a warning with
```
gcc9 -Wall -Wextra -I../../util problem.c
../../util/parsenum.h:124:14: warning: comparison of unsigned expression < 0 is always false [-Wtype-limits]
  124 |     ((((max) < 0) && (errno == 0)) ? \
      |  
```
(this is on FreeBSD 12.1, BTW, but I don't think that influences anything here)

gperciva:
Unfortunately, `PARSENUM_IF` isn't a solution for this code:
```
if ((argc != 2) || (PARSENUM(&sets, argv[1]))
```

(slightly modified from `tests/mpool/main.c`)

Unless we can disable compiler warnings in the middle of an expression (as clang allows), we'd be stuck doing
```
PARSENUM_PROLOGUE
if ((argc != 2) || (PARSENUM(&sets, argv[1]))
        ...
PARSENUM_EPILOGUE
```

Quite vexing!

PR #255, opened 2020-06-11
Fix parsenum bounds

gperciva:


gperciva:
I should have spent more time carefully thinking about the code, instead of trying to use the "suppress warnings" (#254) hammer.

util/parsenum.h
```diff
@@ -121,8 +121,10 @@ _Pragma("clang diagnostic pop")
 			    ((min) <= 0) ? 0 : (min),			\
 			    (uintmax_t)(max), (*(x)), (base),		\
 			    (trailing))),				\
-				((((max) < 0) && (errno == 0)) ?	\
-				    (errno = ERANGE) : 0)),		\
+				(((max) <= INTMAX_MAX) ?		\
```
cperciva, line 124: Not sure what you mean by "if the compiler has to treat (max) as an unsigned value" here.  Does this help if max is an unsigned variable?

gperciva: Right, it took me a while to decypher the commit message.  The idea is basically sanity-checking the upcoming cast to `intmax_t`.  I mean, if max is `INTMAX_MAX + 1`, then we can't do `((intmax_t)(max) < 0) ?`

... at the time, I thought that I had a reason to split up the change into two commits, but at the moment I'm not recreating my thought process.

cperciva:
I'm still not understanding the "if (max) is higher than INT_MAX" bit.

gperciva:
... right, it looks like the commit message is still bogus.  On my 64-bit system, it looks like anything over `LONG_MAX` triggers the problem.  But that's almost certainly a system-specific thing.

Are we discussing the commit message, or the code change as well?  Here's a simplified version of the problem:
```
$ cat simple.c 
#include <limits.h>
#include <stdint.h>
#include <stdio.h>

#define IS_NEGATIVE(x)					\
	((x) < 0) ? 1 : 0

int main() {
	if (IS_NEGATIVE(SIZE_MAX))
		printf("negative\n");
	else
		printf("positive\n");
}
$ gcc -Wall -Wextra simple.c
simple.c: In function 'main':
simple.c:6:7: warning: comparison of unsigned expression < 0 is always false [-Wtype-limits]
    6 |  ((x) < 0) ? 1 : 0
      |       ^
simple.c:9:6: note: in expansion of macro 'IS_NEGATIVE'
    9 |  if (IS_NEGATIVE(SIZE_MAX))
      |      ^~~~~~~~~~~
$
```

Replacing `SIZE_MAX` with `SIZE_MAX/2` makes it work.  Replacing it with `SIZE_MAX/2 + 1` causes the same problem.

My proposed solution is:

```
$ cat prob.c 
#include <limits.h>
#include <stdint.h>
#include <stdio.h>

#if 0
#define IS_NEGATIVE(x)					\
	((x) < 0) ? 1 : 0
#else
#define IS_NEGATIVE(x)					\
	(((x) <= INTMAX_MAX) ?				\
		(((intmax_t)(x) < 0) ? 1 : 0) :		\
		0)
#endif

int main() {
	if (IS_NEGATIVE(SIZE_MAX))
		printf("negative\n");
	else
		printf("positive\n");

	if (IS_NEGATIVE(UINTMAX_MAX))
		printf("negative\n");
	else
		printf("positive\n");
}
$ gcc -Wall -Wextra prob.c && ./a.out 
positive
positive
td@beastie: ~/tmp
$ 
```

(I should clarify that at this point of `PARSENUM`, we're only dealing with unsigned integers, not floats or signed.)

cperciva:
It looks to me like the warning is being triggered when `max` has an unsigned type.  What you're seeing with constants is probably because literal constants are typed as signed unless the only type they fit into is an unsigned type.

gperciva:
Yes.  But looking at kivaloo's `lib/http/http.c`, would you rather see:
```
if (PARSENUM_EX(&clen, (const char *)buf, 0, SIZE_MAX, 16, 1)) {
```
or
```
if (PARSENUM_EX(&clen, (const char *)buf, 0, (long)SIZE_MAX, 16, 1)) {
```

?

cperciva:
Obviously not `(long)SIZE_MAX` since signed integer overflow is undefined behaviour.

I'm not saying that the code is wrong, just that the commit message is wrong.  The issue isn't how *big* `max` is; the issue is what *type* it has.

PR #257, opened 2020-06-18
Tests sanity check

gperciva:


cperciva:
You know the tests code far better than I do, but... I have no clue what purpose this serves.  Can I have some "why" along with my helping of "what"?

gperciva:
Rebased with new commit messages.

This will catch mistakes such as:
    
    setup_check_variables "check_thing"
    check_thing
    # <--- here
 
    setup_check_variables "check_another"
    check_another
    echo "$?" > "${c_exitfile}"
    
At the "<-- here", we should have written the exit code to ${c_exitfile}.  Before this commit, the test suite would be unaware of the lack of the exit code, so an error in check_thing would not be noticed.

gperciva:
In some rare circumstances [1], we know that we won't have an exit code before we reach the next setup_check_variables.  In that case, we can suppress the check by setting check_prev = 0.
    
[1] for example, in one of the spiped tests, we have:

    setup_check_variables "multi open 2 seconds close"
    (
        ( echo "" ; sleep 2 ) | check_thing
            echo $? > ${c_exitfile}
    ) &
    
    setup_check_variables "multi open close"

where the second 'setup_check_variables' will need to suppress this check by adding a 0 to that line.

PR #258, opened 2020-06-21
POSIX: use CFLAGS in checks

gperciva:


gperciva:
The situation for the tarsnap client is icky, but I'm reasonably confident that this is a good addition in theory.

Quick recap: on Solaris, `clang` is packaged such that it automatically defines `__EXTENSIONS__`.

For libcperciva and spiped, there's no danger (at present) from including this definition.  That's not to say that it has no effect -- it makes one of our workarounds unnecessary.  Namely, defining `__EXTENSIONS__` exposes the definitions for `INET_ADDRSTRLEN` and `INET6_ADDRSTRLEN`, so our `POSIXFAIL_INET_ADDRSTRLEN` won't get added.

For tarsnap, `__EXTENSIONS__` produces a conflict with one of our variables, called `CS`.  So I'd like to make sure that we can get the right `POSIXFAIL` when using `clang`, which won't happen unless we pass the `-U` to `posix-cflags.sh`.

(Then the next layer of the problem is autoconf helpfully deciding that we want to add `__EXTENSIONS__`, possibly as a result our `AC_GNU_SOURCE`, which is now considered to be an obsolete version of `AC_USE_SYSTEM_EXTENSIONS`.   ... but that's something for a separate PR in a different project.)

gperciva:
On second thought, it might be cleaner to fix the solaris issue in a different way.

I think this is still a good change in general, though.

PR #259, opened 2020-06-22
getopt: fix for gcc < 9 on solaris

gperciva:


gperciva:
OIptimizers have a hard time with nonlocal gotos.  On Solaris <setjmp.h> contains:

    #pragma unknown_control_flow(sigsetjmp)
    extern void siglongjmp(sigjmp_buf, int) __NORETURN;

The pragma is for Solaris' cc.  __NORETURN expands to __attribute__((__noreturn__)), unless we've requested POSIX compability (via _XOPEN_SOURCE in this case), in which case it expands to nothing.

So as far as gcc is concerned, there's a function called siglongjmp() that has no special control flow issues.  As it happens, gcc7 and gcc8 both miscompile getopt with -O1 and have no problem with -O0 or -O2.  Later versions of gcc are fine even with -O1 and no control flow hints.

Adding __builtin_unreachable() after our siglongjmp(getopt_initloop, 1) gives gcc enough info to produce non-broken code.

gperciva:
I've reproduced this on FreeBSD:
```
$ gcc7 -O1 getopt-gcc7-fail.c && echo "compile ok" && ./a.out ; echo $?
compile ok
1
$ gcc8 -O1 getopt-gcc7-fail.c && echo "compile ok" && ./a.out ; echo $?
compile ok
1
$ gcc9 -O1 getopt-gcc7-fail.c && echo "compile ok" && ./a.out ; echo $?
compile ok
0
```

Code here, and I'll attached it as well (in case you want to investigate and that's easier).
```
/* <setjmp.h> */
typedef long sigjmp_buf[128];
int sigsetjmp(sigjmp_buf, int);
void siglongjmp(sigjmp_buf, int);
#if 1
/* On Solaris, _XOPEN_SOURCE omits the attribute. */
extern void siglongjmp(sigjmp_buf, int);
#else
/* Normal header declaration. */
extern void siglongjmp(sigjmp_buf, int) __attribute__((__noreturn__));
#endif

/* <stdlib.h> */
void exit(int);

/* Our code */
static int initialized = 0;
int main() {
  int badvar = 0;
  while (!initialized) {
    volatile int c = 1, d = 1, e = 0;
    sigjmp_buf h;
    sigsetjmp(h, 0);
    switch (initialized ? c : d++) {
    case 1:
      if (initialized)
        goto badplace;
      siglongjmp(h, 1);
    badplace:
      badvar = 1;
      break;
    case 2:
      initialized = 1;
      break;
    default:
      if (initialized)
        goto done;
      if (!e)
        siglongjmp(h, 1);
    done:
	;
    }
  }
  if (badvar)
    exit(1);
}
```

gperciva:
WTF, github doesn't allow attachments ending in `.c` ?!?!  ok, I appended a `.txt`.


[getopt-gcc7-fail.c.txt](https://github.com/Tarsnap/libcperciva/files/4815312/getopt-gcc7-fail.c.txt)

util/getopt.h
```diff
@@ -40,6 +40,17 @@
 	getopt_initloop_ ## x:
 #define DO_LONGJMP							\
 	goto *getopt_initloop
+
+/* Work around GCC < 9 bug. */
```
cperciva, line 44: Is this a GCC bug, or a Solaris bug?  Isn't the problem that Solaris isn't defining sigsetjmp properly?

gperciva:
I honestly don't know where the bug resides.

I mean, if you ask for POSIX compatibility and don't specify `__EXTENSIONS__`, Solaris' built-in header defines
```
extern void siglongjmp(sigjmp_buf, int);
```

Is that a bug?  There's no standard-compliant way of saying "this function doesn't return" in C99.  (C11 would be so much easier: `_Noreturn`, `_Generic`, static assertions, etc.)

On the other hand, as far as gcc is concerned, if the header doesn't indicate anything special about `siglongjmp()`, is it allowed to optimize it however it sees fit?

I'm a bit confused about the division between "a C compiler", "the C99 standard", and "the header files".  Naively, I'd expect that if a compiler requires certain magic in a header file to obey the standard, then the headers would be tied to that compiler, rather than being tied to a particular operating system.

PR #260, opened 2020-06-23
Test getopt twice

gperciva:


gperciva:
Not certain how we want to approach this:
- we want to test multiple `GETOPT` loops in the same function.
- that causes a false positive from clang's `-Wconditional-uninitialized`.
- clang developers are pretty clear that this flag is expected to give false positives, e.g.,
  https://bugs.llvm.org/show_bug.cgi?id=18709#c1 https://bugs.llvm.org/show_bug.cgi?id=38856#c1

I've gone with disabling that warning in the test code, limiting the scope to the minimum needed.

Another solution is not to have `WARNING_FALSE_POSITIVE_SUPPRESS` at all, but modify `.travis.yml` to include `-Wno-conditional-uninitialized`.  That's not my favourite solution, since it's possible that flag might give a non-false warning about some code in the future and it seems a shame to disable it globally just for this, but I won't complain if you prefer that solution.

A much uglier idea is to suppress the warnings in `getopt.h` itself, not just in the test code.  That would have the benefit of removing the warnings for anybody who wants to use `GETOPT` twice in a function, but it makes `getopt.h` substantially uglier since we need to disable/re-enable that flag in multiple places.

PR #262, opened 2020-07-02
Use goto for GETOPT on Solaris with gcc

gperciva:


gperciva:
This is an alternate fix for #259.  (I forgot about that PR, and solved the problem in a different way.)

I suspect that you might prefer this solution, since it leverages the existing clang workaround, and doesn't involve `__builtin_unreachable()`.

util/getopt.h
```diff
@@ -33,6 +33,11 @@
 #ifdef __clang__
 #warning Working around bug in LLVM optimizer
 #warning For more details see https://bugs.llvm.org/show_bug.cgi?id=27190
+#define GETOPT_USE_GOTO
+#endif
+
+#ifdef GETOPT_USE_GOTO
+/* Workaround with goto. */
```
cperciva, line 45: I'd say `computed goto` here -- longjmp is arguably a type of goto, but the key issue is that *computed* gotos aren't standard-compliant.

util/getopt.h
```diff
@@ -36,6 +36,11 @@
 #define GETOPT_USE_GOTO
 #endif
 
+/* Work around GCC + headers bug on Solaris. */
+#if defined(__GNUC__) && (defined(sun) || defined(__sun))
```
cperciva, line 40: I know you never observed this issue on !GCC, but are you certain it can't occur with other compilers?  Since the issue is in Solaris's libc header, I'd be inclined to just make this apply any compiles on Solaris.

(This could theoretically result in code not compiling if someone uses a non-computed-goto-supporting compiler on Solaris, but that's a better outcome than potentially miscompiling due to broken header files.)

We should probably include a `#warning Working around broken <setjmp.h> header on Solaris` here too.

gperciva: Is it actually a bug in the Solaris libc header, though?  The Solaris header defines

```
extern void siglongjmp(sigjmp_buf, int);
```

There's no standard-compliant way of saying "this function doesn't return" in C99.
But as far as gcc is concerned, if the header doesn't indicate anything special about `siglongjmp()`, is it allowed to optimize it however it sees fit?

I'm a bit confused about the division between "a C compiler", "the C99 standard", and "the system header files". Naively, I'd expect that if a compiler requires certain magic in a header file to obey the standard, then the headers would be tied to that compiler, rather than being tied to a particular operating system.

Suppose I create a GrahamCC compiler, which requires `#pragma GrahamCC func_noreturn` in order to avoid mis-optimizing `siglongjmp()`.  Is it a bug that the FreeBSD <sigjmp.h> doesn't include that line?

gperciva: As for other compilers...

- we know that `clang` is broken, so that's already using computed goto.
- the Solaris header does include
  ```
  #pragma unknown_control_flow(sigsetjmp)
  ```
  which is understood by Solaris' `cc` and `c99`.  (more details: https://github.com/Tarsnap/libcperciva/pull/259#issuecomment-647734786)

Those 4 compilers (cc, c99, clang, gcc) are the only ones I can think of to test.

cperciva: Fair enough, this sort of "toolchain" issue is really in a grey zone between OS and compiler.  But given that Solaris *attempts* to mark siglongjmp as `attribute((noreturn))` but their implementation is broken, I'm inclined to say that it's their fault -- whoever ported GCC to Solaris could very plausibly have looked at the header and said "ok, as long as we handle `noreturn` we're good".

It's rather like the Hyperthreading issue from 2005 -- was it a bug in the chips or a bug in the software?  My view was that software developers had a reasonable expectation that the CPU wouldn't leak information between hyperthreads.

gperciva:
Rebased PR.  I adjusted the comments and `#warning` to be more general to Solaris, but I left the `#if` to only apply to gcc.

PR #264, opened 2020-07-11
Tests md5

gperciva:


gperciva:
I like to keep auto-generated code in a separate commit from real code; that narrows down the scope of investigating bugs.  Since `15-md5.sh` is so small, I'd be tempted to rebase that into the first commit as well.

Doesn't compile with nitpicky clang 10 unless I cherry-pick the most recent update to `util/monoclock.h`, so it might be useful to rebase the entire thing onto master.

gperciva:
also, the test should be number 16.  This code was branched off last dec (why?), so it didn't have `tests/15-readpass-file.sh`.

gperciva:
I also have sha1 tests ready, but I'll wait until you've reviewed this one before uploading it, as it'll need any fixes that this one needs.

cperciva:
Yeah, feel free to rebase this onto the latest bits... I ran `git checkout master` and forgot that git doesn't give me the *latest* master when I do that.

cperciva:
And go ahead with rebasing commits you've marked as such.

gperciva:
Done.

`git rebase master` gives you the latest `master` it knows about.  What, you expected it to give you something that it doesn't know about?  :)

cperciva:
I'm used to "checkout" meaning "download from the server".  You know, like it does on every other VCS...

gperciva:
wait, the test is broken.  Fix incomping

gperciva:
(it was missing the "describe the test" safety check)

gperciva:
fixed, ready for merge now.

gperciva:
I forgot to update `.gitignore`, but I propose to do that in tests-sha1.

cperciva:
Hmm, it looks like `perftest.sh` is broken -- the CFLAGS are getting reset by the build infrastructure to the default `-O2`.

gperciva:
Right, `perftests.sh` is setting its flags like this:

```
export CC="$COMPILER $OFLAG"
```

, so the normal $CFLAGS will come later in the command-line.  When you added it in dd9cf8ccd6922d8eaf00e8e6674f58043f2a3ec9, we didn't have `CFLAGS` in `Makefile`; that was only added in 168fc3c9ba0646f64b55b6831624310b39ae2d82.  Another step in the problem is the "loopback" building, wherein we dip down to the top-level in order to recreate `posix-flags.sh` if necessary.

I'll put a fix in a separate PR first, then revisit this one.

gperciva:
Rebased.  I snuck in a small generalization to `tests/sha256/perftests.sh` so that we can reuse it more easily for md5 and sha1.

PR #265, opened 2020-07-12
Fix tests/sha256/perftest.sh

gperciva:


tests/sha256/perftest.sh
```diff
@@ -1,10 +1,15 @@
 #!/bin/sh
 
+# Extra space, for nicer formatting.
+CFLAGS_ORIG=" ${CFLAGS:-}"
```
cperciva, line 4: This gives me
```
clang  -O2,6.630159
clang  -O3 -march=native,5.363226
```

If `CFLAGS` is empty we don't need the extra space here.

gperciva:
Oops, fixed.

PR #268, opened 2020-07-14
Fix scripts

gperciva:


gperciva:
I saw a cat next to a pipe and a perl, so I got rid of it.  `=(^_^)=`

PR #269, opened 2020-07-15
Style

gperciva:


util/getopt.h
```diff
@@ -42,6 +42,7 @@
 #define GETOPT_USE_COMPUTED_GOTO
 #endif
 
+/* Select the method of performing non-local jumps. */
```
cperciva, line 45: We're not performing non-local jumps, though.  I mean, setjmp/longjmp *can* perform non-local jumps, but the jumps we're doing are local.  And computed gotos can only be used for local jumps.

gperciva: huh, the wikipedia page for `<setjmp.h>` described it as being for non-local jumps.  Ok.

gperciva:
Rebased with "performing local jumps".

util/getopt.h
cperciva: Well, yes.  That is what it's for.  The way we're using it to do a jump within the same function is very unusual.

PR #270, opened 2020-07-16
Tests valgrind no exitcode

gperciva:


gperciva:
This is the first step towards being able to check daemonize'd programs.  Subsequent steps are #271 and #272.

PR #271, opened 2020-07-16
Tests valgrind multiple pids

gperciva:


gperciva:
This now requires valgrind 3.13 or higher, which means that we can't do `USE_VALGRIND=1` on FreeBSD.  But the valgrind port has been broken for years (apparently memory checking is disabled?), so that's not a big problem for us.

cperciva:
I haven't reviewed this in detail, but I assume it's working for you.  And it's "only" test framework anyway...

PR #272, opened 2020-07-16
Tests daemonize

gperciva:


gperciva:
Note to self: renumber this test to 19, because warnp will be 18 before this gets in.

tests/daemonize/main.c
```diff
@@ -0,0 +1,46 @@
...
+#include "warnp.h"
+
+int
+main(int argc, char * argv[])
+{
+	const char * pidfile;
+
+	/*
+	 * This will allocate some memory for the program name, which will be
+	 * free'd by an atexit() function.  The non-daemonized fork of this
```
cperciva, line 16: "free'd"?

tests/daemonize/main.c
```diff
@@ -0,0 +1,46 @@
...
+	 * program will quit without calling that function, which might cause
```
cperciva, line 17: s/quit/exit/

tests/daemonize/main.c
gperciva: No?  But it highlights the function name!  I suppose that I should have written "`free()`d", though...

... ok, changed to "freed".

tests/daemonize/main.c
gperciva: Hmm.  I was trying to avoid using the same name as a library function, since the non-daemonized fork calls `_exit` rather than `exit`.

What about this instead:

> The non-daemonized fork of this program will call _exit() to quit without calling that function.  This might cause a memory checker to issue an invalid warning which should be suppressed.

cperciva:
LGTM.

gperciva:
Rebased, ready for merge.

PR #274, opened 2020-07-19
Warnp: add ability to send to syslog

gperciva:


gperciva:
API assumptions behind syslog for `warnp`:

- we don't want to edit any existing `warn0()` or `warnp()` lines.  (That means that we need to take care of the syslog priority ourselves, instead of deferring it to the caller.)

- the default priority level is `LOG_WARNING` rather than `LOG_ERR`.

- it's ok to limit the length of syslog messages.  (We could theoretically dump more than INT_MAX characters to the console via `warnp()` [1], but it would be impolite to send that to syslog.)

[1] in kivaloo, `lib/http/http.c.`

gperciva:
First draft ready for review.

I'm not happy with the games I had to play to check syslog automatically.  I'm actually wondering if it's worth keeping that.  This PR is literally the only time we've touched `warnp.c` since it was added to libcperciva in 2011-09-07.  (`warnp.h` got 2fixes since then.)  We could simplify the test script by omitting syslog stuff, and just glance at that manually when (or if!) we ever touch `warnp.c` again.

(OTOH, if the current code is actually working, then we might as well keep it, especially since it makes it easy to check that code on various OSes.)

util/warnp.c
```diff
@@ -8,6 +8,7 @@
 
 static int initialized = 0;
 static char * name = NULL;
+static int syslog_use = 0;
```
cperciva, line 12: This code, by Yoda, was written?

Not sure why this isn't `use_syslog`.

util/warnp.c
```diff
@@ -51,12 +52,18 @@ warn(const char * fmt, ...)
 	va_list ap;
 
 	va_start(ap, fmt);
-	fprintf(stderr, "%s", (name != NULL) ? name : "(unknown)");
-	if (fmt != NULL) {
-		fprintf(stderr, ": ");
-		vfprintf(stderr, fmt, ap);
+	if (syslog_use == 0) {
+		/* Print to console. */
```
cperciva, line 59: s/console/stderr/.  The *console* is something quite different.

tests/18-warnp.sh
```diff
@@ -9,24 +9,34 @@ test_output="${s_basename}-stderr.txt"
...
-	# We should have 7 lines in total.
-	test "$(wc -l < "${filename}")" -eq "7"
+	# Console or syslog?
+	if [ "${is_syslog}" -eq "0" ]; then
+		# Check console.
 
-	# We should have 2 lines without a colon.
-	test "$(grep -c -v "test_warnp:" "${filename}")" -eq "2"
+		# We should have 7 lines in total.
+		test "$(wc -l < "${filename}")" -eq "7"
```
cperciva: Why is this 2?  Is there a change to the test code which should be part of this commit?

cperciva: Oh, never mind, I got confused by how diff is matching up the lines.

util/warnp.c
```diff
@@ -62,7 +65,18 @@ warn(const char * fmt, ...)
...
 	} else {
 		/* Print to syslog. */
-		/* Not implemented. */
+		if (fmt != NULL) {
+			/* No need to print "${name}: "; syslog does it. */
+			vsnprintf(msgbuf, WARNP_SYSLOG_MAX_LINE + 1, fmt, ap);
+			syslog(syslog_priority, "%s: %s\n", msgbuf,
+			    strerror(errno));
+		} else {
+			if (errno != 0)
```
cperciva, line 74: Why?  There's no such check in the non-syslog version.

util/warnp.h
```diff
@@ -8,6 +8,8 @@
 #define warn libcperciva_warn
 #define warnx libcperciva_warnx
 
+#define WARNP_SYSLOG_MAX_LINE 255
```
cperciva, line 11: Let's make this 4095.  I can imagine logging warnings which are more than 255 characters.

cperciva:
Yeah, I'm inclined to not bother testing that syslog goes somewhere reasonable -- especially since there's no POSIX-compliant way to test that anyway.

util/warnp.c
gperciva: IMO, this looks messy:
```
static int use_syslog = 0;
static int syslog_priority = LOG_WARNING;e
```
Besides, `syslog_foo` is easier to recognize as "something to do with syslog", rather than `bar_syslog`.

But ok, I can go back to `use_syslog` if you're not a fan of short green aliens.

util/warnp.c
gperciva: done

util/warnp.c
gperciva: renamed to `use_syslog`

tests/18-warnp.sh
gperciva: Yeah, I was annoyed by the way that diff was generated.

util/warnp.c
gperciva: Oops.  Removed.

util/warnp.h
gperciva: done

gperciva:
Since I ended up with almost as many lines in `REBASE` commits as real commits, I rebased the whole thing.

util/warnp.h
```diff
@@ -8,6 +8,8 @@
 #define warn libcperciva_warn
 #define warnx libcperciva_warnx
 
+#define WARNP_SYSLOG_MAX_LINE 4095
```
cperciva, line 11: `/* Maximum length of messages sent to syslog; longer warnings will be truncated. */`

gperciva: done

cperciva:
Looks good, please clean up history.

gperciva:
Done, ready for merge.

PR #276, opened 2020-07-22
tests: use numeric comparison for valgrind check

gperciva:


gperciva:
Whoops.  Fix `=` instead of `==`.

The valgrind check was a last-minute addition, and didn't receive as much scrutiny as the rest of it.  I was doing my detailed testing with bash, since it has nicer error messages.  This bug would only actually matter when valgrind hits version 4.0.  The travis-CI log showed the warning, but it wasn't an error so I had no reason to look at the detailed log.  Also, I was in the middle of fixing additional nitpicks with shellcheck, so I didn't run that against the final PR.  (If I had, it would have warned about this.)

PR #277, opened 2020-07-22
daemonize: flush output buffers before fork()

gperciva:


cperciva:
I don't understand this.  The parent process is exiting with _exit, which doesn't flush stdio streams; where are you getting two copies of output from?

gperciva:
I'm closing this.

It turns out to only be applicable when 1) printing to stdout, 2) using valgrind, 3) on linux, and 4) redirecting the output.  Three of those are used in our test suite, so I've made `test_daemonize` print to `stderr`.

Minimal example:
```
#include <stdio.h>
#include <unistd.h>

int main()
{
    int pid;
    printf("before fork %d\n", getpid() );

#if 0
    fflush(stdout);
#endif

    pid = fork();
    if( pid != 0 )
	_exit(0);

    printf("after fork %d\n", getpid() );
    return 0;
}
```

Compile, then:
```
td@focal:~/tmp$ valgrind --log-file=log.txt ./a.out
before fork 36013
after fork 36014
td@focal:~/tmp$ valgrind --log-file=log.txt ./a.out | cat
before fork 36015
before fork 36015
after fork 36017
td@focal:~/tmp$ 
```

Can't reproduce on freebsd, although I don't know if that's due to the OS itself or just the old valgrind version.

If the printing was happening inside valgrind itself, I'd suspect that it wasn't calling `_exit()` in the child process.  As it is... does valgrind redirect the streams of the binary it's checking?  I don't know enough about its internals to guess at what's happening, and I don't think it's worth digging deeper (compared to some other things I should be doing).

PR #278, opened 2020-07-26
Misc

gperciva:


gperciva:
- `$(...)` instead of backticks: I split this up so that I could easily import the changes to other projects.  (e.g., scrypt and tarsnap don't contain metabuild, so I didn't want that in the same commit as POSIX and cpusupport)

- I added "Tarsnap Backup Inc." to the COPYRIGHT file, since we do that in other projects.  Your personal copyright line ends in 2016; do you want me to increase that to 2020 to match the company line?

cperciva:
Yes, please adjust my personal copyright up to 2020.  Some of the bits which have ended up here have been things I've done wearing my "FreeBSD developer" hat.

util/parsenum.h
```diff
@@ -143,6 +143,9 @@ parsenum_float(const char * s, double min, double max, int trailing)
 	char * eptr;
 	double val;
 
+	/* Sanity check. */
```
cperciva, line 146: This commit is fine, but the message talks about `optarg` which seems a bit of a red herring?

I mean, ok, `s` is `optarg` if you're using this to parse command-line arguments, but you shouldn't need to understand about getopt in order to understand a commit to parsenum.

COPYRIGHT
```diff
@@ -1,6 +1,7 @@
 The code and documentation in this directory ("libcperciva") is distributed
 under the following terms:
 
+Copyright 2005-2020 Tarsnap Backup Inc.  All rights reserved.
```
cperciva, line 4: Make that 2011-2020 since Tarsnap Backup Inc. didn't exist before 2011.

gperciva: ok.  I picked 2005 up from the scrypt COPYRIGHT file, so I'll fix this in other projects as well.

util/parsenum.h
gperciva: Ok, I removed the detailed commit message.

... even though I was so proud of my wordplay about "most often"!

gperciva:
Rebased

util/parsenum.h
cperciva: Umm... the details were fine.  My objection is that the top line says "check if optarg is NULL" and there's no optarg here.

gperciva:
Oops.  Rebased again.

PR #279, opened 2020-07-28
Check file descriptors with valgrind

gperciva:


gperciva:
I discovered valgrind's `--track-fds` while investing spiped.  Adding `closelog()` to `warnp.c` is an extremely nitpicky thing, but since it's the only change needed to make the extra level of testing pass, I figured it was worth it.

(There's no issues discovered by this check in scrypt or tarsnap, but a handful of ones in spiped.)

gperciva:
Huh, I was even testing with Ubuntu bionic, the same version as Travis-CI uses.  I suspect that this is something to do with their virtualization / container / CI setup.

PR #280, opened 2020-07-28
Check file descriptors with valgrind

gperciva:


gperciva:
I discovered valgrind's `--track-fds` while investing spiped.  Adding `closelog()` to `warnp.c` is an extremely nitpicky thing, but since it's the only change needed to make this extra level of testing pass, I figured it was worth it.

(There's no issues discovered by this check in scrypt, one issue already as a PR for tarsnap, and a couple in spiped.)

util/warnp.c
```diff
@@ -17,6 +17,10 @@ static void
 done(void)
```
cperciva, line 17: The comment on this function needs updating.

util/warnp.c
```diff
@@ -112,6 +116,10 @@ void
 warnp_syslog(int enable)
 {
 
+	/* If we're currently using syslog, close the file descriptor. */
+	if (use_syslog)
```
cperciva, line 120: Maybe worth making this `use_syslog && !enable`?  No point closing a descriptor if we're about to use it again.

util/warnp.c
gperciva: done

util/warnp.c
gperciva: Heh, I had that in the first place, then I figured that nobody would call `warnp_syslog(1)` twice so I might as well go with simpler code.

Done.

gperciva:
Rebased (so that I could update the commit with a missing).  I also expanded the rationale in that commit.

gperciva:
This is an expected test failure: this PR depends on #282 to avoid leaving an open file descriptor.

PR #282, opened 2020-08-02
Warnp closelog

gperciva:


gperciva:
This used to be part of #280, but I decided it would make more sense on its own.

PR #286, opened 2020-08-25
Comments

gperciva:


gperciva:
This does push a few comments above 80 chars, but I think that rewrapping those function APIs would complicate this PR too much.  I propose to handle those in a separate PR, along with fixing a few other line length violations.
(Some of the line lengths are unavoidable without unduly odd code constructs, such as changing the json tests.)

PR #288, opened 2020-08-26
Clarify aes comments and array sizes

gperciva:


gperciva:
I noticed that `alg/md5.h` also has a bunch of `uint8_t[16]` without the `static`, so if you like this PR then I'll apply it elsewhere in libcperciva.

cperciva:
`- then store the processed aes_state in ${out} on line 195.`

Line number is incorrect.

cperciva:
I'm fine with adding `static` in places like this, but didn't it cause issues with gcc 4.2.1 which we POSIXFAILed around?  Or was that a different issue.

gperciva:
They were both about `restrict`.  The recent one is that gcc 4.2.1 needs argument names in such declarations:
```
#ifdef POSIXFAIL_ABSTRACT_DECLARATOR
static int func(int ARGNAME[static restrict 1]);
#else      
static int func(int [static restrict 1]);
#endif
```

The old one was that some compilers needed an explicit `-std=c99` before they'd accept `restrict`.

That said, it's a good idea for me to check this in gcc 4.2.1 anyway.

cperciva:
Oh right, it was about declarators but only when they didn't have names.

gperciva:
... nope, apparently it's caused by `static`, not `restrict`.  Whoops.  At least you noticed before this was merged!

gperciva:
... I'm going to nuke the `static`.  I think it's still reasonable to use things like `[16]`, and the extra error checking would be nice, but not nice enough IMO to make it worth having
```
#ifdef POSIXFAIL_ABSTRACT_DECLARATOR
void crypto_aes_encrypt_block_aesni(const uint8_t in[static 16],
    uint8_t out[static 16], const void *);
#else
void crypto_aes_encrypt_block_aesni(const uint8_t[static 16],
    uint8_t[static 16], const void *);
```

cperciva:
Ok, your call.  I'm fine either way.

gperciva:
I believe that this is correct now (including the fixed commit message wrt the line numbers).  Tested on gcc 4.2.1 on OpenBSD as well.

gperciva:
I added a commit fixing `hash/encrypt`, like you suggested for aesctr in https://github.com/Tarsnap/libcperciva/pull/289#discussion_r478047404.

PR #291, opened 2020-08-30
Add `make perftest`

gperciva:


tests/perftest_make.sh
```diff
@@ -29,10 +29,13 @@ fi
...
-	printf "Error: ${data} already exists; will not overwrite\n"
-	exit 1
-fi
+#if [ -e "${data}" ]; then
+#	printf "Error: ${data} already exists; will not overwrite\n"
+#	exit 1
+#fi
+
+# Delete previous data.
+rm -f "${data}"
```
gperciva, line 38: I can't decide whether this should
1) carefully avoid overwriting any old data, or
2) cheerfully delete old data

I originally went with 1, but on second thought it takes less than a minute to recompile and re-run any of our perftests, so now I'm leaning towards 2.

gperciva:
This adds a `make perftest` target to the appropriate directories, which creates an easily-plottable text file.  The Makefile specifies how to extract the x and y values, while the `perftest_*.raw` file retains the full output in case one wants to double-check anything.

I think the overall idea is good, but there's a few details that I'm not certain about.

tests/perftest_compilers_single_value.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+# This script assumes that there is only one x-value (e.g., the
+# perftest does not check multiple buffer sizes).  In particular,
+# it was created for md5, sha1, and sha256.
```
gperciva, line 5: This file resulted from adjusting the `tests/md5/perftest.sh` script for the new `make perftests`, and then I realized that I could generalize it to also handle sha1 and sha256, which also contain that file (but probably just because we copied from md5 when we created those directories).

I'm not certain if it's worth keeping the file around, or modifying it, and even if we modify it, I'm not certain that it belongs in the same PR as `make perftests`.

gperciva:
I have a python & matplotlib script which produces a nice visualization of multiple data (e.g., comparing the 10 different bufsizes for `test_crypto_aesctr` across multiple commits and/or compilers), but that will go in a separate PR.

tests/perftest_make.sh
cperciva: I'm ok with deleting/overwriting old test results.  In fact, don't we do that for the regular tests anyway?  (For tests where the C code generates a file and then we compare it to "known good" output?)

gperciva: Yeah, but deleting output that takes 0.2 seconds to generate is a no-brainer compared to deleting output that takes a minute to generate.

Ok, going with delete.

cperciva:
I'm not sure it makes sense to me that "make perftest" doesn't produce any output.  If anything, I'd say that target should run the `foo -t` and perftest_make.sh should be hard-coded to use "make perftest" as the command it runs.

cperciva:
I'm not 100% sure about what you're aiming for here, though.

gperciva:
Ultimately, I want to be able to create a nice plot of either:
- `benchmark-compilers.sh`: the current directory with different compilers, or
- `benchmark-branch.sh`: benchmarks of every single commit in a branch.  (e.g., "here's a bunch of changes to `crypto_aesctr.c`; see how each one changes the speed at 1024 vs. 1 bufsize; is this worth it?)

The problem is that our "time trials" have slightly different formats:
```
$ ./test_crc32 -t
Hashing 536870912 bytes using hardware CRC32C...
... in 32768 blocks of size 16384:	0.22 s	2405.7 MB/s
... in 65536 blocks of size 8192:	0.15 s	3691.1 MB/s
```

vs.
```
$ ./test_md5 -t
MD5 time trial. Digesting 100000 10000-byte blocks ... done
Digest = 766a2bb5d24bddae466c572bcabca3ee
Time = 3.230208 seconds
Speed = 309577587.573308 bytes/second
```

So if I want to parse these to extract the bufsize and speed (which is probably what we want to plot), we need per-directory instructions:
- for `test_crc32`, find all lines containing "blocks" and look at columns 7 and 10.  Remove any colons from the output.
- for `test_md5`, find all lines containing "Speed", use a fixed value of 10000 and column 3.

Another option would be to format `test_md5` (and sha1 & sha256) like the other time trials:
```
$ ./test_md5 -t
MD5 time trial.
... in 10000 blocks of size 10000:	3.23 s	309.6 MB/s, digest = 766a2bb5d24bddae466c572bcabca3ee
```

then we could use a fixed "columns 7 and 10, and strip colons" for all "perftest-enabled" directories.  I thought that you'd prefer to have per-directory instructions in the `Makefile`, but I'd be happy to standardize the time trial output instead if you'd prefer that one.
(The tentative rewritten output for `test_md5` does exceed 80 chars, but I think that having "one set of perftest on one line" is more important than 80 chars, to enable non-ridiculous grep and/or scripting.)

(`test_mpool.sh`'s output doesn't match that format, but it's part of `make test` so I'm not worrying about that.)

gperciva:
I'll add an example of the "perftests of all commits on this branch" tomorrow.

gperciva:
> I'm not 100% sure about what you're aiming for here, though.

So the overall goal is to easily & automatically produce plots such as:
https://github.com/Tarsnap/libcperciva/pull/304#issuecomment-696984826

At the moment there's 5 PRs that build towards that; you can see their dependencies on the usual webpage.

gperciva:
You might wonder "what's the point of `make perftest`, instead of calling `test_${foo} -t` in a script?".  It's purely a means of abstracting away the `cmd -t` part.  Right now, all of our perftests are triggered by a single binary with the `-t` option, but it's possible that at some point we might to run a script or have other options.

PR #292, opened 2020-09-05
tests/crypto_aesctr: prepare for nonces in C code

gperciva:


tests/crypto_aesctr/main.c
```diff
@@ -364,12 +364,12 @@ selftest_cases(const struct testcase * tests, size_t num_tests, uint64_t nonce)
 		printf("Computing %zu-bit AES-CTR of \"%s\"...",
 		    keylen * 8, tests[i].plaintext_str);
 
-		/* Plaintext length. */
+		/* Plaintext length and prepare input. */
```
cperciva, line 370: Grammar?

tests/crypto_aesctr/main.c
```diff
@@ -339,19 +339,22 @@ selftest_unaligned_access(size_t keylen)
 }
 
 static size_t
-selftest_cases(const struct testcase * tests, size_t num_tests, uint64_t nonce)
+selftest_cases(const struct testcase * tests, size_t num_tests, uint64_t nonce,
```
cperciva, line 342: The commit text is a bit weird here.  Also, we're testing with nonzero counters already since the second block of data (for >16 byte inputs) is xored with AES(00000001).

Normal usage of CTR mode is "each message has a different nonce, and the counter increments as you move through the message"; for repeatedly encrypting the same data it would make sense to use a different nonce value each time (but have the counter reset to zero at the start of each buffer).

gperciva: Hmm, I forgot about having multiple blocks of data in a single test message.  I'll need to rethink this.

PR #294, opened 2020-09-05
tests/crypto_aesctr: check nonce

gperciva:


tests/crypto_aesctr/main.c
```diff
@@ -410,6 +474,17 @@ selftest(void)
 	if (selftest_cases(tests_256, num_tests, 0))
 		failures++;
 
+	/* Test with nonce=*2 (decimal). */
```
cperciva, line 477: I'm confused by "nonce=*2 (decimal)" here.  It looks like you've set a nonce of "2222222" in base 1024?

cperciva: I mean, there isn't anything especially wrong with this value, but it's a strange choice.  I would have gone with something 0x1234567890ABCDEF.

gperciva: I wanted something higher than `0x8` to check that it wasn't `int64_t` or something like that.  How about `0xfedcba987654321`?

cperciva: I think you mean 0xfedcba9876543210, not 0xfedcba987654321, but sure.

I think the nonce value you were calculating was 0x4010040100401004, so not high enough to run into signed issues anyway...

gperciva:
Updated with nonce = `0xfedcba9876543210`, ready for merge.

PR #295, opened 2020-09-05
Tests minor cleanup in time trials

gperciva:


tests/md5/main.c
```diff
@@ -49,20 +49,20 @@ perftest(void)
 	for (i = 0; i < BLOCKCOUNT; i++)
 		MD5_Update(&ctx, buf, BLOCKLEN);
 	MD5_Final(hbuf, &ctx);
-	hexify(hbuf, hbuf_hex, 16);
```
gperciva, line 52: Did you want to keep the `hexify()` inside the timing loop?

cperciva: It's fast enough to be insignificant; but no, it doesn't belong inside the timing code.

PR #298, opened 2020-09-07
tests: remove `rev` for POSIX compatibility

gperciva:


cperciva:
Are you sure this can't be achieved just with shell parameter expansion?  e.g. to remove a suffix ${foo%%${bar}}

gperciva:
Do you dislike the `basename`, or the `tail`?

Ultimately, we want to go from

    /data/td/src/libcperciva/tests-output/01-crc32-00.exit

to

    /data/td/src/libcperciva/tests-valgrind/01-crc32-00

A different option is to do

    base_with_num=$(basename "${exitfile}" ".exit")
    echo "${out_valgrind}/${base_with_num}"

Of course, we could use `sed` and `tail`, or `sed` and shell parameter expansion, or...?

gperciva:
I did double-check that using `basename` to remove a suffix is in POSIX:
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/basename.html

cperciva:
It was the `tail -c -3` which felt most awkward to me.  I'd prefer your alternative of using `${out_valgrind}`; alternatively, I think

```
basefile=${exitfile%%.exit}
echo "${val_basename}-${basefile##*-}"
```

would work, I think?  (And the string manipulation would be entirely within `sh`.)

gperciva:
Updated PR, checked on linux (spiped, for valgrind on multiple files) and solaris.

tests/shared_test_functions.sh
```diff
@@ -269,8 +269,8 @@ setup_check_variables() {
 get_val_basename() {
 	val_basename=$1
```
cperciva, line 270: Looks like `val_basename` is not being used now?

gperciva: Huh.  I'll revisit this later.

gperciva:
Updated, ready for merge, checked on freebsd, linux, and solaris.

cperciva:
Am I right in assuming that `out_valgrind` is a global variable which we can assume is set appropriately?

And did you mean to remove the "draft" label?

gperciva:
Yes, currently we assume that `${out_valgrind}` is set (for example, in `prepare_directories()`).  I should probably check that explicitly, but this PR does not change the assumption-ness of `${out_valgrind}`.

PR #300, opened 2020-09-09
Posix close

gperciva:


gperciva:
Very much a draft; the test doesn't appear to do anything on linux or freebsd.  According to musl[1], Linux doesn't support `EINTR` for close.

[1] https://git.musl-libc.org/cgit/musl/commit/?id=8708e137d68480f4b996bfec2cd2ca596f1105d7

util/posix_close_normal.c
```diff
@@ -0,0 +1,139 @@
...
+#include "noeintr.h"
+#include "warnp.h"
+
+#include "posix_close.h"
+
+/*
+ * The two files pthread_close_normal.c and pthread_close_pthread.c should be
+ * identical, other than the #define USE_PTHREAD line.
+ */
+#define USE_PTHREAD 0
```
gperciva, line 20: I haven't included `pthread_close_pthread.c` in this commit, since I'm sure that I'll need to make a bunch of fixes to `posix_close_normal.c` so I might as well make the copy later.

tests/posix_close/open_close.c
```diff
@@ -0,0 +1,143 @@
...
+	int numbytes;
+	volatile int * stop;
+};
+
+static void
+alarm_handler(int signo)
+{
+
+	(void)signo; /*	UNUSED */
+	alarmed++;
```
gperciva, line 28: I didn't know what to do with the signals with which I was trying to interrupt `posix_close()`, but I figured that counting them would at least show that they were happening.

tests/posix_close/send_alarms.c
```diff
@@ -0,0 +1,66 @@
...
+	volatile int * stop;
+};
+
+static void *
+workthread_send_alarm(void * cookie)
+{
+	struct send_alarms * SA = cookie;
+
+	while (*SA->stop == 0) {
+		if (pthread_kill(SA->to_thr, SIGALRM)) {
```
gperciva, line 22: I'm not at all clear about whether this should be `pthread_kill()` or `raise()`.

tests/posix_close/main.c
```diff
@@ -0,0 +1,77 @@
...
+		warn0("%s should be 0 or 1", argv[2]);
+		goto err0;
+	}
+
+	/* Open, read 0 or 1 bytes, then close the file. */
+	if (start_repeated_open_close(filename, numbytes, &stop_open_close,
+	    &open_close_thread))
+		goto err0;
+
+	warn0("sleep for first thread to start");
```
gperciva, line 46: The question here is "how to wait for the thread to start?".  As you might recall, I wrote a `pthread_create_blocking()` that we didn't end up using in `spipe`.  That seems a bit like overkill, though, especially since this is just a test and there's no harm (?) in slapping a `sleep(1)` in here.  (Although 1 full second is awfully long.)

util/posix_close.h
```diff
@@ -0,0 +1,17 @@
...
+/*
+ * Use a #define in order to avoid namespace collisions with anyone else's
+ * posix_close() (e.g., musl).
+ */
+#define posix_close libcperciva_posix_close
+
+/**
+ * posix_close(fd, flags):
+ * Close a descriptor, but looping until completion if interrupted by a signal.
+ * ${flags} must be 0.
```
gperciva, line 13: This aims to match the upcoming `posix_close()` function:
https://www.austingroupbugs.net/view.php?id=529
comment 0001200.

util/posix_close_normal.c
```diff
@@ -0,0 +1,129 @@
...
+	uint8_t cookie_recv[MAGIC_COOKIE_LEN];
+	int s[2];
+	int i;
+	ssize_t r;
+
+	/* Sanity check. */
+	assert(flags == 0);
+
+	/* Create a socket pair. */
+	if (socketpair(AF_UNIX, SOCK_STREAM, 0, s)) {
```
cperciva, line 51: This needs to be done only once, and the socket pair is kept open for the lifetime of the process (or rather, until an atexit handler cleans it up).  Otherwise you have problems when you try to close the socket pair.

util/posix_close_normal.c
```diff
@@ -0,0 +1,129 @@
...
+ */
+#define USE_PTHREAD 0
+
+#if USE_PTHREAD
+#include <pthread.h>
+
+static pthread_mutex_t mutex;
+#endif
+
+#define MAGIC_COOKIE_LEN 4
```
cperciva, line 28: Needs to be 32 bytes.

util/posix_close_normal.c
```diff
@@ -0,0 +1,129 @@
...
+
+	/* Write the cookie to s[0]. */
+	if (noeintr_write(s[0], cookie_write, MAGIC_COOKIE_LEN) == -1) {
+		warnp("noeintr_write");
+		goto err1;
+	}
+
+	/* In a multithreaded environment, lock the mutex here. */
+#if USE_PTHREAD
+	if ((rc = pthread_mutex_lock(mutex)) != 0) {
```
cperciva: The mutex needs to be initialized before it can be locked.  And, obviously, the initialization should only happen once.

util/posix_close_normal.c
```diff
@@ -0,0 +1,129 @@
...
+	/* Sanity check. */
+	assert(flags == 0);
+
+	/* Create a socket pair. */
+	if (socketpair(AF_UNIX, SOCK_STREAM, 0, s)) {
+		warnp("socketpair");
+		goto err0;
+	}
+
+	/* Generate a random cookie. */
```
cperciva, line 56: We should use actual entropy here.  Yes, recursing makes this hard, but if detect that we're recursing we can record the descriptor and close it later.

util/posix_close_normal.c
gperciva: Are you thinking of having an explicit `_init` function, or should this implicitly detect when it needs to be initialized?  (or both?)

util/posix_close_normal.c
gperciva: Just to confirm: you want this to be `entropy_read()`, and that function should itself call `posix_close()`?

util/posix_close_normal.c
cperciva: It should detect when it needs to be initialized.  Look at `pthread_once`.

util/posix_close_normal.c
cperciva: Yes.

util/posix_close.h
cperciva: Hmm.  I was thinking that this should be grouped with the other "noeintr" functions; it's not exactly a drop-in replacement for `posix_close` since it's *way* more complicated.  Let me think about this...

util/posix_close_normal.c
gperciva: Alternative solution: https://github.com/Tarsnap/libcperciva/pull/301

util/posix_close_normal.c
gperciva: Hmm, that would presumably require linking to `pthread`.  I was going to do this instead:
```
static int s[2] = {-1, -1};

int
posix_close(int fd, int flags)
{
...
        /* Initialize (if needed). */
        if ((s[0] == -1) && posix_close_init()) {
                warn0("posix_close_init");
                goto err0;
        }
```

cperciva: And what happens if both threads run simultaneously and see `s[0] == -1`?

Doing initialization is *exactly why pthread_once exists*.

util/posix_close_normal.c
cperciva: Yeah, that's a better approach than recursing.

util/posix_close_normal.c
gperciva: But we're already wrapping the whole thing in a mutex.  And pthread_once() doesn't let us return a value [1].  This looks weird:

```
/* pthread_once() doesn't allow a return value from init_routine. */
static int init_ret = 0;

noeintr_close(int fd)
{
...
#if USE_PTHREAD
    if ((rc = pthread_mutex_lock(&mutex)) != 0) {
        ...
#endif

     /* Initiailize */
#if USE_PTHREAD
    if (pthread_once(&once_control, noeintr_close_init())) {
         ...
#else
    if (init_status == 0)
        noeintr_close_init();
#endif

    /* Check for any errors during initialization. */
    if (init_status != 1)
```

If we rely on the mutext to ensure that only one thread is running the code at once, then we can do:
```
noeintr_close(int fd)
{
...
#if USE_PTHREAD
    if ((rc = pthread_mutex_lock(&mutex)) != 0) {
        ...
#endif

     /* Initiailize */
    if ((init_status == 0) && (noeintr_close_init()) {
        warn0("noeintr_close_init");
        goto err1;
    }
```

which is much more idiomatic.  The init function has `socketpair`, `atexit`, `entropy_read`, and `noeintr_write`, so I don't think that checking its success is optional.

[1] https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_once.html

gperciva:
Continued in #312.

PR #301, opened 2020-09-10
Add entropy_read_init(), _fill(), _done()

gperciva:


gperciva:
This is aimed at helping posix_close (or whatever we call it).  I think it's a neat solution, but you might hate it.  :)

Rather than "detect that we're recursing" between posix_close() and entropy_read() [1], we could use this new function which will reuse `fd` and defers clean-up to the callee.  In this case, the clean-up would happen in `posix_close_atexit`.

[1] https://github.com/Tarsnap/libcperciva/pull/300#discussion_r486048716

gperciva:
I did consider a more invasive change of making `entropy_read()` always use the same fd, and close that via `atexit()`.  However, that would run into needing `pthread` stuff to ensure that it's done cleanly.

I'm also aware that at some point in the future, `entropy_read()` might use platform-specific functions like `getentropy()` or `getrandom()`, which would eliminate the need for a file descriptor.  I wrote the function API with that in mind, but I'm aware that it makes `entropy_read_reuse()` less attractive.

(I still think it's better than dealing with recursion in `posix_close()`, though.)

gperciva:
Updated with a new API: `_open`, `_fill`, `_close`.

I've arranged this PR with multiple minimal commits, but I anticipate that we'll squash them all together with rebase.  If you'd like them to stay separate, I'm happy to rewrite the commit messages accordingly.

util/entropy.c
```diff
@@ -16,61 +18,157 @@
...
+ * using the file descriptor in ${er}.
+ */
+static int
+entropy_read_fill(struct entropy_read * er, uint8_t * buf, size_t buflen)
+{
+	ssize_t lenread;
+
+	/* Sanity checks. */
+	assert(er != NULL);
+	if (buflen > SSIZE_MAX) {
```
gperciva, line 69: What would you think about replacing this with
```
assert(buflen <= SSIZE_MAX);
```

with the general idea of "use `assert()` for programming errors and don't bother trying to quit gracefully".

I note that `entropy.c` is the only file `.c` in libcperciva, scrypt, spiped, or kivaloo that uses the term "Programmer error".  (Admittedly there's a ton of them in tarsnap.)

util/entropy.c
```diff
@@ -16,61 +18,157 @@
...
 
+struct entropy_read
+{
+	int fd;
+};
+
 /**
- * entropy_read(buf, buflen):
- * Fill the given buffer with random bytes provided by the operating system.
+ * entropy_read_open():
```
gperciva, line 27: Are you _sure_ that you want this to be `_open()` and not `_init()`?  I'm imagining a future where every OS has adopted `getentropy()` and we don't need to mess around with file descriptors.

gperciva: FWIW, here's the "add `getentropy()` to POSIX" proposal:
https://www.austingroupbugs.net/view.php?id=1134

cperciva: It could be `init`/`done` if you prefer.  We can't use `free` since that's reserved for functions which can't fail.

util/entropy.c
```diff
@@ -16,61 +18,157 @@
...
+ * Initialize the ability to produce random bytes, and return a cookie.
```
cperciva, line 28: The words `operating system` are important here.  I don't want to lose the distinction between "ask the OS for entropy" and "get entropy from our internal entropy pool" (in `crypto_entropy`).

util/entropy.c
cperciva: Switching to `assert` is good.  Some or all of the ones in tarsnap should probably be changed too.

util/entropy.c
```diff
@@ -16,61 +18,157 @@
...
+	return (0);
+
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+/**
+ * entropy_read_close(er):
+ * Close any file descriptor in ${er} and free memory.
```
cperciva, line 104: I would say "resources" rather than calling out "file descriptor" and "memory" specifically; a future Windows version of this might obtain a Handle, for example.

gperciva: Hmm, by "memory", I was referring to `struct entropy_read * er` itself.  I've now got
`Release any resources used by {er}.` which sounds a bit too vague to me, but OTOH it's probably good enough for the "clean-up" function?

gperciva:
Updated with the rebase.  I can go back to the non-rebased version if you want.

util/entropy.c
```diff
@@ -17,56 +18,150 @@
  * XXX win32 where there is no concept of /dev/urandom.
  */
 
+struct entropy_read
```
cperciva, line 21: The `{` should be on the same line as this.

Also, it kind of feels like this should be `entropy_reader`, in the sense that a structure is a noun rather than a verb.  Up to you, just thought I'd mention it.

cperciva: While I'm here, I feel like this structure deserves some sort of comment.  Something along the lines of "Entropy reader state.  At present it holds a file descriptor open to /dev/urandom but in the future this structure may gain other OS-dependent state, e.g. a Windows Handle.".

gperciva: Added a REBASE commit.  I went with `struct entropy_read_cookie`.  (I was sorely tempted to call it `entropy_cookie`, but since the functions all start with `entropy_read`, I couldn't justify it to myself)

I tweaked your proposed comment with "file descriptor for /dev/urandom", since at the moment that we allocate the cookie, the file descriptor is not (yet) open to /dev/urandom.

cperciva:
LGTM, please rebase.

gperciva:
Rebased, ready for merge.

PR #302, opened 2020-09-12
Standardize timing tests with perftests.h

gperciva:


gperciva:
I got fed up with finding differences in the perftest code.  (e.g., is it `bufsize`, or `blocksize`, or `perftests[i]`?  When I copied code that used `num_hashes`, did I remember to change it to `num_encrypts`?)

I'm confident that the general concept is good; I mean, just look at `tests/crypto_aesctr/main.c`:

```
        /* Run operations. */
        PERFTESTS_LOOP_OUTER_BEGIN
                /* Time it. */
                PERFTESTS_LOOP_INNER_BEGIN(2)
                        crypto_aesctr_buf(key_exp, perftests_i,
                            perftests_buf, perftests_buf, perftests_bufsize);
                PERFTESTS_LOOP_INNER_END(2)
        PERFTESTS_LOOP_OUTER_END
```

However, there's a number of details that will need attention and thought.  I'll comment on those individually.

tests/crypto_aesctr/main.c
```diff
@@ -241,80 +241,40 @@ perftest(void)
...
-		/* Print results. */
-		printf("... in %zu blocks of size %zu:\t%.06f s\t%.01f MB/s\n",
-		    num_blocks, bufsize, delta_s,
-		    (double)bytes_to_encrypt / 1e6 / delta_s);
-		fflush(stdout);
-	}
-
-	/* Clean up. */
-	free(largebuf);
+	PERFTESTS_LOOP_OUTER_BEGIN
```
gperciva, line 263: I considered calling this something like `PERFTESTS_BUFSIZE_CHOOSE_BEGIN` (although not that precisely!), but I haven't come up with anything as clear as "begin outer loop".

Another nitpick might be the word order.  I went with "loop outer begin" so that if one were to sort the macros of `perftests.h` alphabetically, related functions would appear next to each other.  (I picked up that trick from naming lilypond commands, thinking of the function name index.)

That said, I cannot think of a reason one would want to sort the macros alphabetically.  If we had a dozens of them in the same file, there might be some scenario in which that would be useful.

tests/crypto_aesctr/main.c
```diff
@@ -241,80 +241,40 @@ perftest(void)
...
+		/* Time it. */
+		PERFTESTS_LOOP_INNER_BEGIN(2)
```
gperciva, line 265: Likewise, this might be better called `PERFTESTS_TIMING_LOOP_BEGIN`.

The number is used for `goto`, with
```
#define PERFTESTS_LOOP_INNER_BEGIN(err_label_num)                       \
...
                goto err ## err_label_num;                              \
```

I also considered defining `err_perftests_inner:` instead, but I figured that would be too much of a departure from our normal style.

I also (just now) considered saying "screw it, who cares about unallocated memory from a failing perftest" and just making every error jump to `err0`.

gperciva:
If / when we agree on this approach in theory, I'll apply it to the other perftests.

util/perftests.h
```diff
@@ -0,0 +1,73 @@
...
+		goto err ## err_label_num;				\
+	}								\
+	perftests_delta_s = timeval_diff(perftests_begin, perftests_end); \
+	printf("... in %zu blocks of size %zu:\t%.02f s\t%.01f MB/s\n",	\
+	    perftests_num_buffers, perftests_bufsize, perftests_delta_s,\
+	    (double)PERFTESTS_BYTES / 1e6 / perftests_delta_s);		\
+	fflush(stdout);
+
+#define PERFTESTS_LOOP_OUTER_END					\
+	}
```
gperciva, line 71: This also needs to `free(perftests_buf)`.

(I was experimenting with `uint8_t perftests_buf[PERFTESTS_MAX_BUFSIZE];` but changed my mind about that a few minutes before uploading the PR, and I forgot to fix that until I saw it on github.)

tests/crypto_aesctr/main.c
```diff
@@ -241,80 +241,40 @@ perftest(void)
...
+			crypto_aesctr_buf(key_exp, perftests_i,
+			    perftests_buf, perftests_buf, perftests_bufsize);
+		PERFTESTS_LOOP_INNER_END(2)
+	PERFTESTS_LOOP_OUTER_END
```
gperciva, line 270: In this less-verbose version of the `crypto_aesctr` perftest, it's more obvious that I forgot to `crypto_aes_key_free(key_exp);`.

cperciva:
I like the idea of extracting the boilerplate; I hate the idea of doing it via macros.

Can you restructure this to have a
`int perftest(size_t nbytes, size_t *sizes, size_t nsizes, int func(uint8_t * buf, size_t buflen))` function instead?

gperciva:
That was the first thing I considered, but:
- `test_crc32`: would have to lose the `delta_s_overall += delta_s;`.  (well, I suppose we _could_ use a static variable for that, or refactor the timing to have separate variables for the "overall begin / end" times)
  You're probably ok with that, though.
- calling a function might add time.  I mean, with macros, we're timing
```
for (blah)
    crypto_aesctr_buf(blah);
```
whereas with a callback function, we'd be timing

```
for (blah)
    (func)(blah);
```

If we define `static void func() inline`, can we rely on the compiler inlining it?  Bearing in mind that the loop itself would be in `perftests.c` and the `func()` would be defined in `tests/*/main.c`?
... or maybe it would be ok as long as we defined the function in `perftests.h`?  That would force the code to be in the same object file, at least?

cperciva:
I'm ok with losing the "overall time".

I don't think the overhead of a function pointer call is that bad, but if you're worried about it we can make the API `func(uint8_t * buf, size_t buflen, size_t niters)`.

gperciva:
Revised branch, with a `perftest.c` instead of using header-only macros.

(Potentially) notable timing differences, based on `NUM_REPS=3`.  In most cases these are tiny differences (e.g., 325.2 MB/s vs. 321.9 MB/s), but there's no overlap in the times.  The timing output of many tests show up as a horizontal line, meaning that the box-plot of 3 repetitions are extremely close or identical.  If you want, I could add automatic statistical significance, like `ministat(1)`.

For crc32 + clang, this PR produces faster or equal-speed code with the sole exception of `clang -O2` with a blocksize of 256.  For sizes of 512 and above, there's no change of speed.

crc32-clang

For crc32 + gcc, this PR always produces faster code.  For blocksizes of 64 to 256, it's significantly faster.  No idea why.  For 512 and above, there's (almost) no change of speed.  (very slight improvement with 1024 and `clang -O2`)

crc32-gcc

For crypto_aes and crypto_aes_ctr, speed is unchanged.

For md5 + clang, this PR produces slightly slower code.  No change at all with gcc.

md5

For sha1 + clang, PR is faster; sha1+ gcc is unchanged or slower.

sha1

For sha256 + clang, slightly slowe ror unchaged; for `gcc -O2` it's significantly faster, `gcc -O3 -march=native` is slightly slower.

sha256

gperciva:
Just to double-check, I ran another set of benchmarks on the git master, and they match the "a" set.  So the differences seen here are definitely due to the code, not due to randomness in the benchmarking.

I also split up the "use perftest.h" commits in case that helped with reviewing, but they should probably all be smooshed together before merging.

cperciva:
The differences you're seeing look like they're on the scale which often results from cache alignment and similar issues.  Details like "this code moved around and now this branch collides with that branch and is predicted less efficiently" et cetera.  The 1% difference could be in the opposite direction if object files are passed to the linker in a different order, never mind having a real program using these functions.

Benchmarking is hard. :-/

util/perftest.c
```diff
@@ -0,0 +1,116 @@
...
+
+#include "perftest.h"
+
+/**
+ * perftest_buffers(num_bytes, sizes, num_sizes, num_warm_up, init_func, func,
+ *     cookie):
+ * Time using ${func} to process ${num_bytes} bytes in blocks of ${sizes}.
+ * Before timing any block sizes, process ${num_warm_up} blocks of the maximum
+ * size in ${sizes}.  Before timing each block size, call
+ * ${(init_func)(cookie, buffer, buflen).  Print the time and speed of
```
cperciva, line 18: mismatched `{`

util/perftest.c
```diff
@@ -0,0 +1,116 @@
...
+		if (max_buflen < sizes[i])
+			max_buflen = sizes[i];
+	}
+
+	/* Allocate buffers. */
+	if ((buf = malloc(max_buflen)) == NULL) {
+		warnp("malloc");
+		goto err0;
+	}
+	if ((delta_s = malloc(max_buflen * sizeof(double))) == NULL) {
```
cperciva, line 50: s/max_buflen/num_sizes/ ?

util/perftest.c
```diff
@@ -0,0 +1,116 @@
...
+#include "monoclock.h"
+#include "warnp.h"
+
+#include "perftest.h"
+
+/**
+ * perftest_buffers(num_bytes, sizes, num_sizes, num_warm_up, init_func, func,
+ *     cookie):
+ * Time using ${func} to process ${num_bytes} bytes in blocks of ${sizes}.
+ * Before timing any block sizes, process ${num_warm_up} blocks of the maximum
```
cperciva, line 16: I think it probably makes more sense to have `num_bytes_warmup` bytes rather than `num_warm_up` blocks.

Also, I would call the parameters `nbytes`, `nsizes`, and `nbytes_warmup`.

tests/crypto_aes/main.c
```diff
@@ -44,8 +44,11 @@ static const struct testcase perftestcase = {
 	    "00112233445566778899aabbccddeeff",
 	    "a7cbdceb16a2b37924794d01fa4a5796"
 };
+static const size_t perfsizes[] = {16};
+static const size_t num_perf = sizeof(perfsizes) / sizeof(perfsizes[0]);
 static const size_t bytes_to_encrypt = 1 << 30;	/* approx 1 GB */
 
+
```
cperciva, line 51: ?

tests/crypto_aesctr/main.c
```diff
@@ -268,82 +269,33 @@ perftest(void)
...
-		    num_blocks, bufsize, delta_s,
-		    (double)bytes_to_encrypt / 1e6 / delta_s);
-		fflush(stdout);
-	}
+	/* Time the function. */
+	if (perftest_buffers(bytes_to_encrypt, perfsizes, num_perf, 100000,
+	    perftest_init, perftest_func, key_exp)) {
+		warn0("perftest_buffers");
+		goto err1;
+ 	}
```
cperciva, line 289: Whitespace?

util/perftest.c
```diff
@@ -0,0 +1,116 @@
...
+	if (func(cookie, buf, max_buflen, num_warm_up))
+		goto err2;
+
+	/* Run operations. */
+	for (i = 0; i < num_sizes; i++) {
+		/* Configure and sanity checks. */
+		buflen = sizes[i];
+		assert(buflen > 0);
+		num_buffers = num_bytes / buflen;
+		assert((num_bytes % buflen) == 0);
```
cperciva, line 67: I would be inclined to relax this -- it would allow #309 to have more buffer sizes (I would probably go up to 1024).

gperciva: If we do that, then different buffer sizes wouldn't be comparing the same amount of data.  (Also, at the moment the speed is calculated based on `num_bytes`, but I could change that to `buflen * num_buffers` if necessary)

For #309, I was thinking that we should tweak the number of bytes so that it's a multiple of any buffer size we want to check.  If that means that we change a bufsize of 10000 to 8192 or 9728, that sounds fine to me?

util/perftest.c
gperciva: Ok.  At the moment, we have:
```
static int
perftest_func(void * cookie, uint8_t * buf, size_t buflen, size_t num_buffers)
```
in `tests/`.  Should I rename that to `nbuffers`?

util/perftest.c
gperciva: BTW, that was an empirically-added `assert()`.  Namely, I added some more buffer sizes to one of the tests, didn't realize that they weren't an integer division, and spent 15 minutes trying to figure out why those buffer sizes were faster than the other ones.

(I think it was in crypto_aesctr, and I was testing the 1028-byte buffers we get from spiped.)

util/perftest.c
cperciva: I would probably go with `nbufs`.

cperciva: Or possibly `nreps` since it's not the number of buffers but rather the number of times we should repeat on the same buffer, yes?

util/perftest.c
cperciva: Yes, I meant to change the speed calculation to use `buflen * num_buffers`.  And this should result in the same MB/s speeds, even if a (slightly!) different number of bytes is processed.

cperciva:
Please rebase.

gperciva:
I'm going to have another PR before this one, adjusting `perftest_func()` in existing code.  But yes, once that's done I'll rebase this mess.  :)

PR #303, opened 2020-09-22
Perftests scripts

gperciva:


gperciva:
Minor note: the `--march=native` isn't recognized on gcc 5.5 on NetBSD (and I'm sure OpenBSD as well).

PR #304, opened 2020-09-22
Perftests visualize

gperciva:


gperciva:
So the overall goal of all these perftest PRs is that I can do:
```
cd tests/crc32
../perftest_compilers.sh gcc clang
../perftest_plot.py
```
fig-dots

It also supports log-y, and box plots:
```
../perftest_plot.py --boxplot
```

fig-box-overall

also zooming in:
fig-dots-closeup


(... and no, I haven't yet looked into why `-O3 -march=native` is so much worse than `-O2` on both gcc and clang for buffer sizes of 512 to 4096.  But it certainly looks worthy of investigation!)

Another upcoming script: generate perftests for each commit on a branch.

gperciva:
I'm now seeing completely different numbers from crc32, but I've fixed a few bugs in the plotting -- in particular, making sure that I don't have a potentially random order of keys from a dictionary.  So take the details of the above plots with a huge grain of salt.  I'll re-create them before this PR is up for review.

tests/perftest-plot.py
```diff
@@ -0,0 +1,200 @@
...
+
+import argparse
+import glob
+import os.path
+
+import matplotlib.pylab
+import numpy
+
+# Constants for libcperciva's perftests
+_COLUMN_NAMES = ["blocksizes", "speed MB/s"]
```
gperciva, line 13: As indicated, these column names are hard-coded for libcperciva's (current) perftests.  At the moment, I'm leaning towards using this script for the spiped perftests as well, but using a command-line option to override the column names when necessary.

Another option is to write a YAML file with metadata about the particular perftests, but at the moment that feels like a prematurely over-engineered solution.  Also, I could just have make a copy of this file for spiped, with minor difference for project-specific info (like we do for `Makefile.BSD`).

PR #305, opened 2020-09-26
Move setgroups_none() into a separate .c file

gperciva:


gperciva:
Fixes the problem of compiling spiped on Solaris https://github.com/Tarsnap/spiped/issues/279.

util/setgroups_none.c
```diff
@@ -0,0 +1,60 @@
+/* We use non-POSIX functionality in this file. */
```
gperciva, line 1: This is the only file with non-POSIX functionality in libcperciva.  In scrypt, we isolated such code to `lib-platform/` instead of `lib`.  I figured that you wouldn't want to bother doing that here, but it's easy for me to tweak the build system if you do want that separation.

PR #306, opened 2020-09-28
Fix setuidgid

gperciva:


util/setuidgid.c
```diff
@@ -213,9 +213,15 @@ string_extract_user_group(const char * combined, char ** username_p,
 	/* Sanity check. */
 	assert(combined != NULL);
 
-	/* Search for ":" and get string length. */
-	pos = strcspn(combined, ":");
+	/* Reject a 0-length string. */
 	len = strlen(combined);
+	if (len == 0) {
```
cperciva, line 218: Not sure I like this.  I would handle this case by changing the conditions below:

1. `if (pos == 0 && len > 0)` for detecting `:group`,
2. `if (pos != len)` for detecting `user:group`,
3. `if (len > 0)` for detecting `user`,

and put a

```
else {
    warn0("Invalid user:group string: \"%s\", combined);
    goto err0;
}
```

at the end.

cperciva: It amounts to the same thing but I feel like it's clearer to be specifying conditions which correspond to the different cases rather than special-casing empty strings.

gperciva:
Updated with the refactoring (using slightly more stringent conditions for the cases), but now `gcc -fanalyzer` is complaining again.

The error message is confused, though -- it complains a double free coming from `free(groupname); free (username);`, and points to the `/* Username case */` as being the source of those variables.  I can make the error disapear by adding `assert(*groupname != *username);` in the `/* Username:groupname */` case, or by adding a `if (pos > 1022) goto err0;` in the same section (?).

`gcc -fanalyzer` is relatively new, so I'm content to ignore it for now.  It produces other bogus warnings in our code, so we can't add it to our automatic checks anyway.

As a side effect, the refactoring resovles the codeQL warning.

cperciva:
I think https://github.com/Tarsnap/libcperciva/tree/simplify-setuidgid is cleaner than this.  All of the conditionals comparing `pos` and `len` give me a headache.

PR #307, opened 2020-10-05
aws_sign: use gmtime_r()

gperciva:


gperciva:
History of `gmtime_r()` in POSIX:
- Issue 5: included for alignment with the POSIX threads extension
- Issue 6: marked as part of the Thread-Safe Function soption.
- Issue 7: moved to Base.

History of `gmtime_r()` in a few OSes:
- linux: added to glibc between 1.01 and 1.09.3, actual date 1995-09-18
- openbsd: since 2.5, 1999
- netbsd: since 1.5, 2000
- solaris: included before 2.4, 1994

FWIW, kivaloo is the only project which uses `aws_sign.c`.

cperciva:
FWIW, kivaloo is not the only project; it just happens to be the only *Tarsnap* project.  I also use `aws_sign` in https://github.com/cperciva/qmail-ses and https://github.com/cperciva/bsdec2-image-upload

PR #308, opened 2020-10-09
Add perftest_init, perftest_func

gperciva:


gperciva:
This prepares for a `perftest.h` (#302), which would use a standard
```
        if (perftest_buffers(num_bytes, perfsizes, num_perf, 8000,
￼	    perftest_init, perftest_func, NULL)) {
￼		warn0("perftest_buffers");
￼		goto err0;
￼	}
```
format for all perftests.

In theory, adding a `perftest_init()` and `perftest_func()` shouldn't change the timing.  In the plots, that each pair of boxplots should indicate the same speed -- ideally, we should have two flat lines next to each other.

There's a few oddities, though.  Here's some plots (using #304), where "antecedent" means `master` and "subsequent" means "using `perftest_func()`".  [In the future, I'll just use `a` and `b` since the legend's width is annoying.]

In the CRC32 plots (split into `gcc` and `clang` because otherwise there's too much data), we can see some timing variation in 16 to 1024 blocksizes.  That's part attributable to the fact that the new code does `memset(buf, 0, bufsize)`; the original crc32 perftest only zeroed out the buffer before doing the "warm up", and simply re-used the messy buffer for all the perftests.  That said, if I try removing the `memset()` from `perftest_init()` and only do it once, there's still a bit of a performance hit.

crc32-clang
crc32-gcc

For md5, `perftest_func` is faster with `clang -O3 -march=native`, and variably slower for `gcc -O2`.

md5

For sha1, it's noticeably slower for `clang`, faster for `gcc -O2`, and slightly slower for `gcc -O3 -march=native`.

sha1

For sha256, it's slower everywhere other than `gcc -O3 -march=native`.

sha256

There's no speed change for `crypto_aes` or `crypto_aesctr`.

... in fact, it looks like all of the cases where there's a speed change, the original code was re-using a messy buffer.  Was that deliberate, or an oversight?  If you want, I could prepare another branch which just adds a "zero out the buffer" (or fill it with `i & 0xff` as appropriate) before timing each size, so that we could truly isolate the effect of the function calls from the clean-up.

gperciva:
Oops, only `test_crc32` and `test_crypto_aesctr` uses messy buffers.  `md5`, `sha1`, and `sha256` all only check one buffer size (at the moment) and don't do any warm up.  So that's not a potential explanation for the timing differences for the latter three perftests.

tests/md5/main.c
```diff
@@ -14,39 +14,70 @@
...
 	}
-	for (i = 0; i < BLOCKLEN; i++)
-		buf[i] = (uint8_t)(i & 0xff);
 
 	/* Report what we're doing. */
 	printf("MD5 time trial.\n");
 	fflush(stdout);
 
-        /* Start timer */
+      	/* Set up. */
```
cperciva, line 68: Whitespace?

tests/crypto_aes/main.c
```diff
@@ -126,37 +126,71 @@ parse_testcase(const struct testcase testcase,
...
+	uint8_t plaintext_arr[16];
+};
+
+static int
+perftest_init(void * cookie, uint8_t * buf, size_t buflen)
+{
+	struct perftest_cookie_aes * pca = cookie;
+
+	/* Set the plaintext input. */
+	memcpy(buf, pca->plaintext_arr, buflen);
```
cperciva, line 140: I feel like this needs a "Note that `buflen` will always be equal to 16" or something like that given that `plaintext_arr` is fixed at 16 bytes.

Maybe an `assert` actually?

cperciva:
LGTM, please rebase whenever convenient.

gperciva:
Rebased, ready for merge.

PR #309, opened 2020-10-09
Perftests more buffers for md5, sha1, sha256

gperciva:


gperciva:
Adding a warm-up of 100 MB was pretty arbitrary.  Likewise, I had no particular reason for choosing 256 and 1024 as the buffer sizes to add.

Nothing jumps out at me in the plots, either.

md5
sha1
sha256

PR #310, opened 2020-10-11
Fix humansize

gperciva:


util/humansize.c
```diff
@@ -67,7 +67,13 @@ humansize_parse(const char * s, uint64_t * size)
 	int state = 0;
 	uint64_t multiplier = 1;
 
-	do {
+	/* Special case for "". */
```
gperciva, line 70: I'm not convinced that this is the best way to handle it.  I tried to work it into the main loop, but putting an `if (*s == '\0')` inside the `case 0` didn't seem any better.  Although I suppose then the loop could stay as a `do... while`.

util/humansize.c
```diff
@@ -152,7 +158,7 @@ humansize_parse(const char * s, uint64_t * size)
 
 		/* Move on to the next character. */
 		s++;
-	} while (*s != '\0');
```
cperciva, line 155: How about just changing this to `while (state != -1 && *s != '\0')` ?

gperciva:
Yes, that's much better.

PR #312, opened 2020-10-12
noeintr_close

gperciva:


gperciva:
This is a continuation of #300, but named `noeintr_close` rather than `posix_close`.

However, I think that I'm missing something fundamental about `close` and/or signals: so far I've been unable to produce any `EINTR` on FreeBSD, Linux, NetBSD, OpenBSD, or Solaris.  (8c9b721 should print a message if `close()` has any non-zero return value.)  As such, I have no evidence that this is actually working.

The idea is that `main.c` (in each test directory) spawns open_close to continually open and close a file (optionally reading 1 bytes from it, to show that it's actually opening the file).  Then `main.c` spawns send_alarm, to spam sending `SIGALRM` to (one of the) open_close threads.  In my understanding, this should interrupt the open_close.  I expect that most of the time, it would be interrupting the `open()` or `read()`, but given a hundred thousand iterations, surely at least one would happen in the `close()`?

Sample:
```
$ ./test_noeintr_close_normal /dev/null 0
Opened and closed the file 275425 times.
$ cd ../noeintr_close_pthread/
$ ./test_noeintr_close_pthread /dev/null 0
Opened and closed the file 32180 times.
Opened and closed the file 32717 times.
Opened and closed the file 32001 times.
Opened and closed the file 32213 times.
Opened and closed the file 31130 times.
Opened and closed the file 31770 times.
Opened and closed the file 31529 times.
Opened and closed the file 32238 times.
Opened and closed the file 32332 times.
Opened and closed the file 32079 times.
$ 
```

cperciva:
I believe you can trigger EINTR on FreeBSD close by using NFS interruptible mounts.  It might be necessary to write data to the file before calling close so that there's something needing to be flushed.

But for testing purposes: Create an `evil_close` function which with equal probabilities (a) closes the descriptor normally, (b) closes the descriptor and then returns EINTR, or (c) returns EINTR without closing the descriptor.  Have `posix_close` call that function if POSIX_CLOSE_TESTING is defined.

gperciva:
Updated with `evil_clone()`.  Also revised the whole `noeintr_close` test infrastructure.

gperciva:
I meant `evil_close()`, not `evil_clone()`.

gperciva:
Fixed the git conflict.

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+	if ((rc = pthread_mutex_lock(&mutex)) != 0) {
+		warn0("pthread_mutex_lock: %s", strerror(rc));
+		goto err0;
+	}
+#endif
+
+	/* Initialize (if needed). */
+	if ((!initialized) && noeintr_close_init()) {
+		warn0("noeintr_close_init");
+		goto err1;
```
gperciva, line 148: Clang `-Wthread-safety-analysis` claims that
```
../../../util/noeintr_close_shared.c:211:11: error: mutex 'mutex' is not held on every path through here [-Werror,-Wthread-safety-analysis]
        return (-1);
                 ^
../../../util/noeintr_close_shared.c:139:12: note: mutex acquired here
        if ((rc = pthread_mutex_lock(&mutex)) != 0) {
                  ^
../../../util/noeintr_close_shared.c:130:1: note: Thread warning in function 'noeintr_close'
{
```

I think this is a bogus error.  It goes away if we jump to `err1` (which would unlock the mutex), so perhaps it doesn't realize that `pthread_mutex_lock()` can fail?  I wondered if we should try unlocking the mutex in all error paths, but POSIX says it's undefined behaviour if you unlock and already-unlocked mutex.

(I've seen this error other places that we have mutexes, such as `spiped/lib/dnsthread.c`, but I didn't investigate those in detail.)

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+
+#define MAGIC_COOKIE_LEN 32
+
+static int initialized = 0;
+static int s[2] = {-1, -1};
+static uint8_t cookie_write[MAGIC_COOKIE_LEN];
+
+#ifdef NOEINTR_CLOSE_TESTING
+/*
+ * POSIX is silent about the state of fd if close(2) gets EINTR.  For internal
```
cperciva, line 35: s/gets/fails with errno set to/

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+ * testing, this function randomly selects between the return values which are
```
cperciva, line 36: s/between the return values/among the behaviours/

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+static int
+evil_close(int fd)
+{
+
+	switch (random() % 3) {
+	case 0:
+		/* Normal close(). */
+		return (close(fd));
+	case 1:
+		/* No close(), then return EINTR. */
```
cperciva, line 50: `Fail with EINTR without changing the open file description to which fd refers.`

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+		errno = EINTR;
+		return (-1);
+	case 2:
+		/* Normal close(), then return EINTR. */
```
cperciva, line 54: s/return EINTR/fail with EINTR/

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+		close(fd);
+		errno = EINTR;
+		return (-1);
+	}
+
+	/* Should never arrive here. */
```
cperciva, line 60: `/* UNREACHABLE */`

And I would probably make it `abort()`.

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	/* Set up cleanup function. */
+	if (atexit(noeintr_close_atexit)) {
+		warnp("atexit");
+		goto err1;
+	}
+
+	/* Generate a random cookie. */
+	if (entropy_read(cookie_write, MAGIC_COOKIE_LEN)) {
```
cperciva, line 94: This needs to be split into its component functions, yes?  Or were you planning on doing that when you switch `entropy_read` over to using `noeintr_close`?

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+{
+	uint8_t cookie_recv[MAGIC_COOKIE_LEN];
+	ssize_t r;
+#if NOEINTR_CLOSE_PTHREAD
+	int rc;
+#endif
+
+	/* In a multithreaded environment, lock the mutex here. */
+#if NOEINTR_CLOSE_PTHREAD
+	if ((rc = pthread_mutex_lock(&mutex)) != 0) {
```
cperciva, line 139: I would hoist these into functions `static int lock(void);` and `static int unlock(void);` which are defined in the `#if NOEINTR_CLOSE_PTHREAD` block above and in an `#else` clause to simply "return 0".  It would make the code here easier to read.

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+
+	/* Initialize (if needed). */
+	if ((!initialized) && noeintr_close_init()) {
+		warn0("noeintr_close_init");
+		goto err1;
+	}
+
+	/* Prepare a duplicate file descriptor. */
+	while (dup2(s[1], fd) == -1) {
+		if (errno != EINTR) {
```
cperciva, line 153: I would write this as

```
if (errno == EINTR) {
    /* POSIX specifies that the file descriptor is unchanged, so we can retry. */
    continue;
}

warnp("dup2");
goto err1;
```

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+			warnp("dup2");
+			goto err1;
+		}
+	}
+
+	/* Attempt to close the original descriptor. */
```
cperciva, line 159: `/* Close the descriptor, which now refers to the same socket as s[1]. */`

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+#ifdef NOEINTR_CLOSE_TESTING
+	while (evil_close(fd)) {
+#else
+	while (close(fd)) {
+#endif
+		/* Bail if there's non-EINTR error. */
```
cperciva, line 165: `/* The only way close should fail is with EINTR. */`

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+		if (errno != EINTR)
+			goto err1;
+
+		/* Try to read the cookie. */
```
cperciva, line 169: `/* Check if EINTR was returned with the descriptor still referring to to the same socket as s[1]. */`

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+		if ((r = recv(fd, &cookie_recv, MAGIC_COOKIE_LEN, MSG_PEEK))
```
cperciva, line 170: This needs to be a retry-on-EINTR loop.

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+		    == -1) {
+			/* We might have already closed fd. */
```
cperciva, line 172: `/* If the descriptor is not open to a socket, it must have been closed. */`

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+			if ((errno == EBADF) || (errno == ENOTSOCK))
+				goto done;
+
+			/* Handle non-EINTR errors. */
```
cperciva, line 176: As written, if `recv` returns -1 with EINTR we'll proceed on to comparing uninitialized memory against the cookie.

util/noeintr_close_shared.c
```diff
@@ -0,0 +1,212 @@
...
+				warnp("recv");
+				goto err1;
+			}
+		}
+
+		/* If we read the correct amount, check if it's the cookie. */
+		if ((r == MAGIC_COOKIE_LEN) &&
+		    (memcmp(cookie_write, cookie_recv, MAGIC_COOKIE_LEN)
+		    != 0)) {
+			/* No cookie; don't try to close again. */
```
cperciva, line 187: `If this descriptor doesn't have the cookie we stuffed into s[0], it no longer refers to the same file as s[1]; the close call must have succeeded.`

cperciva:
One bug (handling of recv returning EINTR), some refactoring, and lots of comment rewording.

cperciva:
Oops, didn't mean to close.

util/noeintr_close_shared.c
gperciva: Hmm, that comment changes whether you read `should` as descriptive or prescriptive.  How about
```
/* This function only tries to handle EINTR. */
```
?

cperciva: Well, I'm trying to convey the fact that since `fd` at this point is a copy of `s[1]`, it can't fail in ways (e.g. EBADF) which it could have otherwise.

gperciva:
```
/* Since fd is now a copy of s[1], the only way this close() should be able to fail is with EINTR. */
```

More verbose, but I think it's worth it.

gperciva:
There's so many commits needed that I'm not certain it's useful to keep them un-rebased.  Also, this branch would benefit from being rebased onto current master, so it gets the CI change to github actions.

Can I rebase the fixes together?

cperciva:
Yeah, go ahead and rebase everything

gperciva:
I'm not sure how to fit this in, with respect to `liball.a`.  Quick refresh:
- `noeintr_close_shared.c` has all the code.
- `noeintr_close_normal.c` does `#define NOEINTR_CLOSE_PTHREAD 0` then includes `_shared.c`.
- `noeintr_close_posix.c` does `#define NOEINTR_CLOSE_PTHREAD 1` then includes `_shared.c`.

Both object files export a `noeintr_close()` function.  Since the object files have the same name, I don't think that we can shove them both in `liball.a` and have good results.

I'm wondering if we want to move these files from `util/` to `lib-dependent/` or `pthread-dependent/` or something else -- basically, use the directory order to remind ourselves that these files are not part of `liball.a`, and need to be handled specially in the `Makefile.BSD`.

cperciva:
Ugh, I forgot about this.  Right, we can't say "make sure you link to the right version". :-(

We *could* include both versions in `liball.a` but with name mangling which would give people the right version if they include the right header... but that's not ideal either since these functions might be used in library code.

I'll have to think about this one.  TBH I was never particularly happy with the "two versions" solution... I wonder if there's a (portable) way we can detect at runtime whether we're linked to `libpthread`...?

gperciva:
Depends on how abusive you want to be.  :)

```
/* detect_pthread.c */

static int has_pthread = 1;

int get_has_pthread(void) { return (has_pthread); }

/* Copy an existing pthread function (which doens't involve pthread_t) */
void pthread_testcancel(void) {

        has_pthread = 0;
}
```

If we compile this into an object file, then append this to the **end** of command-line, then it might theoretically work.  If we had `-lpthread`, then the real `pthread_testcancel()` will be used.  If we didn't, then at the time it reaches `detect_pthread.o`, it'll still have an unresolved symbol for `pthread_testcancel()`, so the above code will be executed.

I'll try implementing this tomorrow.

cperciva:
That won't work if `libpthread` is a dynamic library (which it will be on almost every system).

gperciva:
Hmm.  It appears to work on freebsd?  I see that freebsd has both /usr/lib/libpthread.a and /usr/lib/libpthread.so.  Is it unusual to have libpthread.a?

gperciva:
(I did have to tweak the above sketch a bit, but the general idea is the same.  I'm cleaning it up now.)

PR #315, opened 2020-10-19
Simplify setuidgid

gperciva:


gperciva:
This adds to REBASE commits; one to handle with nitpicky clang warnings, the other to fix an off-by-one error in the `NUL`.

I'm not a huge fan of "allocate stuff, then check for errors afterwards", but if it makes sense to you then I'm all for accepting this instead of #306.

cperciva:
Can you rebase your REBASEs into my commit?

cperciva:
And if this was performance-critical code I would definitely go for "validate first and allocate/copy later" but since we don't care about performance here I think it's easier to reason about "if we have a group name, it should be non-empty" than "if pos < len then we should have pos < len - 1".

gperciva:
Rebased, ready for merge.

PR #317, opened 2020-10-21
events.c: limit # of events processed at once

gperciva:


gperciva:
This is a slightly modified version of commit b68021c1f9d497524c68fc6bdc5a042639e00698 from the tarsnap server source, although I ran into it with the kivaloo bench programs.

gperciva:
At the time, you noted it as being a "dirty hack".  I'm not clear on how this could be cleaner, but hopefully this will jog your memory.

cperciva:
The "right" solution for this is to use the `events_interrupt` API you added a while back.  I used this hack in the tarsnap server code because I didn't have `events_interrupt` and I needed to make sure I could handle a signal before too much time had passed.

gperciva:
hmm, ok.

Issue #320, opened 2020-10-24
Test failure with Solaris + clang + strtod + hex strings


gperciva:
The parsenum test is failing with some compilers on Solaris on the four cases with hex, which boil down to
```
strtod("0x7F", &eptr);
```

This is because Solaris' `strtod()` supports C90 by default [1], because checking for hex was a change in C99 and they didn't want to break existing code.  (The effect of this change is discussed in "Application Usage" in POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/functions/strtod.html, so I guess it's not as ridiculous as it sounds at first glance?)

[1]
> $ man 3 strtod
    In default mode for strtod(), only decimal, INF/INFINITY, and
    NAN/NAN(n-char-sequence) forms are recognized. In C99/SUSv3 mode,
    hexadecimal strings are also recognized.
    https://docs.oracle.com/cd/E88353_01/html/E37843/strtod-3c.html

`cc` and `c99` (custom compilers) are fine.  #319 enables gcc to work.  Unfortunately, the same solution doesn't work for clang:
```
td@solaris:~/src/libcperciva/POSIX$ clang -std=c99 posix-strtod.c 
td@solaris:~/src/libcperciva/POSIX$ ./a.out ; echo $?
1
```

Inspired by an old gcc bug report about Solaris and `-std=c99` [2], I can get clang to work by manually using:

```
td@solaris:~/src/libcperciva/POSIX$ clang posix-strtod.c /usr/lib/amd64/values-xpg6.o
td@solaris:~/src/libcperciva/POSIX$ ./a.out ; echo $?
0
```

However, this isn't a great solution.  (`-std=c99` seems to have no effect on this.)

[2] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=40411

gperciva:
At the moment, I plan on digging into this more.  That said, Solaris + clang 6.0 isn't a huge target; we've already had to add an awkward workaround to the Tarsnap `BUILDING` file.  It's not out of the question to give up on that combination and document that it's not working.

One potential workaround is that we could change `parsenum_float()` to use `strtold()` instead, then cast it to a `(double)`.
  > The strtof() and strtold() functions always function in  C99/SUSv3-conformant mode.
     https://docs.oracle.com/cd/E88353_01/html/E37843/strtod-3c.html

[...]

PR #321, opened 2020-10-24
Reject solaris + clang 6.0.0

gperciva:


gperciva:
tl;dr: reject clang 6.0.0 on Solaris for all Tarsnap software.

Background: https://github.com/Tarsnap/libcperciva/issues/320

Our software is aimed at C99/POSIX systems.  We can make allowances for some small violations, but there's a whole host of subtle differences between the 30-year-old C90 and the 21-year-old C99.  Trying to enumerate and handle each of those differences would be an exhausting and bug-prone task.

It would also be fairly pointless, as Solaris works with `cc`, `c99`, and `gcc`.  The problematic interaction between clang and Solaris was fixed in clang upstream, so I assume that at some point Solaris will update their `clang` package and this will all be ok.

PR #324, opened 2020-11-08
style: minor fixes

gperciva:


gperciva:
I've been combining various "check for style problems" scripts into a unified `tarsnap-lint.py` script.  This is what it came up with for libcperciva.

- As for `UNUSED`, we have 7 forms with a tab and 22 with a space, so I went with the space.

```
$ git grep -P "\\t/\* UNUSED" | wc
       7      35     345
$ git grep -P " /\* UNUSED" | wc
      22     110    1066
```

- replacing `void * cookie` with `void *` in `perftests.h` wasn't found by the unified script, but I noticed that while working on style stuff so I included it here.

PR #326, opened 2020-11-11
POSIX: struct stat should contain st_mtim

gperciva:


gperciva:
This is documented in IEEE 1003.1-2008 and 1003.1-2018:
    https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/basedefs/sys_stat.h.html
    https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html

Older versions of POSIX had a st_mtime; in Issue 7, that name should be a macro for st_mtime.tv_sec.

At least one OS decided that their `struct stat` should be compliant with 1003.1-1988 ("POSIX.1"), so this is unfortunately necessary for the tarsnap client.

gperciva:
Alternate solution: we're only using `st_mtim.tv_sec`, not `.tv_usec`, and POSIX Issue 7 specifies that `st_mtime` should be a macro which means that.  So I could just change the code in tarsnap to use `st_mtime`.  The only place it's used is in `tar/read.c`, in `check_skip_file()`:

```
        /*
         * Compare file size and mtime (seconds).  Some filesystems don't have
         * sub-second timestamp precision, so comparing the full timespecs
         * would produce a lot of false negatives.
         */
        if (file_st.st_size != archive_st->st_size)
                goto noskip;
        if (file_st.st_mtim.tv_sec != archive_st->st_mtim.tv_sec)
                goto noskip;
```

gperciva:
If we use this `POSIXFAIL`, here's it in action:
https://github.com/Tarsnap/tarsnap/pull/447

PR #327, opened 2020-11-12
Allow choice of wall-clock or cputime in perftest.h

gperciva:


gperciva:
We're only using cputime for libcperciva, but we need to use the wall-clock time for benchmarking spiped.

PR #328, opened 2020-11-24
Perftest scripts

gperciva:


gperciva:
The "column extraction" looked like it was on the verge of moving it from the Makefile into a separate shell script, but in the end I opted to put it into the Makefile.

This change was motivated by https://github.com/Tarsnap/spiped/pull/287; we only need a small change to `perftest-plot.py` to support that as well.

We'll soon move `tests/perftest*` into a separate project (tentatively named "tarsnap-tools").  The idea is that we can download that project, do a `pip-3.7 install .`, then have various useful scripts (sh and python) available.  I'm still gathering various scripts and cleaning them up, though.

tests/crc32/Makefile
```diff
@@ -41,5 +41,9 @@ warnp.o: ../../util/warnp.c ../../util/warnp.h
 test:	all
 	./test_crc32 -x
 
-perftest:	all
-	./test_crc32 -t
+perftest:
+	@${MAKE} all > /dev/null
```
gperciva, line 45: For certainty, this does not suppress output from cpusupport: that goes to stderr.  When running `perftest-generate.sh`, we do a `make all` before `perftest`, so this isn't a problem for the main "intended" way of generated perftest data".

I did consider capturing stderr as well, just to make sure that `make perftest` didn't have any undesired output, but I figured that if we had anything on stderr then it would be better to make sure it got displayed.

PR #329, opened 2020-11-25
Optimize crypto_aesctr to process blocks

gperciva:


gperciva:
Since this involves encryption, I've split these changes as granular as possible, for easy of reviewing.

Changes per commit:

aesctr-block

gperciva:
Comparing before & after on clang:

aesctr-clang

... and gcc:

aesctr-gcc

All tests on freebsd 12.1 amd64.

crypto/crypto_aesctr.c
```diff
@@ -42,6 +42,20 @@ crypto_aesctr_init(const struct crypto_aes_key * key, uint64_t nonce)
 	return (NULL);
 }
 
+/* Generate a block of cipherstream. */
+static inline void
```
cperciva, line 46: The `inline` keyword shouldn't do anything here on modern compilers.  Do you see any performance difference by removing it?

gperciva: Very slight difference in gcc 9.3.0 with `-O2` and buffer sizes of 16 and higher.  No difference in clang or gcc with `-O3 -march=native`.

- blocksize 16: slightly faster without `inline`
- blocksize 32: equal
- blocksize 64 to 1024: slower without `inline` (but still within 1%).

noinline

I thought this was weird, so I re-ran the gcc tests in the opposite order, with 7 repetitions instead of 3, and with gcc 10.2.1 instead of gcc 9.3.0.  Same results (although this time there's enough data to label a few points as outliers; those are circles).

again


In case you're wondering "why is there an 'outlier' in 1024 inline that's drawn right on top of the other data", here's the raw figures:
```
$ grep 32768 perftest-crypto_aesctr-gcc10-inline.txt
32768	1024	0.096586	347.405
32768	1024	0.096625	347.264
32768	1024	0.096625	347.264
32768	1024	0.096613	347.308
32768	1024	0.096625	347.264
32768	1024	0.096627	347.257
32768	1024	0.096624	347.268
```

(this is after tweaking the perftest system to show 3 decimal places for speed, instead of rounding to 1)

Looks very stable, other than perhaps a slight "warm up" penalty for the first one.

Anyway, it appears that `inline` does have a reproducible effect with `gcc`, but IMO it's not big enough to worry about.  I'm happy to remove it if you prefer that.

(FWIW, we also have `static inline int doevent(struct eventrec * r`, which is why I used `inline` here.)

cperciva:
A few more optimizations I'd like to see:

1. We don't need to `be64enc` the nonce value every time; if we put `pblk` into the stream structure we can just encode the nonce value once.

2. Similarly, we'll probably get a speedup by only calling `be64enc` on the counter value when `(bytectr & 0xfff) == 0` and just updating `pblk[15]` otherwise.

3. Right now the compiler can't optimize the `out = in ^ buf` loop because the buffers might overlap.  Well, actually they can overlap, but if they do then `in` and `out` have to be the same buffer.  Adding appropriate `restrict` keywords should allow the compiler to speed this up.

4. Right now we're calling crypto_aes_encrypt_block which then decides whether to call crypto_aes_encrypt_block_aesni.  I'm not sure how much the overhead is here but it might be worth making that selection (once) at the higher level rather than repeatedly indirecting.

crypto/crypto_aesctr.c
cperciva: Ok, if it helps, keep it.  I was just surprised that it did anything at all -- I thought compilers generally made their own decisions about what should be inlined, but maybe gcc is taking something from the hint.

gperciva:
1. already done in august, but I thought it was too much to include in this PR.
2. already done in august, but I thought it was too much to include in this PR.
3. will do.
4. I've wondered about this.  At first glance there's a lot of overhead, but I wondered if that might have been an artifact of the profiler.  I'll investigate more.

crypto/crypto_aesctr.c
gperciva: I agree that it's surprising that `inline` helps, and it's not something that I tested earlier.

gperciva:
I noticed a flaw in the test suite, so let's put this on hold.

gperciva:
Graphs:
### clang
clang


### gcc

gcc


Both graphs show the speed bouncing around a bit.  The drop (for most buffer sizes) comes in the "process pre-block bytes", where we do a bit more math to figure out how many (if any) bytes are left in the current cipherblock.  So that's expected; as is the fact that this minor drop is blow out of the water when we start dealing with blocks of 16.  (Of course we don't see that benefit with blocksize of 1.)

I'm surprised that we see a significant improvement with blocksize of 4 and 8.

(I disabled the "display % change if the mann-whitney-U test has a p-value of < 0.001" graphics because they obscured details, particularly in the smaller block sizes.  Suffice to say that those calculations don't reveal anything surprising; if it looks significant on the graph, the math backs it up.  I can re-create the graphs with that extra info if you want.)

I tried adding a `crypto_aesctr_stream_cipherblock_use_no_update()` which doesn't adjust `inbuf, outbuf, buflen`, as that adjustment isn't necessary if we're leaving `crypto_aesctr_stream()` early or at the very end.  I didn't see any difference in speed, so in the interest of simplicity I didn't keep that change.

gperciva:
Another potentially intesting data point: testing complier flags, again with N=11.

compilers

- for clang, `-O3 -march=native` is either neutral (bs >= 128) or slower (bs <= 64).
- for gcc, `-O3 -march=native` is either neutral (bs=8) or faster (all others).

crypto/crypto_aesctr.c
```diff
@@ -135,6 +135,24 @@ crypto_aesctr_stream(struct crypto_aesctr * stream, const uint8_t * inbuf,
...
 {
 	size_t bytemod;
+	size_t nbytes;
+
+	/* Do we have any bytes left in the current cipherblock? */
+	bytemod = stream->bytectr % 16;
+	if (bytemod != 0) {
+		/* How many bytes can we process? */
+		nbytes = 16 - bytemod;
+		if (nbytes > buflen) {
```
cperciva, line 145: This could be `>=`, and would probably speed up the small buffer case very slightly.

That said, I would probably write `if (bytemod + buflen <= 16) { ... }` and omit `nbytes` entirely.

gperciva: Yes, that works better, and avoids the pesky `size_t nbytes` that I disliked.

crypto/crypto_aesctr.c
```diff
@@ -113,22 +134,42 @@ void
...
-	size_t pos;
 	size_t bytemod;
 
-	for (pos = 0; pos < buflen; pos++) {
-		/* How far through the buffer are we? */
-		bytemod = stream->bytectr % 16;
+	/* Do we have any bytes left in the current cipherblock? */
+	bytemod = stream->bytectr % 16;
+	if (bytemod != 0) {
+		/* How many bytes can we process? */
```
cperciva, line 142: I would make this `/* Do we have enough to complete the request? */`

crypto/crypto_aesctr.c
```diff
@@ -113,22 +134,42 @@ void
...
+		if (bytemod + buflen <= 16) {
+			/* Process only buflen bytes, then quit. */
```
cperciva, line 144: s/quit/return/ ?

crypto/crypto_aesctr.c
```diff
@@ -113,22 +134,42 @@ void
...
 
-		/* Generate a block of cipherstream if needed. */
-		if (bytemod == 0)
-			crypto_aesctr_stream_cipherblock_generate(stream);
+	/* Process blocks of 16 bytes; we need a new cipherblock. */
+	while (buflen >= 16) {
+		/* Generate a block of cipherstream. */
+		crypto_aesctr_stream_cipherblock_generate(stream);
+
+		/* Encrypt the byte(s) and update the positions. */
```
cperciva, line 160: s/byte(s)/bytes/ since there's definitely 16 of them here.

cperciva:
Minor comment adjustments; please rebase.

gperciva:
Updated.  Here's the branch plots, but nothing noteworthy.  If you look at them closely, you can see a few improvements compared to the earlier timing plots.

### clang

clang

### gcc

gcc

PR #331, opened 2020-11-27
Add align_ptr.h

gperciva:


gperciva:
This functionality will be useful for future optimizations.

gperciva:
(The next time we run `make Makefiles`, it will get added as a dependency for `crypto_aes_aesni.c`.  I didn't bother including those changes in this PR, because they'll get picked up the next time we run it.)

util/align_ptr.h
```diff
@@ -0,0 +1,32 @@
...
+#define _ALIGN_PTR_H_
+
+/**
+ * ALIGN_PTR_DECL(type, name, num):
+ * Declare a variable array called ${name}, large enough to contain an array
+ * of ${num} values of ${type}.  Must not be used until ALIGN_PTR_INIT() has
+ * been called.  This macro may also create an additional variable called
+ * "${name}_buf".
+ */
+#define ALIGN_PTR_DECL(type, name, num)					\
```
gperciva, line 11: Hmm, I just realized that this won't help in `crypto_aesctr.c`.  In that case, we want a `uint8_t` array that's aligned to satisfy `__m128i`.

I'll mess around a bit more, and probably add a REBASE commit to this PR.

util/align_ptr.h
```diff
@@ -0,0 +1,41 @@
+#ifndef _ALIGN_PTR_H_
+#define _ALIGN_PTR_H_
+
```
gperciva, line 3: whoops, this should include `<stdint.h>`.

gperciva:
Aha, I forgot to add it to `tests/buildsingles/Makefile-headers`.  That shows a few problems, will be fixed soon.

gperciva:
Ok, ready for review again.

cperciva:
"aesci" in commit message

util/align_ptr.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+#define ALIGN_PTR_DECL(type, name, num, alignsize)			\
+	uint8_t name##_buf[num * sizeof(type) + (alignsize - 1)];	\
+	type * name
+
+/**
+ * ALIGN_PTR_INIT(name, alignsize):
+ * Initialize the variable called ${name} to point to an array which is
+ * aligned to ${alignsize} bytes.  They must have previously been declared
+ * with ALIGN_PTR_DECL(${name}, ${alignsize}).
```
cperciva, line 23: This deserves a comment "Requires that casting a pointer to uintptr_t behaves naturally, i.e. yields a byte offset relative to aligned memory."

The semantics required for uintptr_t are far weaker -- the values don't have to be meaningful at all, beyond returning the same pointer if you cast them back -- but every platform I'm aware of provides this (I think even including CHERI, which breaks all sorts of pointer-related assumptions).

gperciva:
Revised PR, including two comments about uintptr_t assumptions in align_ptr.h, and adding a slightly edited version of your comment about "the actual semantics required for uintpt_t" to the commit message.

PR #332, opened 2020-11-27
Better crypto_aesctr test

gperciva:


gperciva:
The old test was spending a huge amount of time on `insecure_memzero()`, because we were using `crypto_aesctr_buf()` instead of `crypto_aesctr_stream()`.  That goes back to the very first commit of this test, 6ee50068293988e6e6ab6f30cd17e648f59924fa.

This PR contains two possible fixes, both using `crypto_aesctr_stream()`:
- first commit: include `_init()` and `_free()` in the timed `perftest_func()`.
- second commit: move those functions outside of the timed portion.

I'm leaning towards including the first commit but not the second.  There isn't a significant time difference (in fact, if anything the first commit is faster than the second!), and the code is simpler.  Also, IMO it makes sense to include a little bit of timing info from `_init()` and `_free()`.

fix-tests

gperciva:
This is the flaw I noticed in https://github.com/Tarsnap/libcperciva/pull/329#issuecomment-734519657, and it invalidates the timing graphs for that PR.  I'll re-generate those plots once this (or a modified version) is merged.

gperciva:
I just remembered why I went with `_buf()` instead of `_stream()` -- that's what spiped does.

I wonder if we could entertain adding something like `crypto_aesctr_reset(uint64_t nonce)`.  At first thought, it would be
```
void
crypto_aesctr_reset(uint64_t nonce)
{

    stream->nonce = nonce;
    stream->bytectr = 0;
}
```

That would allow `proto_crypt_enc()` to continue to work, without needing to allocate and de-allocate memory all the time.

cperciva:
Hmm.  I think there's an argument to use `_stream` anyway -- spiped only tests a single size but other users of this code (e.g. tarsnap) may consume a variety of lengths.

gperciva:
Ok.  And do you have any thoughts about timing `_init()` and `_free()` 1 time, or 0 times?  I can imagine arguments either way.

I'm leaning towards "0 times", since with the new `_alloc` / `_init2`, there's no need to have the object creation as part of a time-critical section.

gperciva:
I went ahead and combined the commits, so we don't time `_init` and `_free` at all.  I still have the old branch in case you prefer it the other way.

I updated the timing plot, which also includes new % change text.  We see that with a buffer size of 1024, this change actually makes the test slower (-0.3%).  I'm guessing that this is because the old code had a `struct crypto_aesctr` on the stack (as part of `crypto_aesctr_buf`), whereas the new code allocates that on the heap, and this can influence whether or not the `buf` array is aligned on a 16-byte boundary.

I have other branches that deal with such alignment, but I'd like to merge this PR (and a few other improvements, such as #329) before dealing with them.

gperciva:
On second thought, this should probably wait until we #333, so we can use `_alloc()` and `_init2`.

PR #333, opened 2020-11-27
Add crypto_aesctr_alloc and _init2

gperciva:


gperciva:
It doesn't feel like "reset" is quite the right term.  I considered calling it `crypto_aesctr_reset_with_nonce` instead, but that feels too wordy.

I'm hoping that we can eventually use this for spiped's `proto_crypt_enc()`, once it's made its way through various test programs.

cperciva:
I'm inclined to call this `crypto_aesctr_init2` and also add a new `crypto_aesctr_alloc` function, then document that `crypto_aesctr_init` is the same as calling `_alloc` followed by `_init2`.

gperciva:
It feels a bit weird to pass the `key` to an `alloc` function, but in terms of API design we don't want to require the callee to keep copies of the key to pass it whenever we want to call `init2` again, so ok.

cperciva:
Hmm, I forgot about the `key` parameter.  Ick, you're right that it doesn't really belong in `alloc`...

cperciva:
Here's an idea, how about `init2` takes a `key` parameter, but it can be NULL meaning "keep using the same key"?

gperciva:
I'm on board with that.  I added some REBASE commits so that we can see how it looks in `tests/crpto_aesctr/main.c`; particularly with `perftest_init` and `perftest_func`.

crypto/crypto_aesctr.c
```diff
@@ -40,14 +37,17 @@ crypto_aesctr_alloc(const struct crypto_aes_key * key)
...
+ * is NULL, retain the previous AES key.
  */
 void
-crypto_aesctr_init2(struct crypto_aesctr * stream, uint64_t nonce)
+crypto_aesctr_init2(struct crypto_aesctr * stream,
+    const struct crypto_aes_key * key, uint64_t nonce)
 {
 
+	if (key != NULL)
+		stream->key = key;
```
cperciva, line 50: Maybe add an `assert(stream->key != null)` here?

cperciva:
Looks good, please rebase.

gperciva:
Rebased.

This PR includes using `_init2` for the perftest; as you'd expect, they complete much faster.  This PR therefore replaces #332.

Figure_1

gperciva:
I just updated this (via rebase) to use `crypto_aesctr_init2()` in `crypto_aesctr_buf()`.  This is a new change from the previous version, but I think it's worth it.

gperciva:
Due to gcc being fussy about a potentially NULL pointer, I've reworked this to add sanity checks first.

gperciva:
The reworked code has very slightly different times:

- bs 128: 3.3% to 3.0%
- bs 64: 7.5% to 5.7%
- Other changes are within 0.2%.

If you're curious, I recommend opening both png files in a new browser window, then alt-tabbing between the images.  I don't think this is important by itself; I'm just interested in figuring out a baseline for how much the perftest times change based on recompiling with (virtually) insignificant code changes.

Figure_1

gperciva:
Ignore my "it has different times" comment.  Further testing reveals that such a variation is completely normal, even with exactly the same binaries.

I'll play around with a few options that may provide more stable benchmarks, and/or incorporate more robust statics (potentially involving `ministat(1)`).

PR #334, opened 2020-11-29
Comment: clarify that ${hash}_Final clears state

gperciva:


alg/sha1.c
```diff
@@ -391,6 +402,18 @@ HMAC_SHA1_Final(uint8_t digest[20], HMAC_SHA1_CTX * ctx)
 	insecure_memzero(ihash, 20);
```
gperciva, line 402: `HMAC_SHA1_Final` implicitly clears its context by calling `SHA1_Final` on `->ictx` and `->octx`.  It might be appropriate for it to call the new `HMAC_SHA1_Clear` as well, just for extra clarity.

gperciva:
I'm confident that we want the first commit, but only 75% about the second one.  It's purely an "abstraction convenience" for the error path in code that uses these hashes: the context is cleared when we call _Final(), but it seems weird to call _Final() in an error path.  Another option would be to call

    insecure_memzero(ctx, sizeof(HASH_CTX));

in the callee code, but that seems like a poor abstraction.  However, all these hashes define their context in the header specifically so that they can be instantiated as local variables, so in that sense it wouldn't be weird for callee code to do the `insecure_memzero` as well.

cperciva:
I'm not really a fan of `_Clear`.  The MD5/SHA1/SHA256 functions use a very standard API; I don't really want to offer a mostly-the-same-but-slightly-different API here since someone might want to take code which was written with libcperciva's hash functions in mind and drop in their own hash routines (e.g. optimized with CPU intrinsics).  Adding a new function, even one as simple as state-wiping, would make that annoying.

gperciva:
Ok.  I removed the `_Clear` and slightly reworded the comment-only commit subject line.

PR #337, opened 2020-12-04
Valgrind on freebsd

gperciva:


gperciva:
This allows us to check for memory leaks on FreeBSD with the new valgrind-devel port:
https://www.freshports.org/devel/valgrind-devel

It's possible that a later version of that package/port will automatically handle some of those suppressions, but they're necessary as of valgrind-devel 3.17.0.g20200723,1; and there's no harm in having extra checks in `potential-memleaks.c` that don't demonstrate any leak.

gperciva:
Ready for review (and likely merge, since it's only changing the valgrind test).

PR #338, opened 2020-12-05
Prep for aesctr block optimization

gperciva:


gperciva:
This branch was just supposed to get a pair of trivial commits out of the way, so that #329 could focus on the more important parts of optimizing AES-CT with blocks.  But the first commit (changing `int bytemod` to `uint64_t bytemod`) has a surprising speed increase, so I decided to dig in.

First, the effects of each commit in the branch (including `master`), compiled with `-O2`, running the binary 11 times, and printing the % change if the mann-whitney-U test has a p-value of < 0.001:

### clang 11.0.0
clang

### gcc 9.3.0
gcc

Looking at the effect of the branch as a whole with various `CFLAGS`, only running 3 times:

compilers

So there's definitely an effect happening.  I examined the difference between `master` and the first commit with `objdump -S crypto_aesctr.o | cut -f 2- > filename.txt` (cutting out the hex positions, because that added a lot of fluff to the diff), and saw for both gcc and clang (compiled with `-O2`), we had essentially

clang:
```
-movl	%eax, %r14d
-andl	$15, %r14d
+movq	%rax, %r14
+andq	$15, %r14
```

gcc:
```
-movslq	%r15d, %rcx
...
-movl	%eax, %r15d
-andl	$15, %r15d
+movq	%rax, %r14
+andl	$15, %r14d
```

along with adjusting some `nopl` / `nopw` / `nop` commands at the end of a loop, presumably for alignment.

I'm surprised that there's such a difference from just those commands... and also intrigued that gcc isolates bytes 0-3 (with `%r14d`) whereas clang uses the 8-byte register (with `%r14`).  I'm guessing that this gets into pipelines or memory access patterns or something like that?

Anyway, my curiosity is satisfied with seeing that there's no outrageous change to the assembly, and that this is reproducible on amd64 and arm.

gperciva:
Oops, forgot to mention that in the first two graphs, I plotted slow outliers very faintly (alpha=0.1).  That's because I haven't done any special config for benchmarking ([1] suggests running in single user mode, mounting filesystems as read-only, etc.)   So if 10 iterations of the test taks 0.116 or 0.117 seconds, and 1 time it takes 0.124 seconds, I'm not too fussed.

I'm still considering how best to display such outliers.

[1] https://wiki.freebsd.org/BenchmarkAdvice

cperciva:
> [1] https://wiki.freebsd.org/BenchmarkAdvice

I don't worry too much about such things, especially for single-threaded CPU-bound tasks.  You could be browsing reddit or facebook and there's almost certainly going to be a CPU available nearly all of the time.

The benchmarks where you need to be more careful are things involving disk I/O and sometimes kernel-intensive workloads (although with fine-grained kernel locking this is less of an issue than in past decades).

cperciva:
Do you get the same speedup if you make it `size_t` instead?  I hesitate to ask for an explicit `uint64_t` unnecessarily since on 32-bit systems those are slow -- but 32-bit systems will have a 32-bit `size_t`.

gperciva:
> You could be browsing reddit or facebook

I tend to check hn for tech articles... but that's using my desktop pi, not the freebsd server that's running the benchmarks.  I quite like my "semi-thin client" setup.  :)

Yes, it's exactly the same speedup with `size_t`.  That's what I had originally, but then I noticed that `bytectr` is a `uint64_t`, so I figured that it might be helpful to use the same data type.  Happy to change it back to `size_t` if you don't think this is a concern.

gperciva:
Correction: we still have the same speed increase with gcc, but not with clang.  Details coming in a few minutes.

cperciva:
On 64-bit systems, you'll usually find

```
typedef       __uint64_t      __size_t;
typedef       __uint64_t      uint64_t;
typedef        __size_t        size_t;
```

so generally speaking it will be the same type anyway...

gperciva:
Aha, solved it.  I knew that cc was clang, so I was lazily testing
```
perftest-branch.py
```
But of course cc is clang-8.0.1 (on FreeBSD 12.1), not clang-11 (which is what I was testing earlier).

When I do the proper (on my system)
```
CC=clang perftest-branch.py
```

I see the earlier results.

I've rebased this branch to use `size_t`; potentially ready for merge after the tests finish.

PR #339, opened 2020-12-06
Aes cleanup

gperciva:


gperciva:
This is "harmless prep" for upcoming optimizations to AES that I'll benchmark and upload after updating #329.  (I only mention this in case you're wondering about the motivation for this PR.)

crypto/crypto_aes.c
```diff
@@ -53,9 +53,11 @@ aesnitest(uint8_t ptext[16], uint8_t * key, size_t len)
 	return (-1);
 }
 
-/* Should we use AESNI? */
+/** crypto_aes_useaesni(void):
+ * Return non-zero if AESNI operations are available.
+ */
 static int
-useaesni(void)
+crypto_aes_useaesni(void)
```
cperciva, line 60: This does not match what the commit message says.

gperciva:
Rebased to fix the commit message, and also fix the comment formatting which was
```
/** crypto_aes_useaesni(void):
```
instead of the function name being on a new line.

gperciva:
Whoops, I just noticed that this `useaesni()` won't be defined unless we have `CPUSUPPORT_X86_AESNI`.  I'll revise this.

cperciva:
> rename check to crypto_aes_aesni()

s/aesni/useaesni/ ?

gperciva:
Right, will do.

This PR is kind-of waiting for the follow-up where we use it in `tests/crypto_aes/main.c`, but I ran into some nitpicky problems with that one, and now I'm focusing on Christmas stuff.  I'm aiming to do another chunk of work in the morning.

cperciva:
I'm fine with waiting, or with pulling it as soon as you fix the commit message; either works for me.

PR #340, opened 2020-12-06
Add crypto_aes_encrypt_block_aesni_m128i()

gperciva:


gperciva:
Includes #339.  TL;DR: adds
```
__m128i crypto_aes_encrypt_block_aesni_direct(const __m128i, const void *);
```
to `crypto_aes_aesni.h` for an ~8% speed boost to `test_crypto_aes`.

clang

This is absolutely intended for `crypto_aesctr.c`, not for non-libcperciva code.  Since AES involves 16 8-bit values, shoving all that into a `__m128i` seems very enticing to me.

crypto/crypto_aes_aesni.h
```diff
@@ -23,6 +25,15 @@ void * crypto_aes_key_expand_aesni(const uint8_t *, size_t);
 void crypto_aes_encrypt_block_aesni(const uint8_t[16], uint8_t[16],
     const void *);
 
+/**
```
gperciva, line 28: I just remembered that this needs to be inside an `#ifdef CPUSUPPORT_X86_AESNI`.

crypto/crypto_aes_aesni.h
```diff
@@ -1,6 +1,8 @@
 #ifndef _CRYPTO_AES_AESNI_H_
 #define _CRYPTO_AES_AESNI_H_
 
+#include <emmintrin.h>
```
gperciva, line 4: Technically, `<emmintrin.h>` is all we need for `__m128i`, but since this is already inside an `_aesni.h` file, maybe it makes more sense to use `<wmmintrin.h>`?

And we also need `"cpusupport-config.h"` here.

crypto/crypto_aes_aesni.h
```diff
@@ -23,6 +29,17 @@ void * crypto_aes_key_expand_aesni(const uint8_t *, size_t);
 void crypto_aes_encrypt_block_aesni(const uint8_t[16], uint8_t[16],
     const void *);
 
+#ifdef CPUSUPPORT_X86_SSE2
```
gperciva, line 32: What do we want here?
- we only need SSE2 to use `__m128i`
- OTOH, there's no point defining this unless we have AESNI as well

I've gone back and forth a few times.  I currently have _SSE2, but I'm not very firm about that.  This is the first time we're adding cpu-specific functions to header files, so we don't have a precedent to look at.

crypto/crypto_aes_aesni.c
```diff
@@ -177,22 +177,19 @@ crypto_aes_key_expand_aesni(const uint8_t * key, size_t len)
 }
 
 /**
- * crypto_aes_encrypt_block_aesni(in, out, key):
- * Using the expanded AES key ${key}, encrypt the block ${in} and write the
- * resulting ciphertext to ${out}.  ${in} and ${out} can overlap.  This
- * implementation uses x86 AESNI instructions, and should only be used if
- * CPUSUPPORT_X86_AESNI is defined and cpusupport_x86_aesni() returns nonzero.
+ * crypto_aes_encrypt_block_aesni_direct(in, key):
```
gperciva, line 180: I'm not sold on the name `_direct()`.  I considered `_sse2()` but that looked worse.

gperciva:
I just noticed that the function names in `crypto_aes_aesni.h` have the "aesni" at the end of the function names, e.g., `crypto_aes_encrypt_block_aesni()`.  I assume that's for similarity to `crypto_aes_encrypt_block()`?

My first instinct (which I followed in the upcoming `crypto_aesctr_aesni.h`) was to put the "aesni" where it comes in the filename, e.g., `crypto_aesctr_aesni_stream_cipherblock_generate()` rather than `crypto_aesctr_stream_cipherblock_generate_aesni()`.  If you're certain that you want to keep the `_aesni` at the end, I'll go rename all the new function names.

gperciva:
Oh wait, I lost my train of thought.  I originally wanted to ask: should I follow the existing pattern and make the new function be
```
crypto_aes_encrypt_block_direct_aesni()
```
or should I keep it as:
```
crypto_aes_encrypt_block_aesni_direct()
```
?

tests/crypto_aes/main.c
```diff
@@ -225,13 +258,18 @@ selftest(void)
 	size_t i;
 	size_t failures = 0;
 	size_t num_tests = sizeof(tests) / sizeof(tests[0]);
+#ifdef CPUSUPPORT_X86_AESNI
+	__m128i bufsse;
```
gperciva, line 262: I should use `ALIGN_PTR_DECL` for this.

crypto/crypto_aes_aesni.c
cperciva: I would be inclined to go with `crypto_aes_encrypt_block_aesni_m128i`.

gperciva: Thanks.  I'm reconsidering the whole approach to crypto_aesctr, though, so it's probably best to hold off on reviewing them more.  It looks like we might be able to get 90% of the speed boost with a mess less invasive code change.  If that works out, then I'll submit that as a separate PR, then look at the fancier stuff later (if it even makes sense).

cperciva: Ok.  FWIW, I'm inclined to put this into a different header file which only gets included if CPUSUPPORT_X86_AESNI rather than having it internally decide whether to include the header needed for __m128i.

Also, I would seriously consider having the "aesctr aesni" special case only used for encrypting complete blocks at a time.  For the pre-block and post-block encryption we can take the regular slower path...

gperciva: That's tempting, but unfortunately having any function call inside the `_stream` function immediately drops performance by ~33% (397 to 308 MB/s for buffer size of 1024).  Using SSE2 instructions to xor the buffers creeps us back up to 410 MB/s so technically it's still an overall gain, but nothing near as nice as I was hoping for.  I tested the performance penalty by adding a `_wholeblock()` function to `crypto_aesctr.c` itself, then adding the non-standard `__attribute__((noinline))` to it.

At the moment, I'm leaning towards having some unfortunate duplication of code between `crypto_aesctr.c` and `crypto_aesctr_aesni.c`.  I like the idea of a `_wholeblock` function in `crypto_aesctr_aesni.c`, becaues that will let us use a `crypto_aes_encrypt_block_aesni_m128i()` with a temporary `__m128i` variable, without worrying about memory alignment of `stream->buf`.

(I expect that it will still be useful to do some memory alignment of `stream`, but it would be cleaner to handle that as a separate question from the mere existance of `crypto_aesctr_aesni.c` itself.)

gperciva:
Ok, I've sorted out the new plan for aesctr, and how this will fit in.  I've updated this PR.
- I removed any attempt to store `__m128i` inside a structure; there's no need for that, and it massively complicates things w.r.t alignment.
- Irenamed the function to end `_m128i` as requested.
- I don't think that I understood your comment
  > FWIW, I'm inclined to put this into a different header file which only gets included if CPUSUPPORT_X86_AESNI rather than having it internally decide whether to include the header needed for __m128i.

  so I haven't done anything about that.  In the current version of this PR, it's being added to `crypto_aes_aesni.h`.  If you want it in a separate header file, we would... add a `crypto_aes_aesni_aesni.h` [sic] ?

cperciva:
I was thinking `crypto_aes_aesni_m128i.h`, but yes.  There's nothing wrong with having multiple header files... in many ways, better to have multiple headers than have headers which behave in different ways based on compiler defines.

gperciva:
Revised, ready for review.

PR #341, opened 2020-12-07
Optimize aesctr generate

gperciva:


crypto/crypto_aesctr.c
```diff
@@ -95,7 +101,11 @@ crypto_aesctr_stream_cipherblock_generate(struct crypto_aesctr * stream)
 {
 
 	/* Prepare counter. */
-	be64enc(stream->pblk + 8, stream->bytectr / 16);
+	stream->pblk[15]++;
```
gperciva, line 104: I did try putting this inside an `if ... else` construct.  For some buffer sizes, it had no effect; for others, it degraded performance slightly.  I find the current construct slightly easier to understand, but I have no qualms about switching to the other one if you prefer that.

gperciva:
7% improvement with large buffer sizes.  I omitted the "print statitically significant % changes" because that cluttered up the plots too much.

### clang

Improvement in all cases with clang.

clang

### gcc

Improvements in most places with gcc, apart from a buffer size of 1.  The speed has a very slight drop (~139.6 MB/s to 139.4 MB/s), but valgrind's callgrind reports that the newer code has ~9 million fewer instructions being executed.  I can't think of why these changes might result in a different type of instructions, so I'm tentatively suspecting some sort of cache or memory access thing.

gcc

### CC and CFLAGS

Interestingly, both compilers' `-O3 --march=native` is terrible with buffer sizes of 1, 2, and 4.  After that, gcc shows improvement with the extra optimization, while it doesn't help clang.

compilers

cperciva:
s/IV/Counter Block/.  The term "Initialization Vector" should never be used in the context of CTR mode.

I'm not completely happy with crypto_aesctr_stream_cipherblock_generate assuming that bytectr is increasing by 16 every time it gets called, but I don't see any particularly good way around that.

gperciva:
Well, there's no noticeable penalty to slapping a
```
assert(stream->bytectr % 16 == 0);
```
to the top of `_generate()`; would that make you feel better?

(I checked that the `assert` wasn't getting compiled out; if I change it to `==1` then I hit the abort trap.)

gperciva:
(of course such an `assert` doesn't help if somebody calls `_generate()` twice in a row, but at least we know they're not calling it after only 1-15 bytes?)

Rebased to remove references to IV from the commit messages, and to include the sanity check to the first commit.  No apparent change to the timing plots, so I haven't bothered to upload new ones.

gperciva:
Hmm, I discovered a bug in the perftest generation; `cpusupport-config.h` might not have been correct.  I'll regenerate these graphs shortly (later tonight).

gperciva:
Updated the plots in the earlier comment; no significant changes compared to the earlier version.

crypto/crypto_aesctr.c
```diff
@@ -93,14 +99,19 @@ crypto_aesctr_init(const struct crypto_aes_key * key, uint64_t nonce)
 static inline void
 crypto_aesctr_stream_cipherblock_generate(struct crypto_aesctr * stream)
 {
-	uint8_t pblk[16];
 
-	/* Prepare nonce and counter. */
-	be64enc(pblk, stream->nonce);
-	be64enc(pblk + 8, stream->bytectr / 16);
+	/* Sanity check. */
+	assert (stream->bytectr % 16 == 0);
```
cperciva, line 104: s/assert (/assert(/

(It's a function, not a keyword.)

crypto/crypto_aesctr.c
```diff
@@ -52,9 +52,15 @@ crypto_aesctr_init2(struct crypto_aesctr * stream,
 		stream->key = key;
 
 	/* Set nonce as provided and reset bytectr. */
-	stream->nonce = nonce;
+	be64enc(stream->pblk, nonce);
 	stream->bytectr = 0;
 
+	/*
+	 * Set the counter such that the smallest byte will overflow once
```
cperciva, line 59: s/smallest/least significant/

s/overflow/wrap/

crypto/crypto_aesctr.c
```diff
@@ -93,14 +99,19 @@ crypto_aesctr_init(const struct crypto_aes_key * key, uint64_t nonce)
...
-	/* Prepare nonce and counter. */
-	be64enc(pblk, stream->nonce);
-	be64enc(pblk + 8, stream->bytectr / 16);
+	/* Sanity check. */
+	assert (stream->bytectr % 16 == 0);
+
+	/* Prepare counter. */
+	stream->pblk[15]++;
+	if (stream->pblk[15] == 0) {
+		/* If there was an overflow, write the full value. */
```
cperciva, line 109: "If incrementing the least significant byte results in it wrapping, re-encode the complete 64-bit value."

cperciva:
Style and comment nits (most importantly, "overflow" is something which only happens to signed values; unsigned values "wrap"); otherwise good to go.  Please fix and rebase.

gperciva:
Rebased as requested, although I used the past tense in "byte resulted in it wrapping" instead of the present tense "byte results in it wrapping".

PR #342, opened 2020-12-09
cpusupport: add HWCAP to compile checks

gperciva:


gperciva:
This is required for runtime feature detection on ARM.

Unfortunately there is no cross-platform method of detection CPU
capabilities in ARM (possibly due to its history as mostly being for
embedded processors?).  The official ARM C Language Extensions states:
>    In AArch64, the ABI does not define a standard set of build
>    attributes and takes the approach that binaries are, in general,
>    not portable across platforms.
>    https://developer.arm.com/documentation/101028/0012/2--Introduction?lang=en

That said, the Linux kernel documentation for AArch64 (written by
somebody with an @arm.com email address, so presumably very familiar
with the chip) states:
>    The ARM architecture defines a set of feature registers, which
>    describe the capabilities of the CPU/system. Access to these system
>    registers is restricted from EL0 and there is no reliable way for an
>    application to extract this information to make better decisions at
>    runtime.
>    https://www.kernel.org/doc/html/latest/arm64/cpu-feature-registers.html

The kernel can provide some info via the ELF auxiliary vector; in
particular with the AT_HWCAP option.  Unfortunately this is an
OS-specific function:

- Linux glibc uses getauxval(), added in 2012.
  https://lwn.net/Articles/519085/

- FreeBSD made elf_aux_info() a public libc function in 2017.
  freebsd/freebsd@990c7fb

gperciva:
A PR of this in action (adding CRC32C support for ARM on Linux) coming soon.

At the moment I have no upcoming PRs for FreeBSD support, so `cpusupport-HWCAP-ELF_AUX_INFO.c` is a purely theoretical (/ useless) file.  I'm happy to remove that for now -- I anticipate that I'll use it shortly after freebsd 13.0 is available for t4g.  (although come to think of it, I could try it on freebsd 12.2 anyway)

cperciva:
I thought you were going to put these into a separate "platform support" section?  I'm fine with having them here though.

I'd rather not have unused bits in the tree, so let's add Linux bits first and FreeBSD later once you've had a chance to hook it all up.

gperciva:
Oh, the apisupport stuff?  Hmm.  I was thinking that since hwcap is necessary for cpusupport (on arm), it made more sense there.  I guess that the dividing line in my mind is "how will this change the code?".  In this case, the goal is to use (or not) cpu intrinsics.  By contrast, something like `getrandom()` or `getentropy()` is about API usage.

cperciva:
Yeah, apisupport is what I was thinking of.  Ok, let's keep it here for now at least; we can always move it later if it starts feeling out of place.

cperciva:
Is HWCAP a general mechanism or specific to ARM?  My intention was that when we say e.g. "X86 AESNI" we mean "AESNI on the X86 platform" so "HWCAP GETAUXVAL" feels strange to me -- I would have thought "ARM HWCAP_GETAUXVAL" instead?

gperciva:
Yes and no.  It's general, but not used on x86?

`getauxval(3)` (and `elf_aux_info(3)`) is available on every platform, but calling either one on x86 just gives a 2.

On linux, it looks like it gives real info on arm (32-bit), aarch64, mips, nds32 (whatever that is), s390, and sparc (32 and 64).  That's... 6/25 platforms (assuming that each subdirectory contains exactly 1 platform).  I don't know if they're planning to use `getauxval(AT_HWCAP)` for other platforms.

Jumping back to FreeBSD, looks like it's used for arm, aarch64, and powerpc.

Generally speaking, the articles, man-pages, and kernel source code comments I've skimmed all agree that this is intended to be a general method of getting cpu features.  I suspect that the reason it doesn't give useful info on x86 is just that everybody has been using cpuid for a decade so it's not a priority to change.

gperciva:
Oh, and grammatically, I was intending it to be "get HWCAP info via `getauxval()`" and "get HWCAP info via `elf_aux_info()`".  Mostly so that the two files with HWCAP would appear next to each other when sorted alphabetically.  There's nothing gramatically wrong with "use `getauxval()` to get HWCAP info", but that would ruin the sorting.

PR #343, opened 2020-12-11
Remove AES-192 remnant, improve sha256 timing, reformat comment

gperciva:


gperciva:
These are some odd commits which didn't really fit into other PRs, but would be nice to get in.

The change to `crypto_aes` was previously in #339, but adding it here allows that other PR to be focused on the `crypto_aes_useaesni()` change.

The sha256 test change has a surprisingly large benefit:

(clang)
Figure_1

Evidently the unnecessary code was taking a large portion of the test time.

PR #344, opened 2020-12-12
tests/perftest-*: remove

gperciva:


gperciva:
These scripts are now in the tarsnap-tools repository, and those versions have significant bug fixes and added functionality.

PR #345, opened 2020-12-13
cpusupport: HWACCEL macro

gperciva:


alg/sha256.c
```diff
@@ -80,7 +84,7 @@ static const uint32_t initial_state[8] = {
 	0x510E527F, 0x9B05688C, 0x1F83D9AB, 0x5BE0CD19
 };
 
-#if defined(CPUSUPPORT_X86_SHANI) && defined(CPUSUPPORT_X86_SSSE3)
+#ifdef SELFTEST
 /*
  * Test whether software and SHANI transform code produce the same results.
```
cperciva, line 89: Did you want to rename "shanitest" to "hwtest" or something similar while you're "genericizing" this?

cperciva:
I'm not entirely sure why we want SELFTEST rather than simply writing `#if defined(X86_FOOBAR) || defined(ARM64_BARFOO) || etc` in the appropriate place.

I'm not objecting, but it's not clear to me what the motivation is here.

cperciva:
Also, "sha2356" in the commit message.

gperciva:
I thought it would look weird to have:

```
#if defined(X86_)
...
...
#if defined(X86_)
    stuff_x86();
#endif /* X86_ */
...
...
#endif /* X86_ */
```

While I was looking at the earlier code and trying to anticipate your review comments (so that I could fix it in advance), I figured that changing the outer `#if` to `SELFTEST` would 1) avoid it looking so weird, and 2) clarify the code a bit.

Another option would be change the final comment to `#endif /* X86_ (outer)`.

It's true that for CRC32, we'll probably be adding ARM_CRC32 soon.  But I anticipate it taking a bit longer for AES and SHA2, and in the mean-time we'd have the weird "double #if" thing.

Another option would be to avoid changing each "generic algorithm" file until just before adding the ARM instrinsics.  I'm not as fond of that option, since it means that (for example) the PR that adds ARM_CRC32 would also be changing the "should we use X86 code" lines.

That said, I don't see a huge benefit (or penalty) for any of these options.  I do think that adding the SELFTEST macro is the most clear way to organize the source code and git history, but if another option makes more sense to you the I'm happy to do that instead.

gperciva:
There's a relation question: how can / should we expose "can we use intrinsics" functions?  For example, `crypto_aes_useaesni()` would be invaluable for an upcoming `crypto_aesctr_aesni.c`. 
 (https://github.com/Tarsnap/libcperciva/pull/339)

I'm treating that as a slightly different problem from this; maybe doing something like:
```
int
crypto_aes_useaseni()
{

#ifdef SELFTEST
     return aestest();
#else
    return (0);   
#endif
}
```

Of course, if we didn't have SELFTEST then we could have a bunch of `defined(X86_)` and `defined(ARM_)` in there.  But that strikes me as a bit messier.

gperciva:
BTW, it's possible that within a few weeks, `alg/sha256.c` might contain:
```
#if defined(CPUSUPPORT_X86_SHANI) && defined(CPUSUPPORT_X86_SSSE3)     \
    || defined(CPUSUPPORT_X86_SSE4)        \
    || defined(CPUSUPPORT_ARM_SHANI)
#define SELFTEST
#endif
```

I think it would look a lot neater if we could use `#if SELFTEST` throughout the file, instead of repeating the multiple-condition `#if`.

cperciva:
What's the connection to SSE4?

gperciva:
A possible `alg/sha256_sse4.c`.

Looks nice in clang, but gcc is being a pain.  Still a work in progress.

cperciva:
I think you've convinced me that it makes sense to have a macro, but I'm not sure that "SELFTEST" is the right name for it.  Maybe "HWACCEL"?

I mean, yes, the only thing which is exposed by this macro is the self-test functionality, but it feels more natural to say "these are types of hardware acceleration" than "these are things which need a self test".

cperciva:
> A possible `alg/sha256_sse4.c`.

Let me rephrase that: What does SSE4 have which is useful?

alg/sha256.c
gperciva: Nah, not really.  I could do it if you think it's worth it.

cperciva: Ok, just checking.  We can always change it when we add non-x86 bits.

gperciva:
gcc might be happier with `_mm_extract_epi32`.  It's certainly not happy with `_mm_loadu_si128`.

I'm also not ruling out SSSE3-only, or AVX-only.  When I have various options and benchmarks, I'll let you pick which version (if any) is worth using.

gperciva:
I can go with HWACCEL.

cperciva:
Sounds good.

alg/sha256.c
gperciva: ah, right.  I only glanced at `crc32.c`, for which `crctest()` would still be a valid name.  But you're right that `shanitest()` wouldn't be appropriate for ARM SHA code.

Ok, I'll change them all to hwtest.

gperciva:
Updated with `hwtest()` and potentially ready for merge.

PR #346, opened 2020-12-13
cpusupport crc32c ARM

gperciva:


gperciva:
I worked on CRC32 for the first ARM instrinsics because I figured it would be easier than AES or SHA2.  As you'd expect, using the instrincs blows software CRC32 out of the water, being up to 13 times faster.

### t4g (graviton2)

- clang can't handle `-march=native` on this chip (as of clang-11.0.0).

t4g-crc32

### pi4

- I used the latest versions of gcc and clang that were in raspbian, but that's only gcc-8 and clang-9.
- it's interesting that `gcc -O2` did better than `-O3 -march=native`, but given the relatively old versions, I don't think that investigating in detail would be of much use.

pi4-crc32

cperciva:
"sha2356"?

cpusupport/Build/cpusupport-ARM-CRC32_64.c
```diff
@@ -0,0 +1,22 @@
...
+	uint8_t buf[8] = {0, 0, 0, 0, 0, 0, 0, 0};
+
+	/* Test 8-bit and 64-bit. */
+	state = __crc32b(state, x);
+	state = __crc32cd(state, *(const uint64_t *)&buf[0]);
+
+	return (0);
+}
+
+#endif /* !__ARM_ACLE */
```
cperciva, line 22: Shouldn't this be at the top?  I mean, it makes more sense to have this file fail to compile because `__crc32b` isn't defined on a compiler which is unaware of the extensions than than to fail to link because `main` isn't defined.

cpusupport/cpusupport_arm_crc32_64.c
```diff
@@ -0,0 +1,23 @@
...
+
+#if defined(__aarch64__)
+	capabilities = getauxval(AT_HWCAP);
+	supported = capabilities & HWCAP_CRC32;
+#endif
+#endif /* CPUSUPPORT_HWCAP_GETAUXVAL */
+#endif /* CPUSUPPORT_ARM_CRC32_64 */
+
+	/* Return the supported status. */
+	return (supported);
```
cperciva, line 36: Didn't we decide that these functions should return 1 or 0 rather than a particular bit out of a mask?

gperciva: Right, fixed with `(supported > 0)`.  I considered using `(supported ? 1 : 0)` which is the closest to what all the other "runtime detection" files do, but that seemed unnecessarily complicated.

gperciva:
updated with REBASE commits.

cpusupport/cpusupport_arm_crc32_64.c
```diff
@@ -0,0 +1,37 @@
...
+	supported = capabilities & HWCAP_CRC32;
+#elif defined(__arm__)
+	capabilities = getauxval(AT_HWCAP2);
+	supported = capabilities & HWCAP2_CRC32;
+#endif
+#endif /* CPUSUPPORT_HWCAP_GETAUXVAL */
+#endif /* CPUSUPPORT_ARM_CRC32_64 */
+
+	/* Return the supported status. */
+	return (supported > 0);
```
cperciva, line 36: Hmm, now that I'm looking at the other cpusupport files -- why not use the same layout, i.e.

```
#if defined(CPUSUPPORT_ARM_CRC32_64)
#if defined(CPUSUPPORT_HWCAP_GETAUXVAL)
unsigned long capabilities;

#if defined(__aarch64__)
capabilities = getauxval(AT_HWCAP);
return ((capabilities & HWCAP_CRC32) ? 1 : 0);
#elif defined(__arm__)
capabilities = getauxval(AT_HWCAP2);
return ((capabilities & HWCAP2_CRC32) ? 1 : );
#endif
#endif
#endif

return (0);
```

cperciva: In any case, I like the consistency of having these all `return ((foo) ? 1 : 0)` since that makes it immediately clear that the value is either 1 or 0.

gperciva: Would you be ok with
```
supported = (capabilities & HWCAP_CRC32) ? 1 : 0;
```

but retaining the single return statement?  (currently uploaded version)

The reason is pure banality: avoiding `clang -Weverything` complaining about unreachable code.  It would be possible to avoid this with various `#else \n return (0)` constructs, but that makes the code less readable.

The reason this isn't a problem for the other cpusupport files is that they have a single detection method (cpuid) which is valid for all relevant platforms.  In the case of ARM, we need to do platform-specific detection.  At the moment, we have linux ARM32 and linux AARCH64, but in a few months I expect to have FreeBSD (which may or may not have a similar split between 32 and 64-bit).  Having a single `int supported = 0; ... return (supported)`, and modifying the value of `supported` inside a relevant `#if #endif` construct, keeps this code simpler.

cperciva: Oh yeah, I forgot about the unreachable-code bit.

PR #348, opened 2020-12-28
Add crypto_aesctr_aesni

gperciva:


gperciva:
This is what #340 is moving towards.

The general idea is shown in the `clang -O2` benchmarks:

- 0: `git master`
- 1: switch to a basic `crypto_aesctr_aesni`.  You can see the performance penalty from the extra indirection.
- 2: do the XOR in SSE2.  ~25% speed boost for buffer sizes 16+.
- 3: only use `crypto_aesctr_aesni` if we're doing 16 or more bytes.  The (mostly) avoids the performance penalty for small buffer sizes.

### clang -O2

clang

### comparing compilers, before & after

- nothing fascinating here.  Significant boost in sizes 16+, slowdown for gcc with smaller buffer sizes.

compilers

crypto/crypto_aesctr_aesni.c
```diff
@@ -0,0 +1,230 @@
...
+#include "crypto_aes_aesni.h"
+#include "insecure_memzero.h"
+#include "sysendian.h"
+
+#include "crypto_aesctr_aesni.h"
+
+struct crypto_aesctr {
+	const struct crypto_aes_key * key;
+	uint64_t bytectr;
+	__m128i bufsse;
```
gperciva, line 19: This is illegal; that's the whole reason we have the align-ptr branch.  So I need to fix the bugs in that, before moving forward with this one.

crypto/crypto_aesctr.c
```diff
@@ -150,6 +175,13 @@ crypto_aesctr_stream(struct crypto_aesctr * stream, const uint8_t * inbuf,
 {
 	size_t bytemod;
 
+#if defined(CPUSUPPORT_X86_AESNI)
+	if ((use_aesni == 1) && (buflen >= 16)) {
```
gperciva, line 179: Ouch, I made a terrible mistake here: we can't pick and choose to use the plain vs. aesni one.  At least, not in the current design -- the plain one will be all uninitialized.  It wouldn't be impossible to share data between the two implementations, or to have the aesni version pick and choose which parts of the structure to use.

gperciva:
Once #351 and #340 are merged, this PR only has 4 commits, and 3 of which are significant, and those are 10 or fewer lines.

crypto/crypto_aes_aesni_m128i.h
```diff
@@ -0,0 +1,15 @@
...
+#include <emmintrin.h>
+
+/**
+ * crypto_aes_encrypt_block_aesni_m128i(in, key):
+ * Using the expanded AES key ${key}, encrypt the block ${in} and return the
+ * resulting ciphertext.  This implementation uses x86 AESNI instructions,
+ * and should only be used if CPUSUPPORT_X86_AESNI is defined and
+ * cpusupport_x86_aesni() returns nonzero.
+ */
+__m128i crypto_aes_encrypt_block_aesni_m128i(const __m128i, const void *);
```
cperciva, line 13: Is the `const` in `const __m128i` meaningful?

gperciva: Not to my knowledge.  It was partly because (in my mind) that reinforced that it was an input parameter, and partly just because `void crypto_aes_encrypt_block_aesni(const uint8_t[16], uint8_t[16], const void *);` had one there.  I'll remove it.

gperciva:
Revised code; avoids the `use_aesni` variable, and uses `crypto_aesctr_shared.c` to avoid code duplication.  Performance still looks good:

clang

crypto/crypto_aesctr.c
```diff
@@ -3,17 +3,37 @@
 #include <stdlib.h>
 
 #include "crypto_aes.h"
+#include "cpusupport.h"
```
cperciva, line 6: Ordering?

crypto/crypto_aesctr.h
```diff
@@ -8,6 +8,12 @@
 struct crypto_aes_key;
 struct crypto_aesctr;
 
+/**
+ * crypto_aes_use_x86_aesni(void):
```
cperciva, line 12: Wrong name in the comment.  Also, why expose this?  When would a caller want to use this instead of calling `crypto_aes_use_x86_aesni` directly?

... for that matter, why have it at all?  Unless there's a significant performance benefit from allowing the compiler to inline it and reference a translation-unit-local variable?

gperciva: Why use it instead of `crypto_aes_use_x86_aesni`?:
- if we didn't have `crypto_aes_key_expand()`, this would allow a caller (like `tests/crypto_aesctr/main.c`) to only `#include "crypto_aesctr.h"`.  Granted, since we're almost certain to be calling `crypto_aes_key_expand()` in the same file as the aesctr stuff, this isn't a very motivating reason.

- Performance: on my server, with `clang-11 -O2`, keeping it translation-unit-local variable avoids a penalty of between -2.7% to -0.05% (i.e. not noticeable with a buffer size of 1024).  I could benchmark it on more systems / compilers if you want.

  use-aes

cperciva: Thanks, that graph is enough to convince me that we want to "cache" the value inside that translation unit.

I'm still inclined to not expose it publicly though.  Callers really shouldn't be aware of which AES-CTR code we're going to decide to use anyway (testing notwithstanding) and if they do need to know they can always include the extra header to call the AES function.

crypto/crypto_aesctr.c
```diff
@@ -2,18 +2,51 @@
...
+/**
+ * crypto_aes_use_x86_aesni(void):
+ * Return non-zero if AESNI operations are available.
+ */
+static int
+crypto_aesctr_use_x86_aesni(void)
+{
+	static int aesnigood = -1;
+
+#if defined(CPUSUPPORT_X86_AESNI)
```
cperciva, line 36: Why not make this entire function `#if defined(CPUSUPPORT_X86_AESNI)`?

gperciva: oh right, that construct originated from `crypto_aes_use_x86_aesni()` so that it would `return 0;` if it's not defined, but of course we don't need that any more.

gperciva:
Updated and potentially ready for final review & merge, if you don't want to merge #351 separately.

gperciva:
BTW, there's another potential optimization: using two `__m64i` to handle the input to the cipherblock, and ignoring `stream->pblk` (only updating it at the end of the `wholeblocks`).

So far this is just a sketch of an idea, and even once I've implemeted it, I'm not confident that the speed boost would be worth the additional complexity.  I think it's definitely worth proceeding with the current PR.  But it's not impossible that I'll have another one after this.

PR #349, opened 2020-12-28
Use align_ptr.h for crypto_aesctr

gperciva:


gperciva:
This is an extension of #331, using `align_ptr.h` for crypto_aesctr.  Work in progress, mainly to provide additional motivation for #331.

You may or may not like having "pointer alignment to 16 (or 32) bytes" in the non-hardware-intrinsic code.  I personally think that it's worth it, since having a 16-byte array aligned to 16 bytes in memory is unlikely to hurt performance on any system; also, this would simplify `crypto_aesctr_aesni` #348.  But I can't deny that it spoils the "purity" of the "generic" `crypto_aesctr.c`.

This is still a work in progress; it appears that we get better performance when it's aligned to a 32-byte boundary rather than a 16-byte boundary, and at the moment I can't explain why.  (My first theory is that if `stream->buf` is aligned to 32 bytes, then that guarantees that `stream->pblk` is accessible within the same 32-byte chunk, but so far I have no evidence that this is relevant.)

Timing data with `clang -O2`: the black marks are the "aligned to 32-byte boundarys", and they have either the same performance or a slight improvement.

clang

PR #351, opened 2020-12-30
crypto_aesctr_aesni boilerplate

gperciva:


gperciva:
This does not do anything fancy; that is coming in #348.  (In fact, you probably won't want to pull this PR by itself, until that other PR is ready.  This is mainly so that I can make sure that I've updated the build system correctly, got suitable function names, etc.)

As we'd expect, switching from `crypto_aesctr_stream()` to `crypto_aesctr_aesni_stream()` has a negligible effect on a buffer size of 1024, and a big penalty for small buffer sizes.  (This will be mitigated in #348).

compilers

gperciva:
Rebased onto master, and smooshed the `if (buflen >= 16)` clause.  (You still might not want to merge this on its own, and rather wait until #348 is finished all reviews & fixes.  Up to you; I'm content either way.)

compilers

gperciva:
This now uses the new `crypto_aesctr_use_x86_aesni()` instead of the previous `use_aesni` variable.

gperciva:
Updated and ready for review & merge, if you're ok with merging this boilerplate without any optimizations.  I expanded the commit message for adding `_use_x86_aesni()`.

There's a noticeable performance penalty from `if (buflen >= 16...)` for buffer sizes 2, 4, 8; for large buffers, there's also the penalty of switching to the (currently-unoptimized) `crypto_aesctr_aesni_stream()`.

(I personally think that it's worth merging now, as that will make the git history as clear as possible, but if you'd rather do #348 all at once, that's fine too.)

clang

PR #352, opened 2020-12-30
metabuild.sh: use a better fake cpusupport

gperciva:


gperciva:
Our "cpu intrinsic" .c files look like this:

    #include "cpusupport.h"
    #ifdef CPUSUPPORT_SOME_THING
    #include "header.h"
    ...
    #endif

so if we have a fake blank cpusupport.h, the generated Makefile won't pick up a dependency on "header.h".  This PR fixes `metabuild.sh` so that it picks up all dependencies.

cperciva:
This seems like a better solution since it avoids needing to duplicate the list of features: https://github.com/Tarsnap/libcperciva/pull/353/

Might need some tweaking, not sure if that's how metabuild should be finding the cpusupport code.

PR #353, opened 2020-12-30
cpusupport: Add "--all" flag to cpusupport.sh

cperciva:
Use this in metabuild.sh to create a fake cpusupport-config.h which
indicates that the compiler supports everything (including options
from multiple different architectures simultaneously).

### DIFF
--- a/release-tools/metabuild.sh
+++ b/release-tools/metabuild.sh
@@ -23,10 +23,17 @@ fi
 # Set up directories
 cd ${D}
 SUBDIR_DEPTH=$(${MAKEBSD} -V SUBDIR_DEPTH)
+LIBCPERCIVA_DIR=$(${MAKEBSD} -V LIBCPERCIVA_DIR)
 
-# Set up *-config.h so that we don't have missing headers
-rm -f ${SUBDIR_DEPTH}/cpusupport-config.h
-touch ${SUBDIR_DEPTH}/cpusupport-config.h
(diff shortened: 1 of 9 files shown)


release-tools/metabuild.sh
```diff
@@ -26,7 +26,7 @@ SUBDIR_DEPTH=$(${MAKEBSD} -V SUBDIR_DEPTH)
 
 # Set up *-config.h so that we don't have missing headers
 rm -f ${SUBDIR_DEPTH}/cpusupport-config.h
-touch ${SUBDIR_DEPTH}/cpusupport-config.h
+command -p sh ${SUBDIR_DEPTH}/cpusupport/Build/cpusupport.sh "${PATH}" --all > ${SUBDIR_DEPTH}/cpusupport-config.h
```
cperciva, line 29: Is this right, or do we need ${LIBCPERCIVA} or something else here?

cperciva: Obviously not relevant for this repo but for when we use this in other projects

gperciva:
Added a REBASE commit.  My naive solution failed on `tests/buildnothing/`, so I'm really glad that I added that almost pointless test in 7a4e9295b0cefd281cf8d1e19dd9369a2dd4babf!

Verified that it works in spiped.

cperciva:
LGTM, please rebase.

PR #356, opened 2020-12-31
crypto_aesctr_aesni lulz

gperciva:


gperciva:
Since we want to avoid code duplication, but also want to keep the code inline without function calls, how about flipping it?  Shove the generic code into `crypto_aesctr_internal.h`, then call those inline functions from `crypto_aesctr_stream()` and `crypto_aesctr_aesni_stream()` as necessary?

(I played fast and loose with `const uint8_t *` here, but if you're not opposed to this idea in theory, I can clean it up.)

PR #357, opened 2021-01-02
Add crypto_aesctr_shared

gperciva:


gperciva:
This begins the process of adding `crypto_aesctr_aesni`.  I didn't bother including the changes from `make Makefiles` (to add a dependency on `crypto_aesctr_shared.c`), since those will be easily picked up by a later PR in this series.

crypto/crypto_aesctr.c
```diff
@@ -7,13 +7,7 @@
...
 
 #include "crypto_aesctr.h"
-
-struct crypto_aesctr {
-	const struct crypto_aes_key * key;
-	uint64_t bytectr;
-	uint8_t buf[16];
-	uint8_t pblk[16];
-};
+#include "crypto_aesctr_shared.c"
```
cperciva, line 10: This needs a large explanatory comment.  Something along the lines of

```
/**
 * In order to optimize AES-CTR, it is desirable to separate out the handling
 * of individual bytes of data vs. the handling of complete (16 byte) blocks.
 * The handling of blocks in turn can be optimized further using CPU intrinsics,
 * e.g. SSE2 on x86 CPUs; however while the byte-at-once code remains the
 * same across platforms it should be inlined into the same (CPU feature
 * specific) routines for performance reasons.
 *
 * In order to allow those generic functions to be inlined into multiple functions
 * in separate translation units, we place them into a "shared" C file which is
 * included in each of the platform-specific variants.
 */
```

gperciva:
Rebased to include the comment.  Are you sure you want it in `crypto_aesctr.c` instead of in `crypto_aesctr_shared.c`?

cperciva:
Hmm, good point.  I'd be inclined to have the comment in (each of) the non-shared files and have the comment at the top of `_shared.c` make some mention of how there's a comment in the files including it about why the code is structured that way.

Sorry, it's too late for me to come up with coherent text for that, but I'm sure you'll come up with something.

gperciva:
Rebased.  For the top of `_shared.c`, I couldn't come up with anything better than

>  This code is shared between crypto_aesctr*.c files, and should not be
> compiled as a separate translation unit.  For details, see the comments in
> those files.

which isn't exactly inspired, but it should get the job done.

PR #358, opened 2021-01-12
SHA256 prep

gperciva:


gperciva:
This is mainly to get the "process 100 MB instead of 1 GB" commit into master.  There's some profoundly weird stuff that's coming in a following PR.

alg/sha256.c
```diff
@@ -193,7 +198,7 @@ SHA256_Transform(uint32_t state[static restrict 8],
 
 #if defined(CPUSUPPORT_X86_SHANI) && defined(CPUSUPPORT_X86_SSSE3)
 	/* Use SHANI if we can. */
-	if (useshani()) {
+	if (usehw()) {
```
gperciva, line 201: You might wonder why we don't make this `usehw() == 1`, since the general idea is preparing for non-SHANI sha256 cpu intrinsics.

Well, that change by itself increases the SHA256 software speed of `gcc -O2` by 10%, and drops the speed of `gcc -O3 -march=native` by 3%.  (clang is mostly unaffected)

While working on that problem (trying to understand what's happening), I got annoyed at the time it took to process 36 GB every test (1 GB, * 3 for sizes, * 4 for different compilers & flags, * 3 repetitions).  Hence wanting to get this PR in first, then working off that new base.

PR #359, opened 2021-01-12
sha256: check for exact usehw number

gperciva:


gperciva:
This should be a trivial change, and indeed, on clang it is.  But it on `gcc -O2` is gives +10% speed, and `gcc -O3 -march=native` it's a -3% penalty.

compilers

The diff in clang's assembly code, other than minor tweaks to memory addresses such as
```
     1f1e:      	mov	dword ptr [rax + 24], r12d
-    1f22:      	mov	ecx, dword ptr [rbp - 76]
+    1f22:      	mov	ecx, dword ptr [rbp - 80]
     1f25:      	mov	dword ptr [rax + 28], ecx
```

is exactly what we'd expect:
```
 ; 	if (usehw() == 1) {
-    1398:      	test	eax, eax
-    139a:      	je	0x13bf <SHA256_Transform+0x1bf>
+    1398:      	cmp	eax, 1
+    139b:      	jne	0x13c1 <SHA256_Transform+0x1c1>
 ; 		SHA256_Transform_shani(state, block);
...
 ; 	if (usehw() == 1) {
-    13bb:      	test	eax, eax
-    13bd:      	jne	0x139c <SHA256_Transform+0x19c>
-    13bf:      	xor	eax, eax
-    13c1:      	nop	word ptr cs:[rax + rax]
-    13cb:      	nop	dword ptr [rax + rax]
+    13bc:      	cmp	eax, 1
+    13bf:      	je	0x139d <SHA256_Transform+0x19d>
+    13c1:      	xor	eax, eax
+    13c3:      	nop	word ptr cs:[rax + rax]
+    13cd:      	nop	dword ptr [rax]
 ; 	    ((uint32_t)(p[1]) << 16) + ((uint32_t)(p[0]) << 24));
```

Discussion about gcc's assembly coming in the next comment.

gperciva:
gcc10 (and the unreleased gcc11 development version) produce completely different assembly as a result of this PR.  It turns out that gcc9 has the same speed in before & after, and the assembly has the expected (tiny) diff.

I'll attach the dissassembly for gcc10 before & after.  It looks like the "before" version has a lot more inlining: in the C code (written as comments), `if (usehw() == 1)` appears 6 times.  In the "after" version, it only appears once.  In the clang versions, it appears twice in each of the before & after versions.

I see no speed difference on linux t3a or t4g.

I'm not certain where to go from here.  If you're content to shrug and say "ok, gcc's being weird", that's fine with me.  There's also more weirdness coming with another change in usehw.  I'm vaguely suspicious about the "static variable inside a function" method (instead of setting a static variable) for selecting which intrinsics to use, but I have no solid evidence against it.

gcc assembly, before & after:
[0-sha256.o.txt](https://github.com/Tarsnap/libcperciva/files/5804678/0-sha256.o.txt)
[1-sha256.o.txt](https://github.com/Tarsnap/libcperciva/files/5804679/1-sha256.o.txt)

PR #360, opened 2021-01-12
SHA256: refactor to allow for different intrinsics

gperciva:


gperciva:
This builds on #359, but shows more weirdness.

The final REBASE? commit shows a surprising mitigation of performance degradation, but I have no good reason behind why that would be so.  My vague theory is that it hints to the compiler that we don't expect to loop the `while()` multiple times.

`gcc10 -O3 -march=native`:

gcc-o3

`clang-11 O2`:

clang

gperciva:
More generally, when faced with such weirdness, what evidence should I be gathering?  It feels like a few plots and snippets of dissassembly output is inadequate, but I'm not certain what my next step should be.

gperciva:
FWIW, it looks like the first time we had a `usexyz()` function, including the `while (aesnigood == -1)`, was
https://github.com/Tarsnap/libcperciva/commit/13a6f864485a5c75d806055ce8ad01b57d53d599.

cperciva:
> More generally, when faced with such weirdness, what evidence should I be gathering? It feels like a few plots and snippets of dissassembly output is inadequate, but I'm not certain what my next step should be.

In general, I wouldn't worry too much about trying to "coerce" compilers into producing optimized code.  Something which speeds up one compiler on one CPU by 5% can slow down another compiler by 5%... or slow down the same compiler on a different CPU by 5%.  If changing `while` to `if` in a way which doesn't change the program semantics also provides a speedup, great -- I don't know why it was a `while` in the first place.  But for cases like `if (foo)` vs `if (foo == 1)`, you're probably just tickling an edge case in the "decide if this function is too complicated to inline" heuristic and future compilers are likely to change the behaviour anyway.

gperciva:
Ok, slightly revised the PR.  Still a boost to `gcc -O2` and mild penalty to `gcc -O3 -march=native`.  Potentially ready for merge.

compilers

cperciva:
I'm confused, there's only one commit here but the performance graph looks like it's showing two changes?

gperciva:
Oops, sorry.  I haven't figured out how to upload images directly from my server, so I normally copy the files from the server to my pi, then upload from there.  Looks like I forgot the middle step, and uploaded an older image (because I reuse filenames, and don't delete old images).

(An earlier version of this PR changed the `while` to an `if`.  After I added the `goto`, casual testing suggested that I didn't need make that change, so I was experimenting with the "final" PR by adding an extra commit.)

I've re-created it, making sure that the additional commit plotted here is 55025ec.

compilers

This shows the same changes as before (if we ignore the `while` labels), with one exception: it now appears that `clang -O2` with a buffer size of 256 is faster than the previous png.  In both cases, the same binary produces equal speeds with 1024 and 10000. (I don't think that's particularly interesting; I only mention it in case you compare the images and wonder about it.)

cperciva:
The more I look at this, the more I'm confused by the `== 1` part.  Is the idea here that `usehw` might return different values to tell us which cpufeature-optimized routines to use?  I thought that for any particular set of CPUSUPPORT_FOO_BAR flags there would only be one optimized version we would want to consider so `usehw` could just return true/false.

gperciva:
Yes, the idea is that `usehw` would be able to tell us which cpufeature-optimized routine to use.

Remember that `CPUFEATURE_FOO_BAR` tells us what the compiler can do; `cpusupport_foo_bar()` tells us what the cpu thinks it can do, and `hwtest()` tests whether the cpu can actually do it.  In particular, basically all modern x86 compilers will enable CPUSUPPORT_X86_SHANI, but basically no modern x86 cpus will `cpusupport_x86_shani()`.  So we need to handle "compile can do A, but we can't actually do A; compiler can do B, and we can do B".

If we only cared about correctness, the `_Transform` could contain something like:

```
if (cpusupport_x86_shani() && cpusupport_x86_ssse3() && hwtest(...))
    SHA256_Transform_shani(state, block);
if (cpusupport_x86_avx1() && hwtest(...))
    SHA256_Transform_avx1(state, block);
if (cpusupport_x86_sse2() && hwtest(...))
    SHA256_Transform_sse2(state, block);
```

(along with some `#ifdef` and the like)

but then we'd be re-running `cpusupport_foo_bar()` and `hwtest()` all the time.  So the idea is that `usehw()` will cache the result of `cpusupport_foo_bar()` and the `hwtest()`.

gperciva:
For example, see 'alg/sha256.c' in https://github.com/Tarsnap/libcperciva/pull/363/files#diff-f3bd613685e03033a3c53403ed23d578fab8405036e0a1155cf45f15346c0f8bR220

cperciva:
Got it.  I would make that a `switch` statement though.

gperciva:
Updated to use `switch`.

Performance in gcc is slightly worse than the version with an `if` statement, but the eventual performance of `sha256_sse2` is slightly improved.  I can't track down the exact changes that gcc decided to make, because the whole file is reorganized.  (gcc inlines `SHA256_Transform()`, which is used in several places in that file.)

compilers

alg/sha256.c
```diff
@@ -196,13 +194,17 @@ SHA256_Transform(uint32_t state[static restrict 8],
...
+	switch(usehw()) {
 #if defined(CPUSUPPORT_X86_SHANI) && defined(CPUSUPPORT_X86_SSSE3)
-	/* Use SHANI if we can. */
-	if (usehw()) {
+	case 1:
 		SHA256_Transform_shani(state, block);
 		return;
-	}
 #endif
+	default:
```
cperciva, line 204: Does anything change if we make this
```
0:
    break;
default:
    /* UNREACHABLE */
    assert(0);
```
in order to make it clear that we should be covering all of the same cases as `usehw` can emit?

cperciva: Change performance-wise, I mean.  Sometimes compilers can make smarter deductions about reachability given hints like this; sometimes they get confused and do stupid things.

If there's no performance difference I'd prefer to have the "default" case to go a failed assertion, just as a safeguard against future bugs.

gperciva: Yes, it makes a difference... in a very unpleasant manner.  It's all over the map!  :(

compilers

My reading:
- for `clang`, both `-O2` and `-O3 -march=native`, the best option is the current PR: `warn0(); break;`.  Note that we only expect this to ever occur during development, so maybe a warning message instead of an `abort()` is ok?
- for `gcc -O3 -march=native`, the best option is the same as `clang`, although only slightly.
- for `gcc -O2`, the best speed comes from `warn0(); abort()`.  That gives a +5% speed improvement over the baseline (and over `warn0(); break;`).
- in general, I'm slightly dismayed that there's *this* much variation in the speed.  I'm going to test using a callback function, instead of selecting an option this way.

gperciva:
Oh wait, this might not be a great test -- it's measuring the time to execute the software SHA2.  If we add SSE2 and ARM versions, then almost nobody will be using the purely software version any more, so the time it takes to get to an intrinsic version would be more important.

gperciva:
Ok, my reluctant recommendation is to use `warn0(); assert();`.  That is to say, the current PR, with the `REBASE` commit merged with the previous.  I left it separate for now so that you could see the effect of that one-line change.

I'm reluctant because I cannot explain why adding `warn0()` is helpful.  Empirically, it's definitely worth it, but this seems rather random to me. :(

Adding the `warn0` in there changes a lot of instructions relating to the SHA256 round calculation; the dissassembled object file is a mess.  At the moment, my best guess is that the longer code (due to `warn0`) allows the compiler to find a better way of aligning instructions with desirable boundaries.

My experiment with using a function pointer instead of `usehw()` showed somewhat worse performance than this PR.

### SHANI instructions, overall changes

- all compilers don't like the new "allow for different instrinics" code, but adding the extra `warn0` counteracts this penalty (?).

shani-overall


### SHANI instructions, close-up of 10000 buffer size

- same as above

shani-closeup

### software instructions, overall

- `clang -O2`: slight penalty
- `clang -O3 -march=native`: boost
- `gcc -O2`: significant penalty from "allow for different instrics", then double that as a bonus when we add `warn0` (?).
- `gcc -O3 -march=native`: penalty, somewhat mitigated by the `warn0`.
- I didn't bother with a "close up" png for this one, since the "overall" png is clear.

plain-overall

gperciva:
Based on recent investigations into using a function pointer, it looks like it's better for sha256 if we're using an intrinsic (such as SSE2), but it's worse if we're defaulting to plain C code.

My suggestion is to merge this PR (after rebasing), add the ARM, SSE2, and possibly other intrinsics, then re-evaluate exactly which method of function selection is the best.

Note to self: when looking at function selection methods, the "ideal case" is simply commenting out anything other than the fastest function for that particular platform.  That 1) gives a sanity check of "the function selection can't be faster than this", and 2) allows us to see concrete data that might be something like "the current method has a -0.5% penalty compared to the ideal, so it's not worth fussing about it".

PR #361, opened 2021-01-13
BUILDING: add some comments about ARM / AARCH64

gperciva:


cperciva:
I'm not sure what the relevance of `march=native` is here.  That turns on flags based on what the *currently running* CPU supports; what we want is to enable everything the *compiler* supports...?

gperciva:
Hmm.  I suppose that it's mostly supporting the "no perfect compiler" statement at this point.  Also, I found it surpringly hard to find a definitive answer to "what -march is suitable for AWS t4g?", so I was trying to include it somewhere.

I agree that it's not a good fit, though.  I'll revisit this.

gperciva:
Closing this.  Most of that info is already in the commit message for d7976b19e80d7be55384d498351940b71d94bf06.  If I was adding crc32c for the first time, I'd edit that commit message a little bit, but nothing drastic.

PR #362, opened 2021-01-14
Test & improve sysendian.h

gperciva:


gperciva:
In some of my later (more speculative) work on aes-ctr, I noticed a lack of `bswap` in the gcc assembly, and I found that replacing `#include "sysendian.h"` with the system `<sys/endian.h>` improved performance.

This PR adds a test, and vastly increases performance for gcc.  I orginially had a much simpler test without `perftest.h`, but then I kept on adding hacks to make our `perftest-*.sh` scripts work with it, so I figured I might as well use our typical test infrastructure.

There's no effect for 32-bit endianness on clang, and a minor speed boost for 64-bit endianness on clang (it goes from 6 `bswap`s to 8 `bswap`s, but I haven't bothered to track down exactly which case it's improved).

FWIW, freebsd's `sys/endian.h` uses bitwise OR instead of addition, so this change brings us closer in line with that system.

compilers

I might have another one or two additions to `endian.h`, but they're less convincing.  The first step is obviously to get the test infrastructure in place.  Then I figured that the "bitwise OR" was simple enough to include as well, mostly to provide motivation for a test framework for something as simple as endianness.

tests/sysendian/main.c
```diff
@@ -0,0 +1,235 @@
...
+	be_v = vbe;					\
+	vle = le_v;					\
+	vbe = be_v;					\
+	if (system_le == 1) {				\
+		TEST_ENDIAN(be, N, vle, vbe, vle, vbe);	\
+		TEST_ENDIAN(be, N, vbe, vle, vbe, vle);	\
+		TEST_ENDIAN(le, N, vle, vle, vle, vle);	\
+		TEST_ENDIAN(le, N, vbe, vbe, vbe, vbe);	\
+	} else {					\
+		TEST_ENDIAN(be, N, vle, vle, vle, vle);	\
```
gperciva, line 101: I haven't tested the big-endian code here.  I could theoretically do that in something like QEMU, but at the moment I don't think that's necessary.

gperciva:
BTW: this PR by itself gives a +10% bonus to sha256 with `gcc -O2`.

gcc

cperciva:
I'm confused by the test.  Why not just have an array of tests of the form `{1, [0, 0, 0, 1], [1, 0, 0, 0]}`?

gperciva:
What if an evil cosmic ray bit-flipped the definition of `be32dec`, such that it was
```
-        return ((uint32_t)(p[3]) | ((uint32_t)(p[2]) << 8) |
+        return ((uint32_t)(p[3]) | ((uint32_t)(p[2]) << 7) |
```

When I was testing the test framework, I did randomly change numbers in `sysendian.h`, and that made me realize that only using `0x0123` as input (and not `0x3210`) wasn't sufficient, because deliberately messing up `p[0]` didn't cause the test to fail.

Granted, a screwy optimization is highly unlikely to pass `[0, 0, 0, 1]` but not `[1, 2, 3, 4]`.  And even if it did, the alg and crypto tests would break.  So I'm content to revise the tests as you suggest.

gperciva:
Actually, wait.  `{1, [0, 0, 0, 1], [1, 0, 0, 0]}` tests whether we can byte-swap two arrays.  But functions like `be32dec` don't byte-swap -- on x86 and arm they do, but on a big-endian system, they're a nop.

I'm sure that it's possible to combine le32enc, be32enc, le32dec, and be32dec in some manner such that `[0,0,0,1] -> [1,0,0,0]` checks whether they all work on little-endian and big-endian systems.  But I think it's more straightforward to test each function separately, as in this PR.

cperciva:
Maybe I should have been clearer.  With that test case you can verify that `be32enc(1)` gives the first array and `le32enc(1)` gives the second array.  And similarly test the `dec` functions.

Then you would also have test cases for each of the other powers of 2.

gperciva:
Ah, ok!  I can do that.

gperciva:
Much improved code, completely revised.  There were major problems with the previous timing code; the current results are more much sane.

Note that we now see improvement in clang as well as gcc.  What was happening earlier was that I was copying from a `const` array to a temp volatile array byte-by-byte, and clang started to do the byte-swapping (via the naive method) before it had finished copying the array over.  By contrast, the updated code (which is recognized as being a `bswap`) had to wait until the copying was finished before it could do the swap.  As a result, there was no apparent change of speed in clang.

Now, I'm using a volatile variable (such as `volatile uint64_t`) to copy the input value as a single operation.  I think that's a better test for the way that we use these functions.

compilers

gperciva:
Incidently, in addition to encountering the lack of `bswap` in aes-ctr, I became interested in this while investigating whether it's worth having a `sha256_ssse3.c` in addition to `sha256_sse2.c`.  `PSHUFB` would be a nice addition for initializing the message schedule, but if we have a fast `be32dec()` anyway, then that's not so useful.

cperciva:
I'm a bit dubious about how useful a performance test for this is -- I'm all for microbenchmarks but this is a bit excessively micro and I would rather see how this affects the performance of sha256 -- but I guess there's no harm in having it.

gperciva:
Ok, here's the sha256 performance; similar to https://github.com/Tarsnap/libcperciva/pull/362#issuecomment-760025019, but extended for clang as well as gcc.  No change with clang; clear improvement with gcc.

compilers

PR #363, opened 2021-01-16
sha256 SSE2

gperciva:


gperciva:
This is not for consideration; it's a work in progress.  The only point is to give an example of `usehw()` meaningfully returning different values, to aid in discussing #360.

gperciva:
As such, the only commit of interest is 82a3caa.

gperciva:
This includes #369, which you may or may not want to merge first.

Overall performance:
compilers

Close-up of 1024:
close

Using SSE2 with these two optimizations (SPAN_ONE_THREE and the BBAA rotate trick) is a definite improvement.

The MSG4 commit (red in the latest plot) does show a significant drop with clang; that arises due to combining 4 values at once.  If we just switch MSCH from a `#define` to a function, there's no penalty.  I didn't think that it was worth keeping that as a separate commit, though.  And if I try switching the current MSG4 to a `#define`, I don't see any performance boost.

I have a whole branch of other optimizations, but each one has at least one case in which it's notably worse.  I'm continuing to investigate those, and I expect that after this PR is merged, I'll have one or two other PRs in which we can discuss the potentially questionable optimizations.

PR #365, opened 2021-01-19
Optimize crypto_aesctr_aesni.c with local variables

gperciva:


gperciva:
This more than doubles the speed, from just under 500 Mb/s to over 1200 Mb/s (and over 1600 Mb/s with `gcc`!).  The overall plot is a bit hard to read, but you can see the overall trend (the purple and black bars are the final two commits in this series).

compilers

Here's a close-up of 1024:
clang

There will be at least one more PR after this.  This PR is focused on using local variables instead of `stream->`.  There's more gains (in particular, bringing `clang` up to 1790 Mb/s), but some of them are fairly nitpicky and it's worth discussing them on their own.

gperciva:
I just noticed a typo in the commit message for 4182125.  I'll hold off on changing it now, because that would change the git hashes, which might make the plots harder to understand.  I'll rebase the branch once I have tentative approval for the PR.

crypto/crypto_aesctr_aesni.c
```diff
@@ -26,38 +27,62 @@
...
+	for (i = 0; i < num_blocks; i++) {
+		/* Prepare counter. */
+		block_counter_be_arr[7]++;
+		if (block_counter_be_arr[7] == 0) {
+			/*
+			 * If incrementing the least significant byte resulted
+			 * in it wrapping, re-encode the complete 64-bit
+			 * value.
+			 */
+			be64enc(block_counter_be_arr, block_counter);
```
cperciva, line 60: This may be a dumb question, but is there a way to use SSE instructions to get a 64-bit value into a register in big-endian format?

cperciva:
While I'm here: We should probably expand the CPUSUPPORT_X86_AESNI test to include more SSE instructions, now that we're using more of them in "aesni" code.

crypto/crypto_aesctr_aesni.c
gperciva: Yes:
```
static inline __m128i
_mm_bswaplo_epi64(__m128i x)
{
	/* Swap bytes in each 16-bit word. */
        __m128i a = _mm_or_si128(
		_mm_slli_epi16(x, 8),
		_mm_srli_epi16(x, 8));

	/* Reverse 16-bit words in the lowest qword. */
	a = _mm_shufflelo_epi16(a, _MM_SHUFFLE(0, 1, 2, 3));

	return (a);
}
```

(in case anybody lands here from google: this is a personal version that only byte-swaps the lower qword and gives garbage in the upper qword, because that's all we need in aesctr.  To make it work for the entire 128 bits, just add a `_mm_shufflehi_epi16()` in there.)

That's one of the commits that might come in the later PR (in my `-exploratory` branch).  IIRC it gives a +5% bonus to `clang` and a -10% penalty to `gcc`.  (Or it might be the commit that gives a +10% bonus to everything other than `clang -O3 -march=native`, which gets a -40% penalty (!).  I've got commits like that for both sha256 and aesctr.)

gperciva:
Testing SSE (or SSE2?) instructions: I did look into this, and my understanding is that since the AESNI instructions use `__m128i`, and that was only introduced in SSE2, we're guaranteed that if we have AESNI then we'll also have at least SSE2.  Also, the AESNI test already includes `_mm_loadu_si128()`, which is `movdqu`, which was added in SSE2 [1, 2].

At any rate, if we get any reports of it not working on a certain CPU, I'd be very curious to investigate!

[1] https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=236,3416,3416,3416,3416&text=_mm_loadu_si128
[2] it's possible that Intel might have a typo on their site, but this one also agrees on it being SSE2:
https://www.felixcloutier.com/x86/movdqu:vmovdqu8:vmovdqu16:vmovdqu32:vmovdqu64

PR #366, opened 2021-01-21
sha256: clarify sanitizing the stack

gperciva:


alg/sha256.c
```diff
@@ -392,7 +393,7 @@ SHA256_Final(uint8_t digest[32], SHA256_CTX * ctx)
 	insecure_memzero(ctx, sizeof(SHA256_CTX));
 
 	/* Clean the stack. */
-	insecure_memzero(tmp32, 288);
+	insecure_memzero(tmp32, sizeof(uint32_t) * 72);
```
gperciva, line 396: I should probably automatically recognize that 288 is `sizeof(uint32_t) * 72`, but it's not immediately obvious to me, so more than once I've panicked when browsing this code.  I eventually remember out that 32/8 = 4, but I think that `sizeof()` makes the code a lot more clear.

alg/sha256.c
```diff
@@ -187,7 +187,8 @@ usehw(void)
 
 /*
  * SHA256 block compression function.  The 256-bit state is transformed via
- * the 512-bit input block to produce a new state.
+ * the 512-bit input block to produce a new state.  The arrays W and S may be
```
gperciva, line 190: This comment might be overkill in `sha256.c`, but it's definitely necessary in the upcoming `SHA256_Transform_sse2()`, and I figured that the functions should have similar comments unless they had to diverge.

(Note that we don't need it in `sha256_shani.c`, because `SHA256_Transform_shani()` doesn't take W and S as paramters.

PR #369, opened 2021-02-03
SHA256 SSE2 boilerplate

gperciva:


gperciva:
This PR (which builds on #360) takes care of adjusting infrastructure to prepare for using SSE2 for SHA256.  It doesn't introduce SSE2 intrinsics yet; that will be the next PR.

gperciva:
This change by itself -- which does not actually use any SSE2 intrinsics -- has a mixture of performance implications.
- clang -O2, gcc -O3 -march=native: slight improvement
- clang -O3 -march=native: notable drop (-5%)
- gcc -O2: significant drop (-10%)

If you'd prefer to do #363 all at once (SSE2 boilerplate + easy optimizations) so that there's no performance drop, that's fine.

I'd previously said that we should compare different sets of intrinsics before deciding which ones to support.  So far I've found minimal improvements between SSE2 and SSSE3 and SSE4.2.  Also, SSE2 is the baseline for the amd64 platform.  With that in mind, I think that we should go ahead with SSE2; once that's in, we might add other instrinsics if they provide a sufficiently interesting boost compared to SSE2.

compilers

PR #370, opened 2021-02-07
crypto_aesctr_aesni.c: always do be64enc

gperciva:


gperciva:
- Vast improvement for clang (~1250 MB/s to 1700 or 1800!)
- Noticeable improvement for gcc -O3 -march=native (1500 to 1630)
- slight penalty (?) for gcc -O2 (1620 to 1580)

Overall performance:
compilers


Close-up of 1024:
close

Using SSE2 for the bswap has a mild performance penalty on clang, and increased variability on gcc.  I'm still investigating, but if I can get it working well then it can be a separate PR.

crypto/crypto_aesctr_aesni.c
```diff
@@ -45,6 +46,9 @@ crypto_aesctr_aesni_stream_wholeblocks(struct crypto_aesctr * stream,
 	memcpy(block_counter_be_arr, stream->pblk + 8, 8);
 	block_counter = stream->bytectr / 16;
 
+	/* clang doesn't realize that we'll do the for() loop at least once. */
+	block_counter_be_sse = _mm_loadu_si64(stream->pblk + 8);
```
gperciva, line 48: In theory, we could remove this line and set `__m128i block_counter_be_sse = {0};` instead.  But if we somehow got to this function and didn't run the `for` loop, we'd end up with garbage in `pblk`.  I can't measure any speed difference between the versions, and I figured that this was slightly more clear.

cperciva:
I'll admit that "add block_counter_be_sse" is mostly clear by itself, but if you're going to say more than that, I'm not sure that "This" is really very helpful...

gperciva:
whoops!

crypto/crypto_aesctr_aesni.c
cperciva: What happens if you remove this line and add `if (num_blocks > 0)` before the `_mm_storeu_si64`?

gperciva: Same warning -- clang is really worried about this variable!

gperciva: I added a REBASE commit in case you wanted to take a look at it.

gperciva:
Interesting!  That "REBASE avoid silly clang thing" commit has a significant clang penalty, but a gcc bonus.  If we adopted it, then the end effect of this PR would be positive on all compilers.

Close-up of 1024.  Note the drop in the blue bars (`clang -O2`), but the rise in red (`gcc -O2)`.

close

cperciva:
Are the compilers any less confused if you add `assert(num_blocks > 0)`?

gperciva:
... and yes, I did try adding `assert(num_blocks > 0);` by itself.  That still didn't make clang realize that we'd be executing the `for` loop.

Something else which *does* work is switching the loop to a `do... while`.  At least, it works at satisfying the clang warning, but the speed drops to 1695.

gperciva:
At this point, I could start diffing the assembly output, if you want?   (well, tomorrow, because I have factorio in 5 minutes)

cperciva:
Weird.  I don't want to have "unnecessary code which keeps the compiler happy" if we can possibly avoid it, since that's something which may slow things down with future compilers (not to mention confusing future readers of the code).  Let's see if we can find some way to convince the compilers to be sane (but not tonight).

gperciva:
Significant performance boost for clang, minor boost for gcc.

Overall:
compilers

Close-up of 1024:
close

gperciva:
(the git hashes don't line up because I found a typo in a commit message just after creating the graphs)

crypto/crypto_aesctr_aesni.c
```diff
@@ -51,7 +51,12 @@ crypto_aesctr_aesni_stream_wholeblocks(struct crypto_aesctr * stream,
 	/* How many blocks should we process? */
 	num_blocks = (*buflen) / 16;
 
-	for (i = 0; i < num_blocks; i++) {
+	/*
+	 * This is 'for (i = num_blocks; i > 0; i--)', but ensuring that the
```
cperciva, line 54: Is counting down better than counting up?

If we count up, we could use `*inbuf + i * 16` and similarly for `outbuf` in order to avoid updating them on every loop.

gperciva: Yes, counting down is better.  I added an extra variable `j = 0; ... j++` and saw no speed change (possibly because it was compiled out).  But testing the loop comparison with:
```
#if 0
        } while (i > 0);
#else
        } while (j < num_blocks);
#endif
```

showed 1695 for the `(i > 0)` [as expected], but 1489 for `(j < num_blocks)`.  That actually drops to 1467 when I try the `(*arr + j*16)` trick.  I would have expected the compiler to translate both loops to "counting down", since "jump not equal" is faster than "compare two variables, then maybe jump".

I wouldn't have expected `*inbuf += 16; outbuf += 16;` to have a performance effect, since those can be calculated while the SSE part is encrypting the cipherblock.

I also played around with
```
__m128i * out = *outbuf;
...
     _mm_storeu_si128(&out[j], bufsse);
```

and again saw values in the upper 1400s.

cperciva: Weird.  Oh well, the code is a bit ugly but I'll put up with it for the performance boost.

Issue #371, opened 2021-02-07
automatically collect cpusupport CFLAGS


cperciva:
The 

```
cflags-foo.o:
	@echo '$${CFLAGS_FOO_BAR}'
```

notations are starting to get a bit unwieldy, and I worry that it may also end up becoming error-prone once there's a large number of them.

I'm wondering if we should have something along the lines of

```
/**
 * CPUSUPPORT CFLAGS: FOO_BAR FOO_BAZ
 */
```

in the relevant C source, and have `metabuild.sh` do something like `CF=$(grep 'CPUSUPPORT CFLAGS:' $S | cut -f 2- -d : | while read X; do printf "$${CFLAGS_%s}\n" "$X"; done)` instead of the current `$(${MAKEBSD} cflags-${F})`.

cperciva:
@gperciva If you have time at some point, can you try implementing this so I can see how much it cleans up (or not) the tree?

gperciva:
Can do... but a failure to add the correct CFLAGS_FOO_BAR will just be a compile failure [1], and thus easily noticeable.  If not with a plain old `make`, then at least in the github Action with `clang -Weverything -Werror`.

[1] at least, if the file uses the relevant intrinsics.  Admittedly in `master` at the very moment, `sha256_sse2.c` doesn't use any such intrinsics, but we'd run into compile failures as soon as we start to use them.  Source: personal experience from a few days ago.  :)

gperciva:
Also, I must admit that in that particular case, the problem was that I'd forgotten to update `tests/aws/Makefile.BSD`, which wasn't an obvious place to look when working on `sha256_sse2.c`.  So this system would have saved me some time.

cperciva:
It will result in a compilation failure -- on systems with the relevant compiler support.  If we forget the cflags needed for an ARM-optimized file, it will still compile just fine on x86 since after preprocessing it will be empty.

I'd like to avoid the potential for doing a release which fails to compile on some systems.

PR #374, opened 2021-02-08
Get CFLAGS from c files

gperciva:


gperciva:
#371.

This includes #373.  Disregarding that PR, this doesn't change the generated Makefiles at all, and it looks much better.

release-tools/metabuild.sh
```diff
@@ -52,6 +52,19 @@ add_makefile_prog() {
 	fi
 }
 
+get_cpusupport_cflags() {
+	src=$1
```
cperciva, line 56: You're not using `src` here.  Also, `src` is being set to the name of the `.o` file...

release-tools/metabuild.sh
```diff
@@ -52,6 +52,19 @@ add_makefile_prog() {
...
+
+	# Get the flags from the .c file and format them.
+	str=$(grep 'CPUSUPPORT CFLAGS:' $S | cut -f 2- -d :)
+	cflags_unstripped=$(for X in ${str} ; do		\
```
cperciva, line 60: This seems unnecessarily complicated.  Why not just

```
for X in ${str}; do
    printf " \${CFLAGS_%s}" "$X";
done | sed 's/^ //'
```

and omit the `cflags_unstripped` and `cpusupport_cflags` variables entirely?

gperciva:
Revised.

gperciva:
oops, I just noticed that I'm inconsistent between "cpusupport" and "CPUSUPPORT" in the commit messages.

cperciva:
Along with fixing the commit message (I would say CPUSUPPORT in both cases since we're talking about compile flags), can you add something to STYLE about the magic comments at (almost) the top of files which use special cpu features?

gperciva:
updated

PR #375, opened 2021-02-08
sha256_sse2: use SSE2 for byte-swapping

gperciva:


gperciva:
Notable improvement for everything other than `gcc -O3 -march=native`, which is neutral or very slighly slower.

compilers

(The obvious "next step" is to keep the values of `W` in `__m128i` and use them in `MSG4` in a "rotating state", but that still has an unexplained performance drop in clang, so I'm still investigating that.)

PR #376, opened 2021-02-10
Import minor fixes

gperciva:


gperciva:
I discovered these while importing libcperciva into other projects.

PR #377, opened 2021-02-13
sha256_sse2: MSG4 with rotating message schedule state

gperciva:


gperciva:
The main reason I want to make this change is that is simplifies the code (in MSG4).  The optimization is just a nice side effect.

compilers

alg/sha256_sse2.c
```diff
@@ -189,17 +185,18 @@ SHA256_Transform_sse2(uint32_t state[static restrict 8],
...
-	_mm_storeu_si128((__m128i *)&W[0],
-	    _mm_bswap_epi32(_mm_loadu_si128((const __m128i *)&block[0])));
-	_mm_storeu_si128((__m128i *)&W[4],
-	    _mm_bswap_epi32(_mm_loadu_si128((const __m128i *)&block[16])));
-	_mm_storeu_si128((__m128i *)&W[8],
-	    _mm_bswap_epi32(_mm_loadu_si128((const __m128i *)&block[32])));
-	_mm_storeu_si128((__m128i *)&W[12],
-	    _mm_bswap_epi32(_mm_loadu_si128((const __m128i *)&block[48])));
+	Y[0] = _mm_bswap_epi32(_mm_loadu_si128((const __m128i *)&block[0]));
+	_mm_storeu_si128((__m128i *)&W[0], Y[0]);
```
gperciva, line 193: Admittedly, although MSG4 looks nicer, this code is a bit ugly.  Unfortunately, we face a slight performance penalty (-0.5% to -1%) if we don't store each `Y` value as soon as we calculate it.  :(

An alternate idea would be to add a function like this (untested):
```
static void
load_block(W, i, block)
{
    _mm128i a;

    a = _mm_loadu_si128((const __m128i *)&block[16*i]);
    a = _mm_bswap_epi32(a);
    _mm_storeu_si128((__m128i *)&W[4*i], a);
    return (a)
}
```

but that felt like it was being over-engineered.

alg/sha256_sse2.c
```diff
@@ -225,10 +222,14 @@ SHA256_Transform_sse2(uint32_t state[static restrict 8],
 
 		if (i == 48)
 			break;
-		MSG4(W, 0, i);
-		MSG4(W, 4, i);
-		MSG4(W, 8, i);
-		MSG4(W, 12, i);
+		Y[0] = MSG4(Y[0], Y[1], Y[2], Y[3]);
+		_mm_storeu_si128((__m128i *)&W[16 + i + 0], Y[0]);
```
gperciva, line 226: Again, for performance reasons I'm interleaving MSG4 with _mm_storeu_si128().

We could avoid this by passing `(W, 0, i)` to `MSG4`, but IMO that looks uglier than this.  Another idea I've tried is to pass `(&W[16 + i + 0])` to `MSG4`... but again, that spoiled the "niceness" of MSG4 simply taking X0, X1, X2, X3.

PR #378, opened 2021-02-17
SHA256 Arm boilerplate

gperciva:

PR #379, opened 2021-02-17
Sha256 arm

gperciva:


gperciva:
This builds on #378 by adding 3 commits, which switch to using Arm SHA256 intrinsics.  We get a 5x - 6x speed increase.

(Recall that `clang -O3 -march=native` doesn't work on linux t4g, so it's not present in the plot.)

compilers

gperciva:
I think this is ready now.

Arguably it's missing a `make Makefiles` commit, which would remove `"sysendian.h" from a few `Makefile`s.  I don't think we need to include that, particularly as there's crypto_aes arm boilerplate coming up very soon.

alg/sha256_arm.c
```diff
@@ -91,14 +72,14 @@ SHA256_Transform_arm(uint32_t state[static restrict 8],
 	uint32x4_t S0_orig, S1_orig;
 	int i;
 
+	(void)W; /* UNUSED */
```
cperciva, line 75: Now that these are unused, is there any reason not to drop them like we do for the x86 shani version?

alg/sha256_arm.c
```diff
@@ -98,8 +87,12 @@ SHA256_Transform_arm(uint32_t state[static restrict 8],
 #endif
 {
 	uint32x4_t Y[4];
+	uint32x4_t S0, S1; /* S0 and S1 alternate between ABEF and CDGH. */
```
cperciva, line 71: Have you benchmarked this vs. making it a two-element array?

alg/sha256_arm.c
gperciva: I considered it, but slightly erred on the side of simplicity; particularly because it would require a `sha256_Transform_arm_with_W_S` shim function in `sha256.c`.  But it was only a 55% to 45% "erring", so I'm quite content to add that.

alg/sha256_arm.c
```diff
@@ -98,8 +87,12 @@ SHA256_Transform_arm(uint32_t state[static restrict 8],
...
+	uint32x4_t S0_orig, S1_orig;
```
cperciva, line 72: I would be inclined to make this `uint32x4_t _state[2]`.  I often use `_` prefixes for "this is the same thing as the function parameter but reinterpreted`.

alg/sha256_arm.c
gperciva: No... I can do, but I wouldn't expect that to make a difference...?

cperciva: I've seen that sort of thing confuse optimizers in the past.  If the performance is the same, I would prefer to have it as an array.

gperciva: Done, although of course we can't make that change until we're no longer passes `S` as an argument.

alg/sha256_arm.c
gperciva: done

gperciva:
Performance still looks good.  A noticeable improvement by not passing `W` and `S` with the smaller buffer sizes.   (That's the black bars.)

compilers

alg/sha256_arm.c
gperciva: To clarify the history: I rebased this change into the "use intrinsics for round calculation".  The other improvements had to be in their own commits.

alg/sha256_arm.c
```diff
@@ -108,41 +101,30 @@ SHA256_Transform_arm(uint32_t state[static restrict 8],
 	Y[3] = vld1q_u32(&W[12]);
 
 	/* 2. Initialize working variables. */
-	memcpy(S, state, 32);
+	S0 = vld1q_u32(&state[0]);
```
cperciva: I would write this

```
S[0] = _state[0] = vld1q_u32(&state[0]);
S[1] = _state[1] = vld1q_u32(&state[4]);
```

unless there's a good reason not to.

alg/sha256_arm.c
```diff
@@ -98,9 +98,9 @@ SHA256_Transform_arm(uint32_t state[static restrict 8],
 	}
 
 	/* 4. Mix local working variables into global state. */
-	S0 = vaddq_u32(_state[0], S0);
-	S1 = vaddq_u32(_state[1], S1);
-	vst1q_u32(&state[0], S0);
-	vst1q_u32(&state[4], S1);
+	S[0] = vaddq_u32(_state[0], S[0]);
```
cperciva, line 101: And since we don't particularly care about `S` after this, I would `vst1q_u32(&state[0], vaddq_u32(_state[0], S[0]));` rather than storing the value in `S` and then copying it out.

alg/sha256_arm.c
gperciva: done

alg/sha256_arm.c
gperciva: done

alg/sha256_arm.c
```diff
@@ -87,64 +58,45 @@ static const uint32_t Krnd[64] = {
...
 #else
 void
 SHA256_Transform_arm(uint32_t state[static restrict 8],
-    const uint8_t block[static restrict 64], uint32_t W[static restrict 64],
-    uint32_t S[static restrict 8])
+    const uint8_t block[static restrict 64])
 #endif
 {
+	uint32x4_t Y[4];
+	uint32x4_t S[2]; /* S[0] and S[1] alternate between ABEF and CDGH. */
```
cperciva, line 69: I'm confused by the comment here.  It looks like we're loading ABCD into one and EFGH into the other; is there some permutation I'm missing?

gperciva: I'm confused too, but I'm now fairly confident that it's because I misunderstood the "rotating state" in our generic SHA256 code when I was first working on this.

Looking at our first test case, the SHA256 of "", with our generic `sha256.c` (after removing the "if hwaccel, do this other function") code gives:
```
after 4 rounds
3037e1b5 6fdb22fa 2001b256 18c7e2a2 4c23eacb c4ac5526 9ce564d0 7c08884d 
after 8 rounds
7ad5b1b6 f0205590 292e3b42 4366f730 53f94dcb 82ede06f 4f6540fa e7d0641e
```
That's from printing `S[j]` from 0 to 7.

The Arm code, with slightly different formatting (sorry, could change if you want), gives:
```
after 4 rounds
S[0] 4c23eacb c4ac5526 9ce564d0 7c08884d
S[1] 3037e1b5 6fdb22fa 2001b256 18c7e2a2
after 8 rounds
S[0] 7ad5b1b6 f0205590 292e3b42 4366f730
S[1] 53f94dcb 82ede06f 4f6540fa e7d0641e
```

In other words, if we interpret the generic S[8] as ABCDEFGH, and the Arm S[2] as ABCD, EFGH; then the two implementations agree after 8 rounds.  But after 4 rounds, the Arm implementation is flipped, with S[0] containing what we'd expect to see in S[1] and vice versa.

**However**, future reading about the Arm instructions, and looking at our generic code, makes me think that I got the "alternating" thing the wrong way around.  Namely, it looks like the Arm code always stores ABCD in S[0]; but the generic code rotates the state.  So after 4 iterations, the generic code has S[0] = EFGH; after 8, it's back to S[0] = ABCD.

In my defense, our SHANI implemention does have the state flipping between S[0] and S[1].  Namely:
```
        /* Perform two rounds of SHA256, using the low two words in M. */       \
        S[1] = _mm_sha256rnds2_epu32(S[1], S[0], M);                            \
                                                                                \
        /* Shift the two words of M down and perform the next two rounds. */    \
        M = _mm_srli_si128(M, 8);                                               \
        S[0] = _mm_sha256rnds2_epu32(S[0], S[1], M);      
```

In summary, I think that the solution is to change those comments in `sha256_arm.c` to clarify that `S` doesn't have rotating state in the C code; it's all taken care of by the intrinsics.

cperciva:
Please rebase.

gperciva:
Updated, potentially ready for merge.

PR #380, opened 2021-02-19
sha256_sse2: fix API comment for SPAN_ONE_THREE

gperciva:


gperciva:
Oops.  No harm done; it was only the comment that was screwed up.

alg/sha256_sse2.c
```diff
@@ -119,13 +119,13 @@ s1_128_low(__m128i a)
 
 /**
  * SPAN_ONE_THREE(a, b):
```
gperciva, line 121: Optionally, we could also rename this to `SPAN_THREE_ONE`.

PR #381, opened 2021-02-21
Workaround for missing _mm_loadu_si64 on old compilers

gperciva:


gperciva:
Fixes https://github.com/Tarsnap/tarsnap/issues/473.

crypto/crypto_aesctr_aesni.c
```diff
@@ -60,7 +64,11 @@ crypto_aesctr_aesni_stream_wholeblocks(struct crypto_aesctr * stream,
 		be64enc(block_counter_be_arr, block_counter);
 
 		/* Encrypt the cipherblock. */
+#ifdef NO_LOADU_SI64
+		bufsse = _mm_loadu_si128((const __m128i *)block_counter_be_arr);
```
cperciva, line 68: This performs a 128-bit load from an 8-byte array, which seems a bit optimistic.

cperciva:
I'm not entirely certain this is worth complicating `crypto_aesctr_aesni` for.  We could fix the build breakage by using `_mm_loadu_si64` in the compiler support detection -- people with broken compilers then wouldn't get the optimized version, but they're not going to get a fully optimized version anyway and they should upgrade to a working compiler.

crypto/crypto_aesctr_aesni.c
gperciva: ah right, I was only thinking about stream->pblk.

gperciva:
> I'm not entirely certain this is worth complicating crypto_aesctr_aesni for.

I agree that the current PR is flawed, but I think we should have something.  The previous releases of spiped and tarsnap had working AESNI.

Now, one possible fix would be to still use `crypto_aes_aesni.c` but not `crypto_aesctr_aesni.c` (which is essentially what the previous releases had).  But in some ways that ends up being a more complicated solution, because in that case we'd need cpusupport to detect 1) can we use AESNI, and 2) does the compiler recognize `_mm_loadu_si64`.  (Which brings us one step closer to the dreaded `./configure`.)

gperciva:
Better fix -- we can use `_mm_load_sd()` to do exactly the same thing as `_mm_loadu_si64()`, albeit with a cast.  I've hidden this inside a `load_si64()` shim.  The compiler ends up using `movq` in both cases.

This allows us to (mostly) keep the recent optimizations.  (gcc7 is "only" 1100 MB/s, rather than 1600)

### Compare versions of gcc

compilers

gperciva:
Done.

cperciva:
Commit message needs updating for the changed macro name.

gperciva:
Oops, fixed.

PR #382, opened 2021-03-03
sha256.c: make 'hwgood' a global static variable

gperciva:


gperciva:
I have a series of minor improvements to the cpusupport framework.  Most of them are purely cosmetic or hitting compiler edge cases (such as changing a `while` to an `if`), and will come later.  I've gone back and forth many times on what order to present the changes.

This is the most significant change, so I figured I'd get it out of the way first.

On amd64 with SSE2, it gives a noticeable bonus to `clang -O2`, and a tiny bonus to others.
compilers

On ARM, it gives a tiny bonus to clang.  gcc is a pain; all sorts of changes flip the performance of `gcc -O2` and `gcc -O3`.  This particular commit boosts `-O3` and penalizes `-O2`, but that'll flip again in the next PR or two.  The end result should be positive in both cases, though.
(That said, I'm not too fussed about compiler performance on ARM, since it's a relatively new platform and there's lots of room for improvement in later compiler versions.)
compilers

gperciva:
Also, to anticipate the question: the idea is to make all the changes to sha256 first, then apply the updated framework to crypto_aes, crypto_aesctr, and crc32c.

cperciva:
Hmm.  Really what you're saying here is "there's a performance cost to checking if we've initialized a variable every time we run `SHA256_Transform`; it's faster to perform the initialization in `SHA256_Init` and rely on the fact that it has been done later".

I guess that's fine, and very reasonable in fact since `SHA256_Init` will (presumably!) be called far fewer times than `SHA256_Transform` -- but in that case let's be clear about it and e.g. change `usehw()` to `hwgood_init()`.

gperciva:
Right, I had something very similar waiting in the next PR, so I've adapted it and included it here.  This is a case of agreeing on the end result; the only issue is how we get there.

Here's the timing results on amd64; note in particular the increase in 256 bytes for `clang -O2` and `gcc -O2`.

compilers

As for "how we get there", I'm of two minds.
- the current PR has many commits, with an almost-minimal set of changes per commit.  (The "rename" commit could theoretically be split into 3 separate ones, but that would be silly).  As a matter of curiosity, it's interesting to see the speed boost from the global static variable, and removing the `while`.  But I'm not certain that we need to keep that level of detail in the final git tree.
- another option would be smoosh everything together.  That would make it trickier to investigate if something went wrong, particularly because removing the `while` changes the indentation of a bunch of lines.  But since this is only in "choosing which acceleration to use" code, it's pretty obvious if something breaks.

I'm leaning towards smooshing these commits together, but I figured that I should keep them separate for now, since it's trivial to smoosh but a pain to un-smoosh.

alg/sha256.c
```diff
@@ -17,13 +17,13 @@
     defined(CPUSUPPORT_ARM_SHA256)
 #define HWACCEL
 
-enum {
+static enum {
```
cperciva, line 20: If you add appropriate `#ifdef`s here, do you remove the need for `-Wno-switch-enum`?

gperciva: Technically yes we would avoid that need, but then we'd get warnings about unused macros (due to other platforms).  Hence 37237daa3cf7deaead9d1746b85397e5c7e049d6 a month ago.

What we _could_ do is:
```
#define HW_UNSET 0
#define HW_SOFTWARE 1
#define HW_X86_SHANI 2
#endif
#if defined(CPUSUPPORT_X86_SSE2)
#define HW_X86_SSE2 3
#endif
#if defined(CPUSUPPORT_ARM_SHA256)
#define HW_ARM_SHA256 4
#endif
```

... but that looks so ugly that I didn't bother proposing it.

On the other hand, sha256 is the worst case (in terms of number of platforms), and this code only needs to be written once.  So I'm happy to rewrite the PR to use this, if you want.

cperciva: Did you mean 2d664b8b?

cperciva: I was thinking we could have

```
static enum {
    HW_SOFTWARE,
    HW_X86_SHANI,
#endif
#if defined(CPUSUPPORT_X86_SSE2)
    HW_X86_SSE2,
#endif
#if defined(CPUSUPPORT_ARM_SHA256)
    HW_ARM_SHA256,
#endif
    HW_UNSET
} hwgood;
```

gperciva: Whoops, yeah, that's the commit I meant.

I could have sworn that it didn't work, but as long as we don't have a `default` inside the switch, it looks ok.

cperciva:
Looks good, rebase?

gperciva:
All in one commit, or with multiple commits?

... come to think of it, I'm going to be applying this to crc32, crypto_aes, and crypto_aesctr.  Should I do that all at once right now?

cperciva:
Let's say one commit for each of them?

cperciva:
(But all in one PR.)

cperciva:
Oh, and maybe edit STYLE somewhat since the code for deciding that we should use special intrinsics doesn't look quite like it says there any more.

gperciva:
STYLE is still accurate for things like `crypto_entropy.c`.  Did you want me to add something like
> If there are many possible CPU-feature-exploiting variants (e.g., SHA256 via 
x86 SHANI, x86 SSE2, ARM SHA2), then a switch() may be used instead of the `if()`.

?  Or did you want me to discuss the whole hwaccel framework as an optional format, either in terms of:

> If there are many possible CPU-feature-exploiting variants (e.g., SHA256 via 
x86 SHANI, x86 SSE2, ARM SHA2), then you may define a HWACCEL preprocessor variable, store the best intrinsic method in a `hwaccel` static variable via an `hwaccel_init()` function, then use:
```
#ifdef HWACCEL
        switch(hwaccel) {
#if defined(CPUSUPPORT_FOO_BAR)
        case HW_FOO_BAR:
                /* Call variant code. */
                return;
#endif
...
        case HW_SOFTWARE:
               break;
        case HW_UNSET:
                /* UNREACHABLE */
                warn0("Programmer error: unreachable hwaccel value.");
                assert(0);
        }

        /* Normal implementation of the function. */
#endif
```

gperciva:
Update: let's not bail if we hit HW_UNSET.

We already `assert()` that it's not HW_UNSET in SHA256_Init(), so it would only be unset if there's solar radition bit-flipping memory, or if we refactor sha256.c and mess it up.  But in the latter case, the speed drop from using software should be noticeable.

Motivation: the nitpicky clang compiler fix made us hit another edge case in the compilation of the `switch` and lose performance; this speeds us up.  In most cases, to more than we had previously.

- the green bars are the final performance figures that you've seen before.
- the red bars are the effect of the nitpicky compiler fix.  Note that it drops performance for `gcc -O2`
- the purple bars are from not bailing with HW_UNSET.
 
compilers

cperciva:
> STYLE is still accurate for things like `crypto_entropy.c`. Did you want me to add something like

```
Functions for which special CPU-feature-exploiting variants exist should
take the form
        {
                /* Variable declarations here. */

                /* Asserts here, if any. */

        #ifdef CPUSUPPORT_FOO_BAR
                if (/* We've decided we can use the variant code */) {
                        /* Call variant code and return. */
                }
        #endif

                /* Normal implementation of the function. */
        }
```

It's just that `if` in the middle which bugs me, since we're using a `switch` instead now (for some but not all such functions).  Maybe just a comment afterwards saying "if there are multiple variants, the 'if' could instead be a 'switch' which invokes the appropriate variant function`?

gperciva:
Urgh, this change knocks 10% off gcc's performance on arm.

As you can see below, it's normal for commits to give gcc a +10% or -10% for `gcc -O2` or `gcc -O3`, but this branch currently ends up with -10% in both cases.  :(

That's why I was initially planning on doing this step by step, since there's so many weird quirks.

compilers

gperciva:
> Maybe just a comment afterwards saying "if there are multiple variants, the 'if' could instead be a 'switch' which invokes the appropriate variant function`?

Sure.  I'll do that in a separate PR since this one is cluttered enough already.

gperciva:
Ok, new idea:
- we have an "ideal" format, which happens to work on amd64, which is the PR you saw earlier plus two REBASE commits.
- we add one unfortunate assert, hidden within `#if defined()`, for gcc on ARM.

I normally wouldn't bother suggesting the second point, since I'd figure that you'd never go for it.  But sha256 is a significant part of spiped and tarsnap client speed, and 63bd446 is nicely contained.

### amd64

compilers

### arm 64

compilers

alg/sha256.c
```diff
@@ -267,7 +264,19 @@ SHA256_Transform(uint32_t state[static restrict 8],
...
+
+#if defined(__GNUC__) && defined(__aarch64__)
+	/*
+	 * Optimization for gcc on 64-bit ARM.  This _should_ have no effect,
+	 * because the only place hwaccel is written is inside hwaccel_init(),
+	 * and we have an `assert(hwaccel != HW_UNSET)` immediately after
+	 * calling hwaccel_init() in SHA256_Init().  But evidently that info
+	 * isn't being carried through between functions.
+	 */
+	assert(hwaccel != HW_UNSET);
```
cperciva, line 276: There's no way for gcc, or any other compiler, to know that `hwaccel` can't be `HW_UNSET` at this point, because there's no way for them to know that we will always call `SHA256_Init` before calling `SHA256_Transform`.

Just how much of a performance benefit does this provide?

gperciva: It's shown graphically in the image for "arm 64" (in the previous comment), comparing the purple to black bars.

- for a buffer size of 256, `gcc -O2` goes from 560 to 738.  So a 31.7% boost.
- 256 with `gcc -O3` is 563 to 705.  So only a 25% boost.
- buffer size of 1024 is a more modest boost
- buffer size of 10000 is barely noticeable; 1123 to 1137.  1% boost?

It's worth glancing at the graph, just to see how starkly  `gcc` alternates between ~560 to ~720.

cperciva: Ah, got it.  It's not easy (for me at least) to see what's going on when there's three compilers squished together within a single block size on the graph.

gperciva: Yeah, I've tried a bunch of different visualizations, but nothing's jumped out as ideal.

I could try having two levels of bars?  i.e. a very light grey background for each block size, then within that background, I could have medium-light grey bars for each compiler (within which we'd have the colored boxplots)?

cperciva: Ok, I can live with having this assertion here for gcc/arm64.  I'm guessing it hurts performance on other systems and that's why it's a preprocessor conditional?

But the comment needs to be changed to reflect the reality of "we require that SHA256_Init is called before SHA256_Transform, but the compiler has no way of knowing that".

gperciva: It's a minor performance penalty elsewhere (amd64 with `clang -O2` and buffer size of  10000, it's 198 to 197 MB/s; buffer size of 256, it's 178 to 176 MB/s).

More to the point, though, it's a messy hack that's only necessary due to something weird that `gcc` on ARM is doing with the `switch`.  Remember, the `switch` currently contains:
```
case HW_SOFTWARE:
case HW_UNSET:
        break;
```

So why on earth would `assert(hwaccel != HW_UNSET)` have a 31% speed boost?

I've compared the assembly code from both versions, and nothing jumped out at me.  Some code shuffled around, and it's storing values in different registers, so there's a ton of changed lines simply due to those trivial changes.  Also, this gets compiled to multiple locations;
```
$ grep "assert(hwaccel" assembly.difff | wc -l
16
```

... in fact, that gives a clue about the major difference:
```
td@ip-xxx:~/src/libcperciva/tests/sha256$ grep "assert(hwaccel" orig.txt | wc -l
1
td@ip-xxx:~/src/libcperciva/tests/sha256$ grep "assert(hwaccel" new.txt | wc -l
16
```

So evidently the original code is just on the boundary of having certain parts inlined?

If this was amd64, I'd keep on digging.  Or if I saw this behaviour from other compilers.  But at this point, my instinct is that it's not worth investigating further.  There's some quirky behaviour of `gcc` on ARM; we can manage the quirk by adding an `assert`, so good enough for now.

(Of course, if you want me to spend more time on it, I can do that.  gcc has 264 current open bugs which contain "arm"; I could check if any of those are applicable, and if not, try to create a small example of this behaviour.)

cperciva: No need to investigate further.  Just fix the comment.

gperciva:
Update with rebase, and adjusting the comment.  If this looks plausible, I'll apply the same "new format" to other cpusupport files.

IIRC, my previous experiments suggested that other cpusupport files don't need the same `assert` for gcc + arm, so of course I'll leave that out.

(I realized last night that my comment about the number of instances of `assert(hwassert` was completely bogus.  But let's leave that aside.)

### amd64

compilers

### aarch64

compilers

alg/sha256.c
```diff
@@ -391,6 +396,14 @@ SHA256_Init(SHA256_CTX * ctx)
 
 	/* Initialize state. */
 	memcpy(ctx->state, initial_state, sizeof(initial_state));
+
+	/* Choose hardware acceleration (if any). */
+	if (hwaccel == HW_UNSET) {
```
cperciva, line 401: Is this faster than placing

```
if (hwaccel != HW_UNSET)
    return;
```

into the top of `hwaccel_init`?  I think it's cleaner if we the `hwaccel_init` and `assert` are unconditional here, but not so much cleaner to be worth paying a performance penalty.

gperciva: Doesn't make any difference.  We only call SHA256_Init once in the libcperciva test suite and the spiped proto.  It's called a few times in scrypt's `scryptenc_setup` for the header checksum and signature, and in `scryptdec_setup` to verify the same -- but again, that's a function that's only run once.

tarsnap's `crypto_hash_data_key_2` and `crypto_hash_data_2` does call _Init every time, but I haven't seen those (as a time blocker) in any profiling.

gperciva:
Added the requested change to SHA256_Init.

I also added an experimental `CPUSUPPORT_CHECK_USE` macro.  You might not like it since it's just boilerplate.  But take a look at `hwaccel_init`!  See how nice that looks, and how easy it would be to add support for another cpusupport platform?

alg/sha256.c
```diff
@@ -397,13 +397,17 @@ SHA256_Init(SHA256_CTX * ctx)
 	/* Initialize state. */
 	memcpy(ctx->state, initial_state, sizeof(initial_state));
 
-	/* Choose hardware acceleration (if any). */
-	if (hwaccel == HW_UNSET) {
-		hwaccel_init();
+#ifdef HWACCEL
+	/* We're finished if we've set hwaccel. */
+	if (hwaccel != HW_UNSET)
```
cperciva, line 371: Sorry, I wasn't clear.  I meant putting this bit into `hwaccel_init`.  So `SHA256_Init` always calls `hwaccel_init` but `hwaccel_init` does nothing if it has already initialized.

cpusupport/cpusupport.h
```diff
@@ -94,6 +94,25 @@
...
 
+/**
+ * CPUSUPPORT_CHECK_USE(success_value, cpusupport_checks, check):
+ * Check if we can enable ${success_value}, given the ${cpusupport_checks} and
+ * ${check}.  If the ${cpusupport_checks} pass but the ${check} is non-zero,
+ * produce a warning which includes a stringified ${success_value}, then
+ * fallthrough.  This macro requires an 'hwaccel' variable.
+ */
+#define CPUSUPPORT_CHECK_USE(success_value, cpusupport_checks, check) do { \
+		if ((cpusupport_checks)) {				\
```
cperciva, line 105: There seems to be an excess level of indentation here?

cpusupport/cpusupport.h
```diff
@@ -94,6 +94,25 @@
...
+			if ((check) == 0) {				\
+				hwaccel = (success_value);		\
```
cperciva, line 107: I would make `hwaccel` a macro parameter.  I know we're always using the same variable name, but it's ugly to have implied variables in a macro.

cperciva:
I like the concept behind `CPUSUPPORT_CHECK_USE`.  Maybe `CPUSUPPORT_TEST` would make more sense as a name though?

gperciva:
Interesting, I originally named it `CPUSUPPORT_TEST`, but shifted away from that:
- does "test" mean "check" or "experimental"?  I think it's more clear if this is a macro that performs a `CHECK`.
- is the macro checking whether we have access to the raw intrinsics -- in which case, what's the difference between that and `cpusupport_foo_bar()`?  I tried to emphasize that this macro checks that it's ok to `USE` those intrinsics.

I'm open to other names, or to `CPUSUPPORT_TEST` if you really prefer that one.

cperciva:
I was thinking that "test" captured the fact that we were actually testing whether it worked rather than verifying whether it *should* work.  But we could make it clearer with `CPUSUPPORT_VALIDATE`?

gperciva:
Ok, I'll go with `CPUSUPPORT_VALIDATE`.

gperciva:
I think this is ready now.

I've marked this as a draft, because if it looks good, I'll upload a `cpusupport-standardize` branch, which uses this format for crc32, crypto_aes, and crypto_aesctr.

cpusupport/cpusupport.h
```diff
@@ -94,6 +94,26 @@
...
 
+/**
+ * CPUSUPPORT_VALIDATE(hwvar, success_value, cpusupport_checks, check):
+ * Check if we can enable ${success_value}, given the ${cpusupport_checks} and
+ * ${check}; if so, write to ${hwvar}.  If the ${cpusupport_checks} pass but
+ * the ${check} is non-zero, produce a warning which includes a stringified
+ * ${success_value}, then fallthrough.
+ */
+#define CPUSUPPORT_VALIDATE(hwvar, success_value, cpusupport_checks,	\
+    check) do {								\
```
cperciva, line 105: Is the misaligned `\` a github display issue?

cpusupport/cpusupport.h
```diff
@@ -94,6 +94,26 @@
 	int								\
 	cpusupport_ ## arch ## _ ## feature ## _detect_1(void)
 
+/**
+ * CPUSUPPORT_VALIDATE(hwvar, success_value, cpusupport_checks, check):
```
cperciva, line 98: Commit message needs to be adjusted to reflect this name.

cpusupport/cpusupport.h
gperciva: Yes.  :(

PR #384, opened 2021-03-13
tests: rewrite print_hardware() due to nitpicks

gperciva:


tests/sha256/main.c
```diff
@@ -19,35 +19,37 @@ static const size_t num_perf = sizeof(perfsizes) / sizeof(perfsizes[0]);
...
 {
 	int use_hardware = 0;
 
+	/* Inform the user of the general topic... */
+	printf("%s", str);
+
+	/* ... and whether we're using hardware acceleration... */
 #if defined(CPUSUPPORT_X86_SHANI) && defined(CPUSUPPORT_X86_SSSE3)
-	if (cpusupport_x86_shani() && cpusupport_x86_ssse3())
+	if (cpusupport_x86_shani() && cpusupport_x86_ssse3()) {
```
cperciva, line 33: This feels unnecessarily complicated.  Why not

```
	if (cpusupport_x86_shani() && cpusupport_x86_ssse3()) {
                printf(" using hardware SHANI.\n");
        } else
#endif
#if defined(CPUSUPPORT_X86_SSE2)
	if (cpusupport_x86_sse2()) {
                printf(" using hardware SSE2.\n");
       } else
#endif
...
```

cperciva: I don't think there's any need for `use_hardware` at all in here.

gperciva:
I suppose that my uncomfortableness with `if()` statements spanning preprocessor defines in a weakness.

Updated accordingly.

cperciva:
Normaly I would avoid having an `if` which spans preprocessor conditionals, but in this case it makes everything else so much cleaner I think it's worth it. :-)

PR #389, opened 2021-03-24
build: always expand variables before printing

gperciva:


gperciva:
This was part of the `build-shared` in https://github.com/Tarsnap/spiped/pull/308, but got lost when we switched to the `liball.a` approach.

There's no change to `make Makefiles` on FreeBSD, but this commit is necessary for Linux to produce the same Makefiles.  (Otherwise, `LIBALL=../../liball/liball.a` gets changed to `LIBALL=${SUBDIR_DEPTH}/liball/liball.a`)

PR #388, opened 2021-03-23
Detect being linked to pthread

gperciva:

PR #392, opened 2021-03-28
noeintr_close.c with libpthread_stubs

gperciva:


gperciva:
This is a version of `noeintr_close.c` which can be compiled into `liball.a`.   This results in `liball.a` having the unresolved symbols of `pthread_mutex_lock()` and `pthread_mutex_unlock()`.

For every binary in our git tree, the build system ensures that:
- if `LDADD_REQ` includes `-lpthread`, then there's no special action.  (The symbols are in that library.)
- if `LDADD_REQ` does not include `-lpthread`, then it adds `libpthread_stubs.a` to the command-line.  This allows the linker to resolve `pthread_mutex_lock()` and `pthread_mutex_unlock()` to our custom (no-op) functions.

In https://github.com/Tarsnap/libcperciva/pull/312#issuecomment-805092349, you said that this wouldn't work if `libpthread` was a dynamic library, but so far I haven't seen evidence of it not working.

You also mused about having runtime detection of whether we're linked to `pthread`.  I think that this method is superior, since we still need to link the binary before we can mess around with runtime detection, and using stubs side-steps the question of detection.  But if runtime detection is still needed, I think the sketch in https://github.com/Tarsnap/libcperciva/pull/388 works.

cperciva:
I like this more than I thought I would.  In particular, the "build logic ensures we have either libpthread_stubs OR -lpthread" is nice.  My comment in #312 about problems if libpthread is dynamic was that if we link to the stubs *and* pthread then we can't guarantee which version we get; but if the build logic makes sure we have one or the other I think we're good.

I probably won't be able to review this properly for at least a couple weeks but don't take my silence as disapproval.

gperciva:
I think that technically this is a POSIX violation, unfortunately.

The c99 spec 7.1.3 Reserved identifiers says
> All identifiers with external linkage (including future library directions) and errno are always reserved for use as identifiers with external linkage.

and POSIX reserves the prefix `pthread_`.   A non-authoritative source [1] specifically gives the example of a custom `malloc()` being non-compliant.
[1] https://wiki.sei.cmu.edu/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier

I think we can still get `noeintr_close.c` to work, by adding a `maybe_lock.c` which calls pthread functions within an `#ifdef USE_PTHREAD`.  That file couldn't be part of `liball.a`; any binary which used `noeintr_close.c` would need to compile it.  The build logic would automatically set or not-set USE_PTHREAD.  I'll try implementing it tomorrow.  It's not as cute of a solution, but it's probably more straightforward.

PR #394, opened 2021-03-30
crypto_aes_arm: boilerplate

gperciva:


gperciva:
This is the first of a series of PRs for AES intrinsics support for `crypto_aes`.  I'm confident [1] about the end-point; the quibble is "how exactly do we want the git history to look".

[1] That's not quite true; there's some quibbles about the "native-ish keygen", which are the final 3 commits.  But I'll comment about that on that PR.

As for the git history: the set of PRs here have these changes split up almost as much as possible, since it's easier to merge things upon request, rather than splitting commits up.
- #394 (this one): add the basic cpusupport infrastructure for ARM-AES, and have the execution go to `crypto_aes_arm.c`.  This PR postpones all arm intrinsics; `crypto_aes_arm.c` actually just uses the OpenSSL functions.  (I suspect that you won't want to merge this PR by itself.)
- #395: this switches all the code in `crypto_aes_arm.c` over to the x86 intrinsics, then adds an "emulation layer" of shims (via `#define`s).  The end result of this PR is that we're using ARM intrinsics (with a slight slowdown from some extra `XOR(a, 0)` fluff), without any apparent modification to the x86 code.
- #396: this removes the shims for the AES round compuation, using native ARM intrinsics.
- #397: this removes the shims for the AES key schedule.  There's no native ARM intrinsic for this step, but we can use the AES round computation intrinsic so that we don't need to implement SubBytes.  (This PR has the most quibble room; this isn't a performance-critical part, so it's more a question of "what's the clearest code?".)

gperciva:
I missed two additions to this set of PRs:
- generalizing `crypto_aes_use_x86_aesni()` to `crypto_aes_can_use_intrinsics()`.  I previously thought that I could delay that until AES-CTR, but it'll be useful to do now.
- making the perftest use `crypto_aes_encrypt_block_arm_u8()` directly.  That bumps the final speed from 1100 to 1280 MB/s.

I'll hold off on revising all these PRs until we've discussed the overall strategy for these changes.

gperciva:
This PR doesn't depend on #398 (crypto_aes.c: use test vectors instead of openssl), but if you like that one, it would make sense to merge it first, then rebase this on top of that.

cperciva:
> I'll hold off on revising all these PRs until we've discussed the overall strategy for these changes.

Can you clarify what I should be looking at now?  Presumably I shouldn't merge anything if you're planning on making revisions?

gperciva:
Are you ok with the general plan?  In particular, "add openssl to _arm.c; add shims for x86 and remove openssl from _arm.c; remove shims for x86 in _arm.c".

It might be the case that you would prefer to have a single commit for "Add crypto_aes_arm.c" which uses the native ARM intrinsics from the beginning, and doesn't have any openssl or shims.

gperciva:
To clarify that point: if this was non-cryptography stuff, I'd just submit a PR with the final version of `_arm.c`.  But since this *is* crypto, I figured that an excess of caution would be appropriate.

cperciva:
Got it, sorry about the delay.  IIRC in another case you did "copy x86 optimized version, then make changes to turn it into arm optimized version"; would that make sense here?

I'm not all that interested in a detour through OpenSSL; seeing the change from the known-good code using x86 intrinsics to code using arm intrinsics seems like it might be helpful though -- depending on whether the code structure ends up being equivalent.

gperciva:
Sure, I can omit the detour through OpenSSL.  It does mean that the "minimum viable PR" is a bit bigger, namely:
- add cpusupport stuff for ARM-AES
- copy `crypto_aes_aesni.c`
- do trivial replacements
- add shims

cperciva:
That sounds fine.

PR #395, opened 2021-03-30
crypto_aes_arm: shims

gperciva:


gperciva:
Performance: 160 to 980 or 710 MB/s.

compilers

PR #396, opened 2021-03-30
crypto_aes_arm: native round calculation

gperciva:


gperciva:
Performance of these improvements, compared to #395: go from 980 or 710, to 1100 MB/s across the board.

compilers

gperciva:
Performance isn't exciting.  tl;dr: no change until we start using native AES rounds; afterwards, all 3 working [1] compiler-flag combos produce much faster code.

[1] remember that `clang-11 -march=native` doesn't work on EC2 t4g.


compilers

PR #397, opened 2021-03-30
crypto_aes_arm: native-ish keygen calculation

gperciva:


gperciva:
I can't see any statistically significant changes in performance compared to #396 -- which is expected, since the key schedule is a miniscule part of the perftest.

gperciva:
This has absolutely no performance difference (as measured by `test_crypto_aes`, which admittedly is not designed to time the AES key schedule, since that's not important).

This PR is more a question of personal choice:
- do we want to keep the existing code, which mimics `_mm_aeskeygenassist_si128()`, but has the hard-coded "Undo ShiftRows" table which is somewhat opaque.
- or have a easier-to-understand functions, at the mental cost of being different than the AESNI implementation.

As implied above, I can see good arguments for both cases.

PR #398, opened 2021-04-10
crypto_aes.c: use test vectors instead of openssl

gperciva:


gperciva:
This slightly lessens our dependency on openssl -- if a system has AESNI (or ARM-AES), why bother calling AES_encrypt(), rather than hardcoding the known good result?

PR #399, opened 2021-04-10
AES disable openssl

gperciva:


gperciva:
This builds on #398 by supporting a -DDISABLE_OPENSSL for `crypto_aes.c`.  If you're compiling this on a system which you know has AESNI or ARM-AES, it's possible to omit `-lcrypto` entirely.  (Naturally, this isn't suitable for building packages for general-purpose systems, since not all CPUs have AES hardware intrinsics.)

I figure that there's only a 30% chance that you'll want to merge this, but I was confident that you'll be curious to skim the code.  If you *do* like it, then yes, it can be imported to scrypt, which can then add `./configure --without-openssl`.  (We can't do the same for spiped and tarsnap, because they use `crypto_dh.c` which requires `openssl/bn.h`)

((This is my first reaction to "what's involved in supporting openssl 3.0?".))

cperciva:
Not a fan; this would violate the "should compile and work on any POSIX system" requirement.

One option would be to drop in a pure software implementation to replace OpenSSL; since most systems will use CPU intrinsics it wouldn't even matter if the pure software implementation is particularly fast.  The main reason I didn't do this (aside from AENSI being less widespread N years ago) is that I didn't have a conveniently available good quality implementation and didn't want to spend time writing my own.

PR #400, opened 2021-04-19
Improve CRC32 cpusupport, add 32-bit

gperciva:


gperciva:
I'm not particularly excited about using hardware CRC32 on 32-bit platforms.  I mainly just wanted to clean up the SSE42 support as a preliminary step for more SHA256 work, but once I was there I noticed that it would be easy to add 32-bit support.

cpusupport/Build/cpusupport-X86-SSE42.c
```diff
@@ -0,0 +1,17 @@
+#include <smmintrin.h>
+
+#include <stdint.h>
+
+int
+main(void)
+{
+	unsigned int state = 0;
+	unsigned char x = 0;
+	uint8_t buf[8] = {0, 0, 0, 0};
```
cperciva, line 10: `buf[8]`?

cpusupport/cpusupport.h
```diff
@@ -126,7 +126,7 @@ CPUSUPPORT_FEATURE(x86, aesni, X86_AESNI);
 CPUSUPPORT_FEATURE(x86, rdrand, X86_RDRAND);
 CPUSUPPORT_FEATURE(x86, shani, X86_SHANI);
 CPUSUPPORT_FEATURE(x86, sse2, X86_SSE2);
-CPUSUPPORT_FEATURE(x86, sse42, X86_SSE42_64);
+CPUSUPPORT_FEATURE(x86, sse42, X86_SSE42);
```
cperciva, line 129: I'm confused, you're replacing X86_SSE42_64 with X86_SSE42 in this file, but in `cpusupport.sh` you still have both?

alg/crc32c_sse42.c
```diff
@@ -50,8 +56,16 @@ CRC32C_Update_SSE42(uint32_t state, const uint8_t * buf, size_t len)
...
 	 */
 	for (; i < in_block; i += 8) {
+#ifdef CPUSUPPORT_X86_SSE42_64
 		state = (uint32_t)_mm_crc32_u64(state,
 		    *(const uint64_t *)&buf[i]);
+#else
+		/* Only process 32 bits at once. */
+		state = (uint32_t)_mm_crc32_u32(state,
+		    *(const uint32_t *)&buf[i]);
+		state = (uint32_t)_mm_crc32_u32(state,
```
cperciva, line 66: Is there a reason to not make this a loop with `i += 4` in the 32-bit case?

cpusupport/cpusupport.h
gperciva: That is correct.  It's a subtle point that I didn't sufficiently clarify in the commit messages (and/or direct comments in the code).

`cpusupport_x86_sse42.c` checks bit 20 of ECX.  This tells us if the CPU supports SSE4.2.  It doesn't say anything about whether it's in 64-bit mode.

Further investigation reveals that we could put 0x80000001 in EAX and check bit 29 of EDX, which tells us if the CPU is in "Long mode".  So if we wanted to, we could have a separate `cpusupport_x86_sse42_64.c` which checks both bits to ensure that we're in 64-bit mode and have SSE4.2 available.

However, given the current design of `crc32_sse42.c`, I don't see the need for a runtime detection of `sse42` vs. `sse42_64`.  As long as the CPU has sse4.2, we can call `CRC32C_Update_SSE42()`.  If the environment supports 64-bit code at compile-time, that function will include 64-bit instructions; if it doesn't, then it'll only use 32-bit instructions.

Granted, there's cross-compiling to consider.  But `_mm_crc32_u64()` is hidden within `__x86_64__` inside `<smmintr.h>`, and if we compile with `-m32` then that's not enabled.

alg/crc32c_sse42.c
gperciva: I thought the code was easier to read this way, and the performance impact would be minimal in the 32-bit case.

We could certainly have:

```
#ifdef CPUSUPPORT_X86_SSE42_64
        /*
         * Process blocks of 8.  It's ok if (i % 8 != 0), because we're
         * checking that (i < in_block), not (i != in_block).  For example,
         * if we start with i = 4 and want 1 block, we'll end up at i == 12.
         */
        for (; i < in_block; i += 8) 
                state = (uint32_t)_mm_crc32_u64(state,
                    *(const uint64_t *)&buf[i]);
#else
        /*
         * Process blocks of 4.  It's ok if (i % 4 != 0), because we're
         * checking that (i < in_block), not (i != in_block).  For example,
         * if we start with i = 4 and want 1 block, we'll end up at i == 8.
         */
        for (; i < in_block; i += 4) 
                state = (uint32_t)_mm_crc32_u32(state,
                    *(const uint32_t *)&buf[i]);
#endif
```

but that's also not *ideal* for the 32-bit case: the earlier code ensures that we're processing blocks of 8.  So maybe we should expand that to:
```
#ifdef CPUSUPPORT_X86_SSE42_64
        /* Calculate how many bytes are in aligned blocks. */
        remaining_bytes = len - pre_block;
        in_block = remaining_bytes - (remaining_bytes % 8);

        /* Process bytes before the alignment. */
        for (; i < pre_block; i++)
                state = _mm_crc32_u8(state, buf[i]);

        /* If we would process a block, ensure that it's aligned. */
        assert(!((i < in_block) && !((((uintptr_t)&buf[i]) & 7) == 0)));

        /*
         * Process blocks of 8.  It's ok if (i % 8 != 0), because we're
         * checking that (i < in_block), not (i != in_block).  For example,
         * if we start with i = 4 and want 1 block, we'll end up at i == 12.
         */
        for (; i < in_block; i += 8) 
                state = (uint32_t)_mm_crc32_u64(state,
                    *(const uint64_t *)&buf[i]);

        /* Ensure that we don't have too many bytes remaining. */
        assert((len - i) < 8);
#else
        /* Calculate how many bytes are in aligned blocks. */
        remaining_bytes = len - pre_block;
        in_block = remaining_bytes - (remaining_bytes % 4);

        /* Process bytes before the alignment. */
        for (; i < pre_block; i++)
                state = _mm_crc32_u8(state, buf[i]);

        /* If we would process a block, ensure that it's aligned. */
        assert(!((i < in_block) && !((((uintptr_t)&buf[i]) & 3) == 0)));

        /*
         * Process blocks of 4.  It's ok if (i % 4 != 0), because we're
         * checking that (i < in_block), not (i != in_block).  For example,
         * if we start with i = 4 and want 1 block, we'll end up at i == 8.
         */
        for (; i < in_block; i += 4) 
                state = (uint32_t)_mm_crc32_u32(state,
                    *(const uint32_t *)&buf[i]);

        /* Ensure that we don't have too many bytes remaining. */
        assert((len - i) < 4);
#endif
```

When I was trying to proofread that code, I started to wonder if it was really worth having those extra lines -- particularly because I doubt that anybody will be running this on a 32-bit system.  I figured that moving the `#if` into the `for` loop would allow for a minimal difference between 32-bit and 64-bit, and the slight perfomance impact on 32-bit performance didn't matter.

alg/crc32c_sse42.c
```diff
@@ -50,8 +56,16 @@ CRC32C_Update_SSE42(uint32_t state, const uint8_t * buf, size_t len)
...
+		    *(const uint32_t *)&buf[i+4]);
```
cperciva, line 67: style: `s/i+4/i + 4/`

alg/crc32c_sse42.c
cperciva: Got it.  Yeah, what you had is fine.  One thing to consider: On 64-bit systems if we need to process e.g. 31 bytes, we could do 3x _mm_crc32_u8, 1x _mm_crc32_u32, and 3x _mm_crc32_u64 rather than using just _u8 and _u64.  The difference won't show up in our testing since we're not using odd lengths though.

cpusupport/cpusupport.h
cperciva: Got it.  Right, one run-time test for SSE4.2 since it's a single CPU feature, but two compile-time tests since they use different intrinsics.

alg/crc32c_sse42.c
gperciva: True... but how complicated do we want `crc32c_sse42.c` to be?  Or rather: do we ever expect CRC32C calculation to be a bottleneck?  I'd expect kivaloo to be limited by network or disk speed.

I mean, if we're benchmarking something on a particular system and this turns out to be an issue, I'm happy to spend more time on it.  But I don't think that we're at that point.

gperciva:
Rebased, potentially ready for merge.

alg/crc32c_sse42.c
cperciva: Sure.

PR #402, opened 2021-04-22
noeintr_close with optional_mutex

gperciva:


gperciva:
This is an alternate version of noeintr_close, compared to #392.  It adds two functions, `optional_mutex_lock()` and `optional_mutex_unlock()`.  The file `optional_mutex.c` needs to be compiled by each project separately (i.e. not as part of `liball.a`).  If the project includes `-lpthread`, then build system adds `-DOPTIONAL_MUTEX_PTHREAD` to make those two functions call the relevant pthread functions.

gperciva:
(I'm not sold on the name `optional_mutex`, but `if_necessary_mutex` was too verbose, and two minutes looking at the thesaurus didn't come up with anything better.)

gperciva:
Right, of course it won't work if there's any `#ifdef` in the header.  I don't know what I was thinking.  It'll probably work once the initializer is stuffed into a function.

(also, I don't know how it ran on freebsd)

PR #403, opened 2021-05-11
Optional mutex with (void *)

gperciva:


gperciva:
Another option would be require OPTIONAL_MUTEX_PTHREAD to be defined as 0 or 1.  I can't remember why I went with `_YES` or `_NO` originally.

gperciva:
This will be required for `noeintr_close` (see https://github.com/Tarsnap/libcperciva/pull/402).

util/optional_mutex.h
```diff
@@ -0,0 +1,20 @@
...
+#define OPTIONAL_MUTEX_H_
+
+#include <pthread.h>
+
+/**
+ * optional_mutex_lock(mutex):
+ * If OPTIONAL_MUTEX_PTHREAD_YES is defined, call pthread_mutex_lock();
+ * if OPTIONAL_MUTEX_PTHREAD_NO is defined, do nothing.
+ */
+int optional_mutex_lock(pthread_mutex_t *);
```
gperciva, line 11: I considered making this take `void *` instead.  That would eliminate the requirement to `#include <pthread.h>` in the header, which in turn would be an odd thing to be present if we're not doing `-lpthread`.

In the end I went with the clarity of using `pthread_mutex_t *` rather than `void *` as the arguments, but it was a 55%-45% decision.

tests/optional_mutex/main.c
```diff
@@ -0,0 +1,125 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "optional_mutex.h"
+#include "warnp.h"
+
+#define NUM_TIMES 100
+
+static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
```
gperciva, line 11: As far as I know, it's safe to do this even if we don't `-lpthread`.  I mean, the header is required to be present; I don't see anything like [XSI] to indicate that it's optional in POSIX.  So it should be able to compile it, regardless of whether any pthread symbols are used (or are available) in the link stage.

But I acknowledge that it looks and feels weird.

tests/optional_mutex/main.c
```diff
@@ -0,0 +1,125 @@
...
+static int value = 0;
+static int counter = 0;
+
+/* Wait duration can be interrupted by signals. */
+static inline void
+millisleep(size_t msec)
```
gperciva, line 17: I don't think it's worth importing `millisleep.h` from spiped to libcperciva just for this, but I'm happy to be corrected.

gperciva:
Another idea: also add `optional_mutex_init()` and `optional_mutex_destroy()`, to allow client code to operate without any `#include <pthread.h>`

PR #404, opened 2021-05-24
crypto_aes_arm with shims

gperciva:


gperciva:
New beginning for crypto_aes_arm; this replaces #394 and #395.

The most important commit message is in f867e3a.  In case it's useful to see how to move from shims to native code, I've updated #396 and #397.

Overall performance (this PR is `1_mvp`):
compilers

crypto/crypto_aes.c
```diff
@@ -7,18 +7,22 @@
...
 #include "cpusupport.h"
 #include "crypto_aes_aesni.h"
+#include "crypto_aes_arm.h"
 #include "insecure_memzero.h"
 #include "warnp.h"
 
 #include "crypto_aes.h"
 
-#if defined(CPUSUPPORT_X86_AESNI)
+#if defined(CPUSUPPORT_X86_AESNI) | defined(CPUSUPPORT_ARM_AES)
```
cperciva, line 16: `||`

crypto/crypto_aes.c
```diff
@@ -79,6 +83,13 @@ hwtest(struct aes_test * knowngood)
 	crypto_aes_encrypt_block_aesni(knowngood->ptext, ctext_hw, kexp_hw);
 	crypto_aes_key_free_aesni(kexp_hw);
 #endif
+#if defined(CPUSUPPORT_ARM_AES)
```
cperciva, line 86: I'm not incredibly happy with this -- I know we're never going to have CPUSUPPORT_X86_AESNI and CPUSUPPORT_ARM_AES defined at the same time, but this code looks like it's writing to `ctext_hw` twice.  Or zero times (again, also not possible since we would never call hwtest if neither of those is defined, but still...).

Would this work better as `hwtest_x86_aesni` and `hwtest_arm_aes` functions instead?

gperciva: Yeah, I winced at that a few times.  I have code that basically does that (via a callback to `hwtest()`), but I was saving that PR for after this one.  I'll swap the order and do that one first.

gperciva: Followed up in #405.

gperciva:
Rebased on master, potentially ready for merge.

PR #405, opened 2021-05-25
AES prep

gperciva:


crypto/crypto_aes.c
```diff
@@ -63,30 +63,47 @@ static struct aes_test {
...
+	/* Failure! */
+	return (-1);
+}
+
 #if defined(CPUSUPPORT_X86_AESNI)
-	if ((kexp_hw = crypto_aes_key_expand_aesni(knowngood->key,
-	    knowngood->len)) == NULL)
+static int
+test_x86_aesni(const uint8_t * key, size_t len, const uint8_t ptext[16],
+    uint8_t * ctext)
```
cperciva, line 89: Why not `ctext[16]`?

crypto/crypto_aes.c
```diff
@@ -63,30 +63,47 @@ static struct aes_test {
...
 
 	/* Sanity-check. */
 	assert((knowngood->len == 16) || (knowngood->len == 32));
 
-	/* Expand the key and encrypt with hardware intrinsics. */
+	/* Expand the key and encrypt with the provided function. */
+	if (func(knowngood->key, knowngood->len, knowngood->ptext, ctext))
+		goto err0;
+
+	/* Do the outputs match? */
```
cperciva, line 78: Might as well fix this now that we're comparing an output against a known good value rather than comparing two outputs.

gperciva:
Rebased with fixes.  Also adjusted the test vectors to have `ctext[16]` instead of `ctext[32]`.  (Whoops!)

crypto/crypto_aes.c
```diff
@@ -63,30 +63,47 @@ static struct aes_test {
...
+err0:
+	/* Failure! */
+	return (-1);
+}
+
 #if defined(CPUSUPPORT_X86_AESNI)
-	if ((kexp_hw = crypto_aes_key_expand_aesni(knowngood->key,
-	    knowngood->len)) == NULL)
+static int
+test_x86_aesni(const uint8_t * key, size_t len, const uint8_t ptext[16],
```
gperciva, line 88: It occurred to me that we could name this function `x86_aesni_oneshot()` or `x86_aesni_Buf()`, since this function itself isn't "testing" anything.

(The name `x86_aesni_Buf()` follows the pattern of `sha256.h` such as `SHA256_Buf()`, but I hate the capitalization.  I only mention that name as a "brainstorming" thing.)

cperciva:
This looks correct to me, but `hwtest` is starting to look like a rather silly function -- I mean, all it does is call a function pointer and then memcmp.  It might be a bit more useful if it takes an array of test cases and runs through all of them?

gperciva:
Rebased with many fixes to the fixes.  I'm now using `x86_aesni_oneshot()` instead of `test_x86_aesni()`.

PR #406, opened 2021-06-03
crypto_aes.c: test openssl as well

gperciva:


gperciva:
I'm not at all convinced that this is a good idea, but since we had (almost) all the bits and pieces laying around, I figured I'd put them together so you could judge it.

- generalize hwtest() to functest()
- add openssl_oneshot()
- use the above to check the openssl AES implementation; if there's a
  problem, print a warning but continue using it

I experimented with trying to return an error if openssl AES failed, but that required changing the whole `crypto_aes.h` API, so I abandoned it.

crypto/crypto_aes.c
```diff
@@ -154,12 +178,19 @@ hwaccel_init(void)
...
-	    hwtest(arm_aes_oneshot));
+	    functest(arm_aes_oneshot));
 #endif
+
+	/*
+	 * If we're here, we're not using any intrinsics.  Test OpenSSL; if
+	 * there's an error, print a warning, but continue.
+	 */
+	if (functest(openssl_oneshot))
+		warn0("OpenSSL gives incorrect AES values.");
```
cperciva, line 193: I would be inclined to add `abort()` here.  If we don't have any working AES routines, we can't do anything useful anyway.

gperciva:
Rebased to include `abort()`.

PR #407, opened 2021-06-04
crypto_aes.c: typo in openssl_oneshot()

gperciva:


gperciva:
This is such an embarrassing mistake that I'm sorely tempted to rebase it into the previous commit and do a force push to replace master.  (i.e. totally rewriting history from ~6 hours ago)

I'm also really curious about how it worked -- the tests pass, even when run within valgrind.  I guess that the openssl AES_KEY structure happens to store the key first in memory?

gperciva:
Another option would be to switch that function to be:
```
{
        struct AES_KEY kexp;

        /* Expand the key, encrypt, and clean up. */
        AES_set_encrypt_key(key, (int)(len * 8), &kexp);
        AES_encrypt(ptext, ctext, &kexp);
        insecure_memzero(&kexp, sizeof(AES_KEY));

        return (0);
}
```

gperciva:
Ignore that last comment; I must have been confused about which terminal window I was in, when I ran `make test`.

cperciva:
Yeah, let's change it to use `struct AES_KEY` rather than `void *`.

gperciva:
I tentatively think this works.

PR #408, opened 2021-06-07
crypto_aesctr_arm

gperciva:


gperciva:
This doesn't actually depend on the "AES native keygen", so I'm putting it up now.

Nothing exciting about the performance -- I mean, it's almost three times faster with ARM intrinsics, but that's not unexpected.

compilers

PR #409, opened 2021-07-03
Perftest improvement

gperciva:


gperciva:
Same rationale as https://github.com/Tarsnap/spiped/pull/317#issuecomment-873458734.

Just like that PR, this has no change of behaviour.

PR #411, opened 2021-07-24
CPUSUPPORT for FreeBSD with ARM

gperciva:


cpusupport/cpusupport_arm_aes.c
```diff
@@ -31,6 +35,20 @@ CPUSUPPORT_FEATURE_DECL(arm, aes)
...
+
+#if defined(CPUSUPPORT_HWCAP_ELF_AUX_INFO)
+	unsigned long capabilities;
+
+#if defined(__aarch64__)
+	if (elf_aux_info(AT_HWCAP, &capabilities, sizeof(unsigned long)))
+		return (0);
+	supported = (capabilities & HWCAP_AES) ? 1 : 0;
+#else
+	if (elf_aux_info(AT_HWCAP2, &capabilities, sizeof(unsigned long)))
```
gperciva, line 47: This code (for non-64bit arm) is untested, but it's so similiar to the Linux arm32 code that I think it's worth including.

cpusupport/cpusupport_arm_aes.c
```diff
@@ -15,6 +15,10 @@
 #endif /* __arm__ */
 #endif /* CPUSUPPORT_HWCAP_GETAUXVAL */
 
+#if defined(CPUSUPPORT_HWCAP_ELF_AUX_INFO)
+#include <sys/auxv.h>
```
cperciva, line 19: Is there a reason to not merge this with the previous block, since they both `#include <sys/auxv.h>`?

cpusupport/cpusupport_arm_aes.c
```diff
@@ -31,6 +35,20 @@ CPUSUPPORT_FEATURE_DECL(arm, aes)
 	supported = (capabilities & HWCAP2_AES) ? 1 : 0;
 #endif
 #endif /* CPUSUPPORT_HWCAP_GETAUXVAL */
+
+#if defined(CPUSUPPORT_HWCAP_ELF_AUX_INFO)
+	unsigned long capabilities;
+
+#if defined(__aarch64__)
+	if (elf_aux_info(AT_HWCAP, &capabilities, sizeof(unsigned long)))
```
cperciva, line 43: Looking at this, I'm wondering if we should have a `readcaps` macro which invokes either `getauxval` or `elf_aux_info` depending on the macros?

cpusupport/cpusupport_arm_aes.c
gperciva: That felt weird, since the two `<sys/auxv.h>` files are very different.  I figured it would be clearer to have two completely different sections.

For example, what if we added Solaris support for detecting cpu features?  Casual googling suggests that they don't use `<sys/auvx.h>` (if they provide that functionality at all!).  Or what if we wanted to support M1 features (if / when there's documentation about that chip)?

Ditto for having a `readcaps` macro, followed by "cross-platform" code which relied on the same `HWCAP_FOO` names.  Empirically, it looks like that would work, but I'm leery about future expandability.

I'm happy to merge the blocks as much as possible, including `readcaps`, if you think I'm being overly cautious about future expandability about this system-specific code.

cperciva: Ok, make sense.  We can reassess this if/when other platforms land similar support to see if there's enough consistency to be worth merging; I agree that it doesn't make sense to merge the code for 2 platforms if all the rest will be different.

PR #414, opened 2021-09-22
Improve POSIX-compatibility handling

gperciva:


cperciva:
For the cross-compile issue, why not just check if we can compile and run `int main(int argc, char * argv[]) { return (0); }` and disable the check if that program compiles but doesn't run?

gperciva:
Ok.  I'm moving the "runtime checks" to a separate PR; this one is now only for the out-of-tree compiling.

PR #416, opened 2021-10-01
Add DISABLE_POSIX_RUNTIME_CHECKS

gperciva:


gperciva:
Looks like the macOS failure is about `#include <openssl/aes.h>`, not these changes.  I'm investigating further.

Issue #413, opened 2021-08-25
libressl compat


cperciva:
According to my notes, LibreSSL 2.9.0+ has SSL_set1_host, and LibreSSL 3.3.2+ has SSL_set_hostflags.  We should adjust network_ssl_compat.c to reflect this.

gperciva:
Fixed in #422.

PR #422, opened 2021-11-25
Revise OpenSSL compat and add LibreSSL compat

gperciva:


gperciva:
Fixes #413.

I'm not a fan of nested `#if` statements (without indentation).  In email, you said that if necessary we could have `network_ssl_compat_foossl.h`.  I think I'd slightly prefer that, but I can live with what's in this PR.

You mentioned that LibreSSL 2.9.0 added SSL_set1_host.  I haven't made use of that, because there appears to be no harm in using the compat code for SSL_set1_host which is enabled by `NETWORK_SSL_COMPAT_CHECK_HOSTNAME` (whcih we need for `SSL_set_hostflags`).  If desired, I could give that the same treatment I gave `NETWORK_SSL_COMPAT_TLS_VERSION`.

Also worth mentioning is that LibreSSL only added support for modern (Sep 30 onwards) Let's Encrypt certificates in 3.2.7 and 3.3.5.  Additional fixes were added in 3.4.1.  So testing with 2.7.0 or 3.3.2 produces failures in our `perftests/https/test_https.sh`.  https://github.com/libressl-portable/portable/issues/688

gperciva:
Also: I considered splitting this into "revise OpenSSL compat" and "revise LibreSSL compat" PRs.  In the end, I decided that although this is crypto stuff (so an abundance of caution is appropriate), it boils down to "does it compile or not?", so I didn't have to go overboard in ensuring that the changes were as easy to read as possible.

(Related thought: should we have a separate commit which only makes the `0x10101000L` to `0x1010100fL` change?  It would make the git history slightly easier to read, and if this was in C code then I'd probably do it, but it didn't seem worth it for a "does it compile or not?" commit.)

PR #425, opened 2021-12-11
Fix build on arm32

gperciva:


cperciva:
This looks like it's a "fix" in the sense that the build is unbroken, but we won't use AES intrinsics on arm32?

gperciva:
The is aimed at importing into spiped.

e9ae55e fixes the issue reported in https://mail.tarsnap.com/spiped/msg00197.html by testing whether we can compile `vdupq_laneq_u32` as well as the AES-specific instructions.

This would have the effect of breaking AES support for any 32-bit arm system.  However, I don't have access to any such system -- there's no AES support on raspi 2, 3, or 4.  If / when somebody has a 32-bit aes system, I'd be happy to work up a shim for vdupq_laneq_u32 so that they can test it.  At any rate, I don't think it's worth holding back spiped-1.6.2 for some code that I'm unable to (easily) test [2].

[1] https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=vdupq_laneq_u32

[2] that said, I have a vague memory of having ssh access to a server somewhere which had 32-bit aes.  But I'm going to submit this comment now, since you've already commented.

gperciva:
As for [2], I took a look at the current server list that I might have access to, but all of their arm machines are running aarch64.

cperciva:
Ok, just wanted to make sure I understood what the intention was here.  I'm totally fine with saying that arm32 doesn't get acceleration support until we have a user who can test it.

Can you add a commit here which s/ARM/arm64/ in the changelog?  Or does it have sha acceleration even if it doesn't have aes?

gperciva:
("here" is libcperciva; I'll modify the changelog in spiped when I import this.  :)

In theory, our sha256 code should work on an arm32 system which has the Cryptography Extension.  (In the case of the raspi 4, that was rumoured to be a $1 million addition to the licensing fee.)

It's untested, though.  Probably best to have that s/ARM/arm64 in the changelog; if somebody has an arm32 system and suddenly notices that sha256 is faster, I don't think they'll complain.

cperciva:
Oops, yeah I didn't notice which repository this was.  Thanks!

PR #426, opened 2022-01-15
build: remove unnecessary command substition

gperciva:


gperciva:
I'm not 100% confident about this.  It looks good to me, but I can't reproduce the original problem that this POSIXFAIL is trying to avoid.

Recall: back in 2017, macOS X 10.11 with XCode 8 defined `clock_gettime()` in `<time.h>`, but it did *not* include that symbol in the system library.  Also, the default compiler used weak links (or weak symbols), so it didn't try to resolve it until runtime.

macOS X 10.11 reached its end-of-life in 2018-10.  (10.12 is a free upgrade)  FWIW, 10.12 is also eol'd, as is 10.3 and 10.14.

As it happens, I still have a 10.11 machine (I rarely turn it on, so I didn't bother updating).  However, it looks like the compiler uses strong links now -- `posix-cflags.sh` always fails the "can we compile with `clock_gettime`", so the runtime check isn't needed.

/shrug

If you want to merge the patch now, that's cool.  If you want to wait until after the next tarsnap release, that's fine with me.  Or if you want to keep it around until anybody can test it (which probably means "never"), that's also a perfectly reasonable position to take.

gperciva:
History: the previous commit was c4ecffc8eaad7c53565a0956921fb1270d00e875, as part of https://github.com/Tarsnap/libcperciva/pull/75.

PR #428, opened 2022-02-03
tests/valgrind: FreeBSD and getaddrinfo()

gperciva:


gperciva:
No wait, this only works if we're testing against `localhost`; if we're doing `getaddrinfo` with an actual DNS lookup, it leaks more memory.  I'll continue to investigate.

gperciva:
This adapts https://github.com/Tarsnap/spiped/pull/343 for libcperciva.

PR #429, opened 2022-02-04
Add sys/queue.h, use in events_immediate and netbuf_write

gperciva:


gperciva:
Builds on #428 (which was useful to check `perftest/http`).

... whoops.  Turns out it uses `__containerof`, which at first glance seems to be a FreeBSD-specific thing.  I just assumed that `queue.h` would be c99-compliant.

Oh well, that's why we have a test suite!

PR #430, opened 2022-02-04
Import & use sys/queue.h from NetBSD

gperciva:


gperciva:
Replaces #429, because I forgot to re-open that PR before force-pushing (which github can't handle).  Uses `sys/queue.h` from NetBSD, because when you want portability, you obviously look there.  That file is licensed with the 3-clause BSD.

Builds on https://github.com/Tarsnap/libcperciva/pull/428 (which was useful to check perftest/http).

netbuf/netbuf_write.c
```diff
@@ -230,6 +230,7 @@ uint8_t *
 netbuf_write_reserve(struct netbuf_write * W, size_t len)
 {
 	struct writebuf * WB;
+	struct writebuf * OB;		/* Old buffer. */
```
cperciva, line 231: Why not just reuse `WB` for this?

netbuf/netbuf_write.c
```diff
@@ -289,20 +291,27 @@ netbuf_write_reserve(struct netbuf_write * W, size_t len)
 int
 netbuf_write_consume(struct netbuf_write * W, size_t len)
 {
+	struct writebuf * OB;		/* Old buffer. */
```
cperciva, line 287: I would call this `WB`.

netbuf/netbuf_write.c
gperciva: I figured that

>    "use `struct writebuf * OB` instead of `W->tail`"

would be a better git commit message than

>  "use `struct writebuf * WB` to mean `W->tail` sometimes, in addition to also being a new malloced buffer in the same function".

No, really -- the first version of this commit used WB for everything, but when I was double-checking the branch history, it looked like yuck.

Granted, this is a concern for `netbuf_write_reserve` and not `netbuf_write_consume`, but once I'd introduced OB for `_reserve`, I figured that it made sense to use the same name for `_consume`.

... I mean, ok, I'll switch them both to WB if that's definitely what you want.

cperciva: "Refactor code to extract W->tail into a variable for reuse", perhaps?

gperciva: I just went with "use `struct writebuf * WB` instead of W->tail".  Evidently I was over-thinking it.

gperciva:
Rebased to include the refactoring and `WB` stuff in the middle of this branch.

cperciva:
Commit message needs a fix:

`- add a sanity check to ensure that `OB != NULL` in netbuf_write_consume().`

Aside from that I'm ready to merge this.

gperciva:
Rebased with corrected commit message.

PR #432, opened 2022-03-10
Minor refactoring to prepare for queue.h

gperciva:


gperciva:
I'm not certain if you'll like this in a separate PR or not.  My instinct is to do the refactoring first, then have a PR purely for the queue.h stuff.  But you might prefer to have the queue-inspired-refactoring in the same PR as adding queue.h.

cperciva:
Nah, if anything I would split it as "import queue.h" vs. "refactor code and use queue.h".  But the way you did it in #430 is fine.

PR #433, opened 2022-03-11
http: don't malloc(0) if there's no headers

gperciva:


gperciva:
malloc(0) is implementation-defined, so it would be nice to avoid it.

There's no mandatory headers in HTTP/1.0, so we can't simply bail if
a server send a response of:

    HTTP/1.0 200 OK\r\n\r\n

gperciva:
I considered modifying `http_findheader()` to include

```
if (headers == NULL)
        return (NULL);
```

for clarity, but as long as `nheaders == 0` then it'll return `NULL` anyway, so it should be safe.

cperciva:
Hmm, IMALLOC should take care of the differing behaviours here?

gperciva:
```
imalloc(size_t nrec, size_t reclen)
{

        if (nrec > SIZE_MAX / reclen) {
...
} else {
       return (malloc(nrec * reclen));
}
```

Looks like 0 is not greater than SIZE_MAX / reclen, so it'll malloc?

gperciva:
(the github actions failure was that old mpool timing variability, not related to this PR)

I agree, I expected IMALLOC() to handle this case, but apparently it doesn't.  Or rather: it will still call malloc(0), but then return false due to the \#define funkiness.

gperciva:
Note that in this case, we don't want IMALLOC(..., 0, ...) to return non-zero.  Or rather: that wouldn't help the overall program.  I mean, there's nothing wrong with the "HTTP/1.0 200 OK" response, so it would be weird if `gotheaders()` called `die(H)`.

gperciva:
Oh, I almost forgot: I have a tiny python program which implements an HTTP server which gives the minimal response to any GET request.  I'm happy to add that to the git repository if you want, but I expect that you won't python in there.

```
#!/usr/bin/env python3

import http.server

class NoHeadersServer(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        """ Reply to any GET request with:
                HTTP/1.0 200 OK\r\n\r\n
        """
        self.send_response_only(200, 'OK')
        self.headers = ""
        self.end_headers()

def run():
    address = ("", 8000)
    httpd = http.server.HTTPServer(address, NoHeadersServer)
    httpd.serve_forever()

if __name__ == "__main__":
    run()
```

gperciva:
I wonder if we should swap the operands in IMALLOC.  This is what it is now:
```
#define IMALLOC(p, nrec, type)						\
	((((p) = (type *)imalloc((nrec), sizeof(type))) == NULL) &&	\
	    ((nrec) > 0))
```

so that means that it'll call `malloc(0)` and *then* return non-zero.  Maybe the `((nrec) > 0)` should come first?

(I still think that http.c should check for (nheaders > 0) so that it can handle a no-headers response, but the design of IMALLOC is a different issue.)

cperciva:
> I wonder if we should swap the operands in IMALLOC. This is what it is now:
> 
> ```
> #define IMALLOC(p, nrec, type)						\
> 	((((p) = (type *)imalloc((nrec), sizeof(type))) == NULL) &&	\
> 	    ((nrec) > 0))
> ```
> 
> so that means that it'll call `malloc(0)` and _then_ return non-zero. Maybe the `((nrec) > 0)` should come first?

If we do that, `p` won't get initialized.  We could maybe do a `((nrec) > 0) ? imalloc(...) : NULL` but I think it's better to just change imalloc to unconditionally return NULL for `nrec == 0`.

> (I still think that http.c should check for (nheaders > 0) so that it can handle a no-headers response, but the design of IMALLOC is a different issue.)

What's wrong with allocating zero bytes of memory to store zero headers?

gperciva:
> but I think it's better to just change imalloc to unconditionally return NULL for nrec == 0.

That's a perfectly acceptable choice for the API, and I'm happy to make that change.  We'll have to modify some code in kivaloo to accommodate it, though.

According to the c99 spec, `memcpy(NULL, src, 0)` is not allowed [1][2].  There's at least two cases in kivaloo where `test_kvlds.sh` calls IMALLOC with a nrec of 0.

For example, `btree_node_dirty()` contains
```
        /* Leaf or parent? */
        if (N->type == NODE_TYPE_LEAF) {
                printf("N->nkeys: %zu\n", N->nkeys);
                /* Duplicate key-value pairs. */
                if (IMALLOC(N_dirty->u.pairs, N->nkeys, struct kvpair_const))
                        goto err1;
                assert(N_dirty->u.pairs != NULL);
                memcpy(N_dirty->u.pairs, N->u.pairs,
                    N->nkeys * sizeof(struct kvpair_const));
```

(I added the `printf` and `assert`)

At first glance, the uses of IMALLOC in libcperciva, spiped, and tarsnap will be unaffected by returning NULL if nrec == 0.

[1] 7.1.4 Use of library functions
> If an argument to a function has an invalid value (such as... a null pointer..

[2] 7.21.1 String function conventions
> Where an argument declared as size_t n speciﬁes the length of the array for a
function, n can have the value zero on a call to that function. Unless explicitly stated
otherwise in the description of a particular function in this subclause, pointer arguments
on such a call shall still have valid values, as described in 7.1.4.

cperciva:
IMALLOC was already allowed to set a pointer to NULL in response to a request for zero bytes, so the code in question in kivaloo was already wrong.  Yes, it should be fixed (unless we know for some reason that it can't be a zero-byte allocation -- nodes aren't normally empty, but I suppose it could happen for the case of an empty btree?)

gperciva:
> unless we know for some reason that it can't be a zero-byte allocation

I was fully prepared to try unexpected things (such as calling `btree_sync()` twice in a row so that there would be no dirty nodes), but I saw the assertion failures just from a plain `test_kvlds.sh` run.

I'll fix that code.

gperciva:
Ok, I think this is potentially ready (although it depends on #435 and #434).  It might need some nitpicks about comments, though.

http/http.c
```diff
@@ -834,13 +834,17 @@ http_request_cancel(void * cookie)
  * http_findheader(headers, nheaders, header):
  * Search for ${header} in the ${nheaders} header structures ${headers}.
  * Return a pointer to the associated value or NULL if it is not found.
+ * If ${nheaders} is zero, then ${headers} may be NULL.
```
gperciva, line 837: The reason I wanted to add this sentence is because the first sentence is:

> Search for ${header} in the ${nheaders} header structures ${headers}.

If ${headers} is NULL, is it technically "a header structure"?  Or at least, is it a "valid" header structure?
(I'm thinking of the POSIX requirement that `char *` is not NULL.)

As it happens, `http_findheader` can deal with a `headers==NULL` perfectly happily (as long as `nheaders == 0`, which is now enforced with a sanity check).  If we don't want to pass it a NULL, it's certainly possible to modify `http.c` to avoid doing that, but it would involve a bigger set of changes.

cperciva:
The "clarify that ${headers} can be NULL" commit looks good.  That's the only one here which won't go away once you rebase this, right?

gperciva:
That's correct.

cperciva:
Ok, can you do whatever github needs to make the already-merged commits go away?

gperciva:
Done, ready for merge.

PR #434, opened 2022-04-03
imalloc.h: return NULL for zero-byte requests

gperciva:


gperciva:
Split off from #433, but I'm leaving this as a draft until I've fixed the related problems in kivaloo.

Also, since we're altering the API: should we officially state that ${reclen} cannot be zero, and add an assert for that?  Previously, we could end up dividing `SIZE_MAX / 0` if that happened.

cperciva:
Normally `IMALLOC` is used with compile-time fixed size records, so there would be a compile-time division by zero if anything.  But sure, we might as well make it an assertion as well -- it will get compiled away in the common case (since `reclen` is typically known at compile time) and if someone is doing something very wrong it's better to get an assertion error than a SIGFPE.

This isn't really altering the API though -- division by zero is undefined, and whether malloc(0) returns NULL is implementation-defined, so the new behaviour is entirely consistent with the previously documented API.

gperciva:
Right, I should have written "adding to the API" or "strengthening the guarantees of the API", rather than "modifying".

gperciva:
Ok, having looked at things some more, I'm convinced that it's ok to merge this before modifying http.c -- and in fact, if it's done before http.c, that will simplify a bunch of comments.

cperciva:
Good change, but I think the commit message deserves more context.  In particular:

* returning NULL in response to a request for zero bytes was already allowed by the standard, although it is (to quote the FreeBSD man page) "a silly response to a silly question".
* Explicitly returning NULL in this case makes it easier for static analysis tools to recognize that NULL is possible and may help them to find more bugs.

gperciva:
I added the bullet points you suggested, and also:

- Having a fixed response for `malloc(0)` eliminates a source of
  implementation-defined behaviour.

I'm now starting to second-guess the word "response", but I'm not coming up with anything better within a minute.

PR #436, opened 2022-04-19
Misc fixes

gperciva:


util/sock.c
```diff
@@ -90,6 +90,9 @@ sock_resolve_host(const char * addr, const char * ports)
 	for (n = 0, r = res; r != NULL; r = r->ai_next)
 		n++;
 
+	/* Sanity check. */
+	assert(n < SIZE_MAX);
```
gperciva, line 94: There's zero chance that `getaddrinfo()` is actually going to generate a linked list with SIZE_MAX entries in it, but I can't blame static analyzers for not knowing that, so their warning about `n+1` being "potentially an allocation of 0 bytes" is valid.

I have a separate branch which uses an `if (n == SIZE_MAX) warnp("insane number of addresses"); goto err0; }` here, but on second thought I think an `assert` is better: if they're compiling with NDEBUG then it'll be removed, and both humans and computers know that `assert` is used for "if this fails, then something *really* wonky is going on".

cperciva: `assert` is correct here.  Normally we wouldn't use it for sanity-checking something which arrives over the network (like DNS lookups) but having SIZE_MAX addresses shouldn't be possible no matter what evil bits arrive over the network so it's appropriate in this case.

PR #437, opened 2022-05-04
clang nitpicks

gperciva:


alg/md5.c
```diff
@@ -294,11 +294,11 @@ MD5_Buf(const void * in, size_t len, uint8_t digest[16])
  * Initialize the HMAC-MD5 context ${ctx} with ${Klen} bytes of key from ${K}.
  */
 void
-HMAC_MD5_Init(HMAC_MD5_CTX * ctx, const void * _K, size_t Klen)
+HMAC_MD5_Init(HMAC_MD5_CTX * ctx, const void * arg_K, size_t Klen)
```
gperciva, line 297: I've gone back and forth about what to rename this to.  `orig_K`? `var_K`?  `K_`? `p_K`?

In the end, I couldn't think of any reason to choose one over the others, so I figured I'd pick one and open the PR.

events/events_network_selectstats.c
```diff
@@ -88,12 +88,12 @@ events_network_selectstats_select(void)
  * function was called.
  */
 void
-events_network_selectstats(double * _N, double * _mu, double * _va,
+events_network_selectstats(double * arg_N, double * _mu, double * _va,
```
gperciva, line 91: I also couldn't pick whether I should make a minimal change here (only modifying `_N`) or make a consistent change (also modifying `_mu`, `_va`, and `_max` as well, even though those aren't reserved identifiers).

alg/md5.c
gperciva: A similar question, although not part of this particular PR: what naming scheme should we use for functions which are reserved identifiers?  For example, `_SHA256_Update: should that be `p_SHA256_Update`?  `SHA256_Update_`?  `real_SHA256_Update`?  `SHA256_Update_with_tmp`?

gperciva: Another possibility to replace `_SHA256_Update`: we could use `_sha256_Update`.  Of course we want the public API to be in mostly upper-case letters to match the RFC, but the underscore-prepended version is strictly an internal function, so it would be fine to use lower-case for it.

cperciva:
Let's standardize on `_[a-z].*` for this sort of thing.  My reading of the standard is that those are reserved for file-scope identifiers, which should be the only place we're doing this.

gperciva:
So `events_network_selectstats()` should contain

    /* Copy statistics out. */
    *n = N;

? (after adjusting the function argument)

Mathematically, that looks a little bit odd, but ok.  (Also, it's not like `events_network_selectstats()` has confusing math.)

cperciva:
No, let's make it `_n`; the leading underscore denotes "internal implementation name", not just "trying to avoid a namespace collision".

gperciva:
Ok, I'm content that we've sorted out the argument names.

The final item, not included in this PR at the moment, is the naming of ~~cats~~ functions.

> Let's standardize on `_[a-z].*` for this sort of thing. My reading of the standard is that those are reserved for file-scope identifiers, which should be the only place we're doing this.

It's not clear whether your comment was about argument names or function names.

clang complains about `_[a-z].*` static functions:
```
$ more static_func_underscore.c 
static void _foo(void) { }

int main(void) {
        _foo();
        return (0);
}
$ clang -Weverything static_func_underscore.c 
static_func_underscore.c:1:13: warning: identifier '_foo' is reserved because it starts with '_' at global scope [-Wreserved-identifier]
static void _foo(void) { }
            ^
1 warning generated.
```

I thought that was a mistake (`_foo` isn't in the global scope?) so I was getting ready to file a bug report with them.  However, further digging this morning shows that they're deliberately checking that this case produces an error [1].  There was a number of comments about the PR which added `-Wno-reserved-identifires` [2] but not about that particular line.

[1] https://github.com/llvm/llvm-project/blob/main/clang/test/Sema/reserved-identifier.c#L14
[2] https://reviews.llvm.org/D93095?id=314329

I think that the relevant part of 7.1.3 Reserved identifiers is:
> All identifiers that begin with an underscore are always reserved for use as identifiers with file scope in both the ordinary and tag name spaces.

We currently have:
* `alg/sha256.c`: `_SHA256_*`, `_HMAC_SHA256_*`
*  `alg/sha256_sse2.c`: `_mm_bswap_epi32`
* `events/events.c`: `_events_run`

cperciva:
Err, yeah.  I was misreading the standard (in particular, what "reserved for" meant).

For the SHA256 and events functions, lets just call them foo_internal.  `_mm_bswap_epi32` is weird since it's pretending to be am intrinsic but not "shadowing" anything; let's just strip the leading `_`.

gperciva:
Ok, potentially ready for merge.

You might want me to merge the 2nd and 3rd commits.  For development, I kept the "replace _K with K" commit separate from the "replace _X with _x" commit, but in the final version it's probably not important enough to be on its own.

gperciva:
On second thought, I'm sufficiently convinced that you'll prefer to have all the "fix function arguments" changes in one commit that I'll go ahead and merge them right now.

PR #439, opened 2022-05-09
Import network_connect_bind()

gperciva:


gperciva:
Oops, this should wait for https://github.com/Tarsnap/spiped/pull/355.

gperciva:
I've rebased the fixes from https://github.com/Tarsnap/spiped/pull/355 into the appropriate commits, so that the git history will be cleaner.

gperciva:
On second thought, we should wait for https://github.com/Tarsnap/spiped/pull/358 as well.

gperciva:
No reason to delay importing these changes, now that we're working on `sock_addr_ensure_port` in libcperciva rather than spiped.

PR #440, opened 2022-05-20
mpool.h: add sanity check

gperciva:


gperciva:
I'm somewhat half-hearted about the final commit.

* on one hand, it's (practically) impossible to have `M->allocsize == 0`, so in that sense it's a wasted `assert()`.
* on the other hand, clang-tidy spews out 366 lines of warnings (due to stepping through earlier code), which is really annoying.

The test breakage is because this is the first PR which introduces a dependency between header files (`mpool.h` would use `ctassert.h`), and `tests/buildsingles/` didn't have that in mind.  So this PR will have to be delayed anyway.

cperciva:
This looks good to me, any reason it's marked as draft?

PR #443, opened 2022-07-10
Add sock_addr_ensure_port()

gperciva:


gperciva:
This was previously https://github.com/Tarsnap/spiped/pull/358, but I want to add it to libcperciva first because it belongs in the test suite here.

I believe that this PR has addressed all the concerns in the previous PR.

tests/sock_util/main.c
```diff
@@ -0,0 +1,75 @@
...
+} tests_port[] = {
+	{ "/tmp/mysock", "/tmp/mysock" },
+	{ "localhost", "localhost:0" },
+	{ "localhost:80", "localhost:80" },
+	{ "[127.0.0.1]", "[127.0.0.1]:0" },
+	{ "[127.0.0.1]:80", "[127.0.0.1]:80" },
+	{ "[::1]", "[::1]:0" },
+	{ "[::1]:80", "[::1]:80" },
+	/* The below lines are not officially supported. */
+	{ "127.0.0.1", "127.0.0.1:0" },
```
gperciva, line 20: Since we're not officially supporting this form of address (as noted by the comment above), you might prefer it if I deleted them.  I think it's worth keeping, though, since many (most?) platforms will resolve "127.0.0.1" perfectly happily, and if spiped stopped accepting it then some people would report it as a bug.

The ipv6 form of "::1" (without `[]`) is not handled by this function, though.

cperciva: Let's take it out.  It only works because most DNS resolvers will resolve "127.0.0.1" to 127.0.0.1, but I don't think that's required and it could break if your DNS resolver is nonlocal anyway.

util/sock_util.c
```diff
@@ -276,3 +278,70 @@ sock_addr_prettyprint(const struct sock_addr * sa)
...
+			if (asprintf(&bind_addr, "%s:0", addr) == -1) {
+				warnp("asprintf");
+				goto err0;
+			}
+		} else {
+			/* ... which already has a port number. */
+			bind_addr = strdup(addr);
+		}
+	} else {
+		/* addr[0] == '[', so it's an address of type... */
```
cperciva, line 319: This seems like way too much work.  There's no need to distinguish between IPv4 and IPv6; we can simply check "is there a rightmost `:`, and does it have a `]` immediately before it?"

gperciva: Hmm, could I get a third opinion?  A month ago you didn't want to look for `]` and suggested `strspn()` instead:
https://github.com/Tarsnap/spiped/pull/358#discussion_r889726814

gperciva: Updated, although that ended up being a bigger change.
- added a sanity check for not going out of memory if given ":"
- added a test for "should fail" strings, namely ":" and ":1".

gperciva: PS: there's no real point to checking ":1", so I've removed it.

util/sock_util.c
```diff
@@ -276,3 +278,76 @@ sock_addr_prettyprint(const struct sock_addr * sa)
...
+		}
+	} else {
+		/* addr[0] == '[', so it's an address... */
+		if (cr == NULL) {
+			/* ... ipv4, without a port number, so we add ":0". */
+			if (asprintf(&bind_addr, "%s:0", addr) == -1) {
+				warnp("asprintf");
+				goto err0;
+			}
+		} else if (cl == cr) {
```
cperciva, line 332: You know this could also be `cr[-1] == ']'`.

And if you collapse those two, then you can collapse the two "without a port number" cases too:

```
if ((cr == NULL) || (cr[-1] != ']')) {
    /* IPv[46] address without a port number; add ":0". */
    ....
} else {
    /* IPv[46] address with a port number. */
    ...
}
```

gperciva: Ok, changed.  I used different comments (matching ones in the "hostname" section), but the code is the same as you suggested.

Shall I rebase this PR?

util/sock_util.c
```diff
@@ -276,3 +278,67 @@ sock_addr_prettyprint(const struct sock_addr * sa)
...
+		goto err0;
+	}
+
+	/* Figure out what type of address $addr is. */
+	if (addr[0] == '/') {
+		/* It's a Unix domain socket and doesn't need a port number. */
+		bind_addr = strdup(addr);
+	} else if (addr[0] != '[') {
+		/* It's a hostname... */
+		if (cl == NULL) {
```
cperciva, line 314: If you make this `cr` instead of `cl` then you can get rid of `cl`.

util/sock_util.c
```diff
@@ -276,3 +278,67 @@ sock_addr_prettyprint(const struct sock_addr * sa)
...
+	/* Sanity check. */
+	assert(addr != NULL);
+
+	/* Where are the left-most and right-most colons in $addr? */
+	cl = strchr(addr, ':');
+	cr = strrchr(addr, ':');
+
+	/* If the right-most colon is the left-most character, bail. */
+	if (cr == addr) {
+		warn0("Invalid address: %s", addr);
```
cperciva, line 304: I'm inclined to strdup the address here.  The string we return won't be a valid address, but it won't be any more invalid than the one we were given; let the caller decide how to report invalid addresses.

gperciva:
Revised.

gperciva:
I removed the warnp, because we can let the calling code handle the error-printing.  (I noticed that it was inconsistent: it was warning for `asprintf` but not `strdup`.)

cperciva:
Looks good.  Yeah, I usually warn for `asprintf` but since the only plausible error from it is ENOMEM we don't need to report that here.

Please rebase and then I'll merge.

gperciva:
Rebased, ready for merge.

PR #444, opened 2022-07-14
sock_resolve_one: optionally use sock_addr_ensure_port

gperciva:


gperciva:
This PR contains #443.

This would simplify the spiped -b PR, but I'm not certain if you think it's appropriate to include as part of `sock_resolve_one`.

util/sock.c
```diff
@@ -297,21 +298,38 @@ sock_resolve(const char * addr)
...
+		/* ... after ensuring that it has a port (if appropriate). */
+		if ((bind_addr = sock_addr_ensure_port(addr)) == NULL) {
+			warnp("sock_addr_ensure_port");
+			goto err0;
+		}
+		if ((sas = sock_resolve(bind_addr)) == NULL) {
+			warnp("Error resolving socket address: %s", bind_addr);
+			goto err0;
+		}
+		free(bind_addr);
```
gperciva, line 332: This seemed like the most "self-contained" implementation of using `sock_addr_ensure_port()`, but as a result, it will report problems (below) using the original address, rather than the version containing `:0`.

I've gone back and forth about this a few times.  Sometimes I think that's desirable (since the original address is what people would have given it on the command line), sometimes I think it's sloppy (and we should be reporting the `bind_addr`).

On a related note: if you think it's ok to report problems with `addr`, then I should do the same thing for the above warning.

util/sock.c
```diff
@@ -297,21 +298,38 @@ sock_resolve(const char * addr)
...
 	struct sock_addr ** sa_tmp;
-
-	/* Resolve target address. */
-	if ((sas = sock_resolve(addr)) == NULL) {
-		warnp("Error resolving socket address: %s", addr);
-		goto err0;
+	char * bind_addr;
+
+	/* Resolve target address... */
+	if (addport == 0) {
```
cperciva, line 316: I would write this differently:

```
char * resolv_addr = addr;
if (addport && (resolv_addr = sock_addr_ensure_port(addr)) == NULL)) {
    ...
}
if ((sas = sock_resolve(resolv_addr)) == NULL) {
    ...
}
if (resolv_addr != addr)
    free(resolv_addr);
```

Printing errors with the original `addr` is appropriate; while it's not what we're actually resolving but it's what the user gave us and (very importantly) there shouldn't be any errors relating to `resolv_addr` which don't also apply to `addr`.

gperciva: Unfortunately your code has a "discarding const qualifier" thing, since we have `const char * addr`.

I've rewritten this commit to use `strdup()` for the non-addport case.  It's a bit ugly, but this isn't high-performance code.  And playing games with casting to & from a `(uintptr_t)` would be uglier than a `strdup`, I think.

cperciva: Ah, I forgot about that detail.  Canonical way of doing this is with an extra variable.

```
char * resolv_addr = addr;
char * resolv_addr_alloc = NULL;
if (addport && (resolv_addr = resolv_addr_alloc = sock_addr_ensure_port(addr)) == NULL)) {
    ....
}
if ((sas = sock_resolv(resolv_addr)) == NULL) {
    ....
}
free(resolv_addr_alloc);
````

Comment (gperciva) [reply util/sock.c]: grumble, I swear that I checked this twice last night.  Huh, that's a sneaky way of doing it!

I've revised this PR accordingly, and it looks much better now.

File: util/sock.c [hunk 3]
```diff
@@ -297,27 +298,37 @@ sock_resolve(const char * addr)
...
+sock_resolve_one(const char * addr, int addport)
 {
 	struct sock_addr ** sas;
 	struct sock_addr * sa;
 	struct sock_addr ** sa_tmp;
+	char * addr_alloc = NULL;
+
+	/* Prepare the address to resolve. */
+	if (addport && ((addr = addr_alloc = sock_addr_ensure_port(addr))
+	    == NULL)) {
```
cperciva, line 317: Wrap after the `&&` since that's the highest level.

PR #445, opened 2022-08-04
Add apisupport with two checks

gperciva:


gperciva:
Depends on #446.

gperciva:
In #83, you mused about adding support for other APIs (in particular, `getrandom(3)`).  `apisupport.sh` sets of the framework which would allow for that.

At the moment, this PR only uses it for detecting some nitpicky compiler things.  In a few places, we don't strictly obey the C99 & POSIX standards, so some compilers legitimately complain about it.

This PR does not add the capability for our code to do things like `#ifdef APISUPPORT_NONPOSIX_SETGROUPS`.  I have that available in a separate branch, but at the moment there's no need for such things, so I haven't included it.  Likewise, there's no runtime checks (such as a theoretical `apisupport_linux_getrandom()`, similar to `cpusupport_x86_shani()`) in this particular PR.

apisupport/Build/apisupport.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+# Standard output should be written to apisupport-config.h, which is both a
+# C header file defining APISUPPORT_ARCH_FEATURE macros and sourceable sh
+# code which sets CFLAGS_ARCH_FEATURE environment variables.
+SRCDIR=$(command -p dirname "$0")
+
+CFLAGS_HARDCODED="-D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700"
+
+feature() {
+	ARCH=$1
```
cperciva, line 15: s/ARCH/PLATFORM/ ?

gperciva: Hmm, good point.  I'm happy calling it PLATFORM, but at the moment, I'm using that variable to be `NONPOSIX` and `LIBSSL`.  Is that stretching the meaning of "platform"?

cperciva: Maybe a bit of a stretch but it's better than anything else which comes to mind.

gperciva: ok, done as a REBASE commit.

cperciva:
Ok, LGTM.  We should probably change the NO_SETGROUPS check to a APISUPPORT_NONPOSIX_SETGROUPS in setgroups_none.c at some point though?

gperciva:
At some point, yes.

> This PR does not add the capability for our code to do things like `#ifdef APISUPPORT_NONPOSIX_SETGROUPS`

Although maybe "some point" should be "now".

It would be easy to add that back in -- actually, the first 3 drafts of this PR contained that code, but I removed it since it wasn't needed.  This gives me an excuse to get that infrastructure done now.

gperciva:
> We should probably change the NO_SETGROUPS check to a APISUPPORT_NONPOSIX_SETGROUPS in setgroups_none.c at some point though?

On second thought, I'm not sure about this.

Suppose that FreeBSD 14.0 moved `setgroups()` out of `<unistd.h>` and into `<grp.h>`.  What would we want libcperciva to do?

* give a compile error (because we expected `util/setgroups_none.c` to have the correct includes)
* silently make `setgroups_none()` do nothing

I'd say that "compile error" is the appropriate response.

I'll still add the "support preprocessor `#if APISUPPORT_FOO_BAR`" stuff to this PR, but I won't mess with NO_SETGROUPS.  We can revisit that in a separate PR targeted at that case.

gperciva:
Updated.  I put the `cpusupport.sh` changes in a separate #449, but if you'd prefer to just merge it all in this PR, that's fine with me.

There's now a ton of `Makefile` changes to add `-DAPISUPPORT_CONFIG_FILE=\"apisupport-config.h\"` everywhere.

apisupport/Build/apisupport.sh
```diff
@@ -0,0 +1,67 @@
...
+SRCDIR=$(command -p dirname "$0")
+
+CFLAGS_HARDCODED="-D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700"
+
+feature() {
+	PLATFORM=$1
+	FEATURE=$2
+	EXTRALIB=$3
+	shift 3;
+	if ! [ -f "${SRCDIR}"/apisupport-"$PLATFORM"-"$FEATURE".c ]; then
```
cperciva, line 19: Excessive quotes (as in cpusupport.sh).

gperciva:
Fixed, rebased, ready for merge or more review.

PR #446, opened 2022-08-04
Build: remove double spaces in Makefiles

gperciva:


gperciva:
The only real change here is in fbb9715.  `release-tools/metabuild.sh` used to write the build instructions as:
```
echo "	${OUT_CC_BEGIN} ${CF} ${OUT_CC_MID} -c ${S} -o ${F}"
```

but if `${CF}` was blank, then there would be an ugly two spaces between the end of the `CPUSUPPORT_CONFIG_FILE` arg and the `-I ..` arg.

Fixing this problem would allow us to modify `metabuild.sh` to accommodate APISUPPORT without having a ton of "add an extra space" changes.

PR #449, opened 2022-08-10
cpusupport.sh: clarify hardcoded flags, nitpick

gperciva:


gperciva:
Split from #445.  There's more changes to `cpusupport.sh` now, and it's not technically part of apisupport so I think it looks better as a separate branch / PR.

cpusupport/Build/cpusupport.sh
```diff
@@ -9,17 +9,21 @@ fi
...
 SRCDIR=$(command -p dirname "$0")
 
+CFLAGS_HARDCODED="-D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700"
+
 feature() {
 	ARCH=$1
 	FEATURE=$2
 	shift 2;
-	if ! [ -f ${SRCDIR}/cpusupport-$ARCH-$FEATURE.c ]; then
+	if ! [ -f "${SRCDIR}"/cpusupport-"$ARCH"-"$FEATURE".c ]; then
```
cperciva, line 18: Why not just quote the entire path?

gperciva: > Why not just quote the entire path?

I had a vague memory of somebody saying that it was better style to only quote variables.

In any case, when I looked at this file again, I noticed that this mess was defined twice.  I've moved it to a separate variable, which makes the function easier to read.

gperciva:
Revised, including adding a few more comments and the above-mentioned variable.

PR #451, opened 2022-08-29
apisupport: fix -Wno-reserved-macro-identifier name

gperciva:


gperciva:
`-Wreserved-macro-identifier` was only added in clang 13.0.0 [1][2],
as a synonym for `-Wreserved-id-macro`.  For compatibility, we should use
the earlier flag name.

[1] https://releases.llvm.org/13.0.0/tools/clang/docs/DiagnosticsReference.html
[2] https://releases.llvm.org/12.0.0/tools/clang/docs/DiagnosticsReference.html

PR #452, opened 2022-08-29
Avoid reserved identifiers

gperciva:


util/getopt.h
```diff
@@ -45,9 +45,9 @@
 /* Select the method of performing local jumps. */
 #ifdef GETOPT_USE_COMPUTED_GOTO
 /* Workaround with computed goto. */
-#define DO_SETJMP _DO_SETJMP(__LINE__)
-#define _DO_SETJMP(x) __DO_SETJMP(x)
-#define __DO_SETJMP(x)							\
+#define DO_SETJMP libcperciva_DO_SETJMP(__LINE__)
```
gperciva, line 48: I'm not certain how you want to approach this.  The whole set of
#define A _A(foo)
#define _A(foo) __A(foo)

is fairly idiomatic, but it's technically violating C99's rules about reserved identifiers.

In a previous version, I used "p_" and "p__" for "private".  That had the advantage of not being so long, but it looked a bit icky.  That said, the current version also looks a bit icky.

cperciva:
Can we just fix `queue.h` rather than adding extra gunk to our build system?

util/getopt.h
cperciva: I'm fine with `p_foo`, `libcperciva_foo`, or even `foo_`.  The `libcperciva_foo` option minimizes the odds of namespace collisions though, so I have a slight preference for sticking with this.

gperciva:
This version of the PR is now based on the "experimental" #454, plus changing the header guards.  No `Makefile` changes are necessary now.

PR #453, opened 2022-08-30
Cpusupport flag order

gperciva:


gperciva:
There's two competing principles here:

  1) the environment CFLAGS should be able to override all
     build-system flags.
  2) the CPUSUPPORT and APISUPPORT framework asks "given the specified
     CFLAGS, what additional flags do we need to add in order to compile
     xyz?"; and we should then use those flags in the build system.

Previously, our build system implemented 1; this commit changes it to 2.

The reason this question didn't come up earlier is that

    clang -Weverything

has one surprising quirk: it enables all warnings *that have not
previously been disabled*.  By contrast, if you specify a warning
manually, then the later command-line flags take precedence.

In other words, as we would expect,

    clang -Wno-cast-align -Wcast-align

enables `-Wcast-align` since that flag comes last; however,

    clang -Wno-cast-align -Weverything

does *not* enable `-Wcast-align`.

gperciva:
As a result, if you have `CFLAGS=Wcast-align`, then you end up with the puzzling:

```
 -c ../alg/crc32c_sse42.c -o crc32c_sse42.o
../alg/crc32c_sse42.c:61:8: error: cast from 'const uint8_t *' (aka
      'const unsigned char *') to 'const uint64_t *' (aka
      'const unsigned long *') increases required alignment from 1 to 8
      [-Werror,-Wcast-align]
                    *(const uint64_t *)&buf[i]);
                     ^~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
*** Error code 1
```

when `cpusupport-config.h` clearly contains
```
export CFLAGS_X86_SSE42_64="-msse4.2 -Wno-cast-align"
```

Granted, nobody is likely to add warnings flags with clang; it's more idiomatic to slap a `-Weverything` in there.

However, `gcc` has no such option; all warnings need to be enabled manually (other than `-Wall` and `-Wextra`).  So unless we change this, there's no way for `cpusupport-config.h` to override any warnings in CFLAGS.

gperciva:
An alternate solution (which I don't think is a good one), would be to add a separate `CFLAGS_WARNINGS` which comes before the cpusupport flags.  We could then leave the regular `CFLAGS` at the end of the command-line.  That would allow cpusupport to override the `CFLAGS_WARNINGS`, but still allow `CFLAGS` to override anything.

I think the principle of least surprise leads to this commit, rather than something like `CFLAGS_WARNING`.

Oh, another oddball idea: use `CPPFLAGS` for warnings, then our cpusupport flags, then `CFLAGS`.  Again, it would be a rather surprising solution and I don't like it, but technically it would work.

PR #454, opened 2022-08-31
Use trailing underscores for macros

gperciva:


gperciva:
Split from #452.  This one:

- changes macro names in `queue.h`
- uses trailing underscores

I know that you said that you have a "slight preference" for `libcperciva_` to avoid namespace collisions, but I figure that if `GETOPT_OPT` doesn't have a collision, then `GETOPT_OPT_` would also be safe.  And if `GETOPT_OPT` has a collision, then using `libcperciva_` for the underscored version won't help.

This PR is 80% an experiment, but I prefer this version for because keeps the diffs small and easier to understand, particularly in `queue.h`.  (We needed multiline `#define` macros for longer `GETOPT` names such as `GETOPT_MISSING_ARG`)

PR #456, opened 2022-09-11
Actions: update to Ubuntu 22.04

gperciva:


gperciva:
Github actions is deprecating ubuntu 18.04.

The normal course of action would be to update to 20.04, but unfortunately valgrind 3.15 (which is what comes with ubuntu 20.04) claims that there is an uninitialized variable in `__get_cupid()` when we attempt to use it for `cpusupport_x86_shani`.

The definition of `__get_cpuid()` in `cpuid.h` header looks fine to me, and valgrind 3.19 (with ubuntu 22.04) doesn't report any problem.

I couldn't find any relevant bug fixes in valgrind, but then again it used to be rare to find CPUs which supported SHANI.  (I'm quite surprised that it's included in the github actions servers!)

For ease of history, I kept two separate commits here: upgrading to 20.04 and then 22.04, since each step requires different clang nitpicks.

PR #457, opened 2022-09-15
Misc test & script improvements

gperciva:


apisupport/Build/apisupport.sh
```diff
@@ -11,6 +11,14 @@ SRCDIR=$(command -p dirname "$0")
 
 CFLAGS_HARDCODED="-D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700"
 
+# Do we want to record stderr to a file?
+if [ "${DEBUG:-0}" -eq "0" ]; then
```
gperciva, line 15: I considered calling this something like `DEBUG_SUPPORT`, but in the end I thought that was less clear than simply `DEBUG`.

Another option would be to call it `DEBUG_CPUSUPPORT`.  Granted, that's slightly misleading since it applies to apisupport as well as cpusupport... but at least it would be clear that the flag refers to debugging our platform compatibility layer(s), rather than setting `-DDEBUG` in the compiler or anything like that.

cperciva:
Will merge as soon as the edit conflict is fixed.

gperciva:
Rebased, ready for merge.

PR #459, opened 2022-09-20
gcc nitpicks

gperciva:


gperciva:
This is the final set of nitpicks for gcc-12 with the maximum set of warnings I could enable.  (There's a few more warnings, but those arise from bugs in gcc).

You'll probably want me to merge the latter two commits together.  I put them separately because I personally found it hard to double-check it when I made all those changes at the same time, but you're much more familiar with integer promotion, so maybe it'll be obvious to you and not worth keeping as separate commits.

PR #460, opened 2022-10-18
crypto_aes.c cleanup

gperciva:


cperciva:
Is there a reason to use `!= 0`?

gperciva:
Because openssl can't effing decide whether to return 0 or 1 on success.  In this case, they decided to use 0 for success; but for example with BN_*, they use 1 for success.  I feel safer if the expected value is spelled out when dealing with libcrypto.

If you don't think it's necessary (and in this case, they *did* use the expected values), I'm quite content to remove the `!= 0`.

cperciva:
No, that's fine, as long as there was a reason. :-)

PR #461, opened 2022-11-05
build: use .PHONY where appropriate

gperciva:


gperciva:
Accepted into POSIX issue 8, but not yet published. https://www.austingroupbugs.net/view.php?id=523

PR #463, opened 2022-11-25
actions: enable verbose output upon failure

gperciva:


gperciva:
Example of a deliberate test failure: https://github.com/Tarsnap/libcperciva/actions/runs/3545279806

PR #464, opened 2022-11-25
Add warnp() to close()

gperciva:


gperciva:
I debated whether to add

    if (errno != EINTR)

before the `warnp()`, but I figured that I'd raise it as a question for you, rather than changing the code yet.

My initial instinct is to not add it yet; we can re-evaluate it if/when we start seeing warning messages about EINTR.  Also, if we're thinking about switching to a `noeintr_close()` then that's a moot point anyway.

PR #466, opened 2022-12-18
Add warnp() to fclose()

gperciva:


util/readpass.c
```diff
@@ -173,8 +173,10 @@ readpass(char ** passwd, const char * prompt,
 	resetsigs(savedsa);
 
 	/* Close /dev/tty if we opened it. */
-	if (readfrom != stdin)
-		fclose(readfrom);
+	if ((readfrom != stdin) && fclose(readfrom)) {
+		warnp("fclose");
+		goto err1;
```
gperciva, line 178: This does change program behaviour: previously, if we failed to `fclose()`, then `readpass()` would still exit successfully as long as `strdup` worked.  This commit changes that to make `readpass()` fail if `fclose()` had an error.

My initial instinct is that if anything goes wrong, we'd want `readpass()` to fail.  But in this case, we've opened `/dev/tty`.  I'd expect that we should be able to close that, but is there any weird OS that treats `/dev/tty` specially?  I'm not confident about answering that question.

... oh wait, this gets into the EINTR thing again, doesn't it?  If `fclose()` receives an EINTR, then we probably want to still proceed with the `readpass()`.  (whether or not we've released the file descriptor is a separate question, for which we have a solution but it's postponed)

gperciva:
Ignore the previous comment; I've fixed the PR to avoid changing program behaviour (other than printing error messages).

PR #467, opened 2022-12-30
daemonize.c: close unneeded pipe

gperciva:


gperciva:
You're probably not going to like this one, but I figured that I'd try anyway.  There's two schools of thought about freeing resources immediately before an `_exit()`.

1) we're exiting anyway, so the OS will clean up after us, so why bother?
2) it's polite to free things explicitly, and that reduces false positives from error-checking programs, so why not?

(I fall into category 2, of course.)

util/daemonize.c
```diff
@@ -62,6 +62,10 @@ daemonize(const char * spid)
 				/* EOF -- the child died without poking us. */
 				goto err1;
 			case 1:
+				/* We don't need to read from this any more. */
+				if (close(fd[0]))
```
gperciva, line 66: FWIW, the client closes `fd` in lines 108 to 119.

cperciva:
Hmm.  *Does* this reduce false positives from leak-checkers?  There are all sorts of resources which are forked when we daemonize -- memory allocations, files and sockets which were opened before forking, etc -- so any decent leak-checker is going to need to be aware of the fork-and-parent-exits idiom.

I'm ok with this if it actually helps, but considering that we can't do anything about all the resources which were created before daemonize gets called, I'm not sure that closing this pipe makes a difference here.

gperciva:
> *Does* this reduce false positives from leak-checkers?

Reduce?  Yes.  On current master,

```
$ make test USE_VALGRIND=1 N=19
...
$ more tests-valgrind/19-daemonize-00-2248.log
...
==2248== FILE DESCRIPTORS: 5 open (3 std) at exit.
==2248== Open file descriptor 4:
==2248==    at 0x492F56A: pipe2 (in /lib/libc.so.7)
==2248==    by 0x2022BA: daemonize (in /data/td/src/libcperciva/tests/daemonize/test_daemonize)
==2248==    by 0x20224A: main (in /data/td/src/libcperciva/tests/daemonize/test_daemonize)
...
```

(the pid will change on each run)

((valgrind itself uses a fd, so we expect to have 3 std + 1 valgrind fd open when a process exits.))

That "leak" doesn't show up as an error in our testing because we deliberately ignore the "leaks" in the parent PID thanks to:
>https://github.com/Tarsnap/libcperciva/pull/271/commits/cc4cfc39d4a2bc1d0ff9d065305649e6ef4ac401
>#271 

After applying this PR, the report about an unexpectedly-open file descriptor goes away; and in fact, there will be no leak reports which arise in `daemonize.c` at all.


This whole thing arose because I was using `daemonize.c` as a template for an upcoming `daemonize_func.c` (or maybe `fork_func.c`?), and I was surprised to find "leaked" resources from my code.  Granted, `tests/daemonize/main.c` will still result in one valgrind warning (arising from `WARNP_INIT` not being free'd in the parent pid).  But I still think it's useful to eliminate any leak which arises in `daemonize.c` itself (as opposed to external code).

gperciva:
PS to further clarify:

- in the context of a real program, this would reduce the number of leak reports in the parent (which we ignore anyway) from N to N-1, basically all from `atexit()` functions (deliberately) not being called by the parent pid.

  For example, `spiped` has 11 "leaks" from the parent pid.  (But 0 from the client pid, which is the one we care about.)

  So this isn't too exciting when used with real programs.

- However, if we had a minimal test specifically for our `daemonize()` function, we wouldn't need to set up `getopt_atexit()` and related functions; so this could take the warnings from 1 to 0.

cperciva:
Ok, I'll take it, but please amend the commit message to say "this is purely for the benefit of leak checkers; since the parent process is about to call _exit there's no functional effect to closing this pipe here" so future readers of the code are less likely to be confused.

gperciva:
Modified.  I also added the comment about `_exit()` freeing resources as a code comment, since the git history is less handy for finding information.

gperciva:
github actions is finding hash mismatches when running `apt-get update`.  Nothing to do with our code.

I'll try triggering the compile again in a few minutes or hours.

gperciva:
Looks like the actions thing is known: 

> We recently moved to a pull-through cache to improve some elements of performance and storage allocation. We configured a low Time-To-Live (TTL) for the metadata, but we see that is insufficient, as artifacts can fall out of sync (i.e. client gets a fresh InRelease file but a cached Packages file). We're expediting a fix to disable caching of these metadata files, to ensure requesters always get the latest metadata.
> https://github.com/microsoft/linux-package-repositories/issues/34

Anyway, this PR is ready for another review and potentially merging.

PR #468, opened 2023-01-13
Slightly refactor daemonize.c

gperciva:


gperciva:
This includes #467.

I'm fairly confident that `noeintr_readbyte()` is a good clarifying addition.  Less confident about `writepid()`... I found it useful to abstract that away when I was trying to understand this code, but you might prefer to keep it in `daemonize()` itself.

util/daemonize.c
```diff
@@ -7,6 +7,41 @@
 
 #include "daemonize.h"
 
+/* Read one byte from ${fd}, looping upon EINTR.  Do not store that byte. */
+static int
+noeintr_readbyte(int fd)
```
cperciva, line 12: Since we have a `noeintr.c`, we shouldn't have functions named `noeintr_*` in other files.

util/daemonize.c
```diff
@@ -53,29 +88,19 @@ daemonize(const char * spid)
 			warnp("close");
 			goto err1;
 		}
-		do {
-			switch (read(fd[0], &dummy, 1)) {
```
cperciva, line 57: I think we can get rid of the dummy variable now?

gperciva: Unfortunately it's needed for writing to `fd`:

```
        /* Tell the parent to suicide. */
        if (noeintr_write(fd[1], &dummy, 1) == -1) {
                warnp("write");
                goto die;       
        }
```

I was very tempted to move that `noeintr_write()` into a separate function (maybe called `sendbyte()`?), to get rid of `dummy` and to be semantically clearer, but I figured that would be over-engineering it.

util/daemonize.c
gperciva: done

util/daemonize.c
```diff
@@ -62,6 +62,10 @@ daemonize(const char * spid)
 				/* EOF -- the child died without poking us. */
 				goto err1;
 			case 1:
+				/* We don't need to read from this any more. */
```
gperciva: I added this in #467, which you haven't commented on.

(I just wanted to make sure that you're fully aware of this change)

util/daemonize.c
```diff
@@ -7,6 +7,89 @@
...
+	return (0);
+
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+/* Send a byte to ${fd}, ignoring EINTR. */
+static int
+sendbyte(int fd)
```
gperciva, line 47: I added this as a `MAYBE???` commit.

I'd actually prefer to include this change, but I'm betting that you'll say that the (source-code) overhead of `sendbyte()` isn't worth it.

util/daemonize.c
cperciva: Right, I forgot about that.  Yeah no point moving that code.

util/daemonize.c
cperciva: Yeah, not worth it.

gperciva:
Rebased.

util/daemonize.c
```diff
@@ -7,6 +7,69 @@
 
 #include "daemonize.h"
 
+/* Read one byte from ${fd}, looping upon EINTR.  Do not store that byte. */
```
cperciva, line 10: "Read and discard one byte from ${fd}, looping upon EINTR."

cperciva:
Good aside from minor rewording of a comment.

gperciva:
Rebased.  I also added "and discard" to the comment where we're calling `readbyte()`:

```
/* Read and discard a byte from the read end of the pipe. */
if (readbyte(fd[0]))
```

PR #469, opened 2023-01-14
Misc

gperciva:


gperciva:
(I meant to push this as a branch called `misc`, but oh well.  I've added 3 minor commits to this branch.)

It would be nice to be consistent in how we name the `atexit()` functions, and as an added bonus for grepping or tools like `ctags`, it would be lovely if those names weren't generic names such as `done()`.

The function-renaming commit does not modify `mpool.h`, because I figured that it's not worth messing with the macros in that file.

cperciva:
Can you make the atexit handlers named `foo_atexit` rather than `foo_shutdown`?  I've used "shutdown" to refer to a cleanup which gets called deliberately rather than one which is invoked automatically.

cperciva:
The rest looks good.

gperciva:
Rebased, ready for merge.

PR #470, opened 2023-02-08
Optional mutex with (pthread_t *)

gperciva:


gperciva:
This is the original version of #403, where we use `pthread_t` in the header.

cperciva:
I thought I'd like the (void *) version better, but after comparing them I prefer this one since it's simpler; also in many cases we're going to need PTHREAD_MUTEX_INITIALIZER anyway so requiring <pthread.h> isn't much of a burden.

I'm not a fan of needing to specify optional_mutex.c in Makefile.BSD; it's ok for now but later on (once it starts appearing in more places) we might want to push that into Makefile.inc?

gperciva:
> I'm not a fan of needing to specify optional_mutex.c in Makefile.BSD; it's ok for now but later on (once it starts appearing in more places) we might want to push that into Makefile.inc?

Are you thinking that:
1) we should always compile `optional_mutex.o` regardless of whether or not we need it, because the linker can decide if it needs to use it or not; or

    This strikes me as being "untidy"; I don't like the idea of seeing files like `tests/sysendian/optional_mutex.o` given that the test doesn't include any open/close or synchronization at all.


2) we should add some black magic to `Makefile.inc` and/or `metabuild.sh` so that it figures out if we use `optional_mutex` anywhere?

    This would require a very "messy" solution.  It's not enough just to grep all source files mentioned in `Makefile.BSD` for `#include "optional_mutex.h"`, because that could be used by a source file that's already compiled into `liball.a`.  (For example, the upcoming `noeintr_close.h`.)

    So to implement this, I think that `make Makefiles` would need to compile everything, check if can link; then if it fails to link, add `optional_mutex.c` to the sources, recompile, and if it then works then add it to the final generated `Makefile`.  That was vastly decrease the speed of `make Makefiles`, and hugely increase the complexity.

    I mean, it wouldn't be impossible to do.  And a small part of me would enjoy the challenge of a) creating such a build system, and b) trying to make that build system as simple and elegant as possible.  But I honestly can't imagine it being worth the complexity just to avoid
    ```
    # Handle optional_mutex
    .PATH.c	:	${LIBCPERCIVA_DIR}/util
    SRCS	+=	optional_mutex.c
    IDIRS	+=	-I${LIBCPERCIVA_DIR}/util
    ```
    in the relevant projects.

cperciva:
I was thinking "always compile `optional_mutex.o`" since I figured that once `noeintr_close` used it we would end up pulling it into most binaries anyway.

But now I'm wondering if we can have `liball/Makefile` compile two `.o` files, *not include them in `liball.a`*, and have `metabuild.sh` pick the right variant instead of adding the extra preprocessor option.  Let's leave it for now though.  There's no rush.

FWIW the main reason I don't like having those lines isn't because I want to avoid a few extra lines of Makefile, but rather because "some parts of libcperciva use optional mutexes" is an implementation detail which I don't really want exposed.

PR #474, opened 2023-04-08
build: add LIBS variable

gperciva:


cperciva:
I'm guessing this is for the benefit of another project which has more than one library?

gperciva:
Mainly for optional mutex.  I'm not finished double-checking, but #475.

While I was working on that, it occurred to me that this was a gap in the "generality" of the build system, so whether or not you liked #475, this would be a worthwhile improvement.

cperciva:
Got it.  I'll merge this now and we can handle #475 when you're satisfied that it's ready.

PR #475, opened 2023-04-12
Use a library for optional mutex

gperciva:


gperciva:
Ok, this commit and #476 are ready for review.  This eliminates the need for programs to manually compile `optional_mutex.c` themselves; instead, `liball.a` will have two separate libraries compiled, and programs need to link to the appropriate one.

* I'm 80% confident this is the approach we want to use in the end.  The only quibble is whether having `optional_mutex_lock()` in a separate library would be slower than compiling it into `liball.a` itself.  Given a Sufficiently Advanced Compiler (tm) this would be a non-issue, but I'm not certain where C compilers are at these days.
  The alternative would be to compile a `liball_normal.a` and `liball_pthread.a`; then programs would simply link to the appropriate one, rather than `liball.a` containing missing symbols.
* I'm 60% confident that the git history is appropriate to get the source tree to the result we want.

cperciva:
Just to confirm, #476 includes this one so I can just merge that once I'm happy and ignore this PR?  (Thank you for offering both options but I'm ok with merging it all at once rather than merging "create libraries which aren't used by anything yet" first.)

gperciva:
Yes, #476 includes this one.

PR #477, opened 2023-04-14
elasticarray: avoid undefined behaviour

gperciva:


gperciva:
Discovered UB due to a clang-16 warning.  Very nice & useful comment about casting function pointers: https://stackoverflow.com/a/559671

I'm confident that this is the code change we want; I'm less certain about exactly what you'd like to have in the commit message.

cperciva:
```
The previous prefix##_iter() was invoking undefined behaviour.  Given
two functions:
    void fp_one(rectype *)
    void fp_two(void *)
prefix##_iter() was accepting a function pointer in the form of fp_one,
then casting it to the form of fp_two and passing it to
elasticarray_iter(), which violated:
    A pointer is used to call a function whose type is not compatible
    with the pointed-to type (6.3.2.3).
    - C99 Annex J.2 Undefined behaviour
```

Casting the function pointer and passing it to elasticarray_iter wasn't the undefined behaviour; the undefined behaviour was that elasticarray_iter called the function without casting the pointer back to the correct type (which elasticarray_iter didn't know, of course).

Probably worth noting in the commit message that another alternative would be to have elasticarray_iter invoke a callback which converts the function pointer back the correct type; this would add the performance overhead of an additional function call, but would allow elasticarray_iter to take advantage of internal knowledge of the elasticarray type rather than making repeated _get and _getsize calls.  Right now I don't think we use elasticarray_iter for anything where this matters but it's something to be aware of if future profiling reveals that we're spending too much time here.

gperciva:
Agreed about what the undefined behaviour was.  I thought that's what the second paragraph of the commit message said?  (other than the typo "benen")

I'm a bit confused about what you want the commit message to be.  Based on the github formatting, it looks like you want me to keep the first paragraph only?

I've removed the second paragraph, but kept the final bit which explains why the commit is removing `elasticarray_iter()`.

> another alternative would be to have elasticarray_iter invoke a callback which converts the function pointer back the correct type

I don't see how that's possible, unless you meant `prefix##_iter()` instead of elasticarray_iter()?  I mean, in order for `elasticarray.c` to compile, the function declaration of the callback would have to return `(void *)`, which gets us no closer to having the correct type.

Oh wait, are you suggesting that the fp we pass to `prefix##iter` should take `void *` rather than `rectype *`?  Sure, that would work, but it would be a shame to lose the type-checking.

> Right now I don't think we use elasticarray_iter for anything where this matters but it's something to be aware of if future profiling reveals that we're spending too much time here.

Right, `_iter()` is not used in any of our released projects, and even when it's used, it's only in clean-up.

cperciva:
> Agreed about what the undefined behaviour was. I thought that's what the second paragraph of the commit message said? (other than the typo "benen")
> 
> I'm a bit confused about what you want the commit message to be. Based on the github formatting, it looks like you want me to keep the first paragraph only?

No, I think the second paragraph was fine.  The first paragraph incorrectly claimed that *passing the function pointer* was UB; the UB happens when the pointer gets used to invoke a function.

> > another alternative would be to have elasticarray_iter invoke a callback which converts the function pointer back the correct type
> 
> I don't see how that's possible, unless you meant `prefix##_iter()` instead of elasticarray_iter()? I mean, in order for `elasticarray.c` to compile, the function declaration of the callback would have to return `(void *)`, which gets us no closer to having the correct type.
> 
> Oh wait, are you suggesting that the fp we pass to `prefix##iter` should take `void *` rather than `rectype *`? Sure, that would work, but it would be a shame to lose the type-checking.

I'm saying that `elasticarray_iter` could operate on (void)(void *) and we could have `prefix##_iter` which calls `elasticarray_iter` and has callbacks into `prefix##_iter_callback`, which casts from `void *` to `rectype *` before calling the *real* callback.

> > Right now I don't think we use elasticarray_iter for anything where this matters but it's something to be aware of if future profiling reveals that we're spending too much time here.
> 
> Right, `_iter()` is not used in any of our released projects, and even when it's used, it's only in clean-up.

gperciva:
Oh!  I thought that when you had the \`\`\` in the github comment, that was what you wanted the commit message to be.  I didn't realize you were quoting it for commenting upon.

gperciva:
Tweaked the commit message.

PR #479, opened 2023-04-27
Clean up building optional_mutex.c

gperciva:


gperciva:
These modifications do not change the generation `Makefile`s at all.

PR #482, opened 2023-05-26
scripts: use double quotes where appropriate

gperciva:


gperciva:
This is not an entirely stylistic change; the test suite currently does not support spaces in directory names.  So if somebody tried:
```
git clone https://github.com/Tarsnap/libcperciva.git "evil libcperciva"
cd evil\ libcperciva
make test
```

then they'd see an ugly failure.

I have this fixed in a separate branch for `make test`, but not `make test USE_VALGRIND=1.  But this is a good step towards supporting it.

(Yes, I suspect that changing the definition of `${scriptdir}` to avoid an absolute path would fix most (if not all) of the "evil libcperciva dirname" problems.  But we'd still have a problem if we wanted to add a test called "87-evil test name.sh")

PR #484, opened 2023-05-26
netbuf_read: refactor with _resize_buffer()

gperciva:


gperciva:
There's no pressing need for this (it arose from an spiped experiment which didn't pan out).  But I think that it usefully simplifies `netbuf_read_wait()`.

PR #485, opened 2023-05-29
API comments for internal functions

gperciva:


gperciva:
This is the question posed in https://github.com/Tarsnap/libcperciva/commit/d66012de8078e3aca8c69cca885ca0ea48958194#r115505879.

I've moved it into a (temporary?) PR for better visibility.

PR #486, opened 2023-06-05
tests: add wait_while(), plus more nitpicks

gperciva:


tests/shared_test_functions.sh
```diff
@@ -108,16 +108,32 @@ has_pid() {
 	return 1
 }
 
-## wait_for_file (filename):
-# Waits until ${filename} exists.
-wait_for_file() {
-	filename=$1
-	while [ ! -e "${filename}" ]; do
+## wait_while(func):
+# Wait while ${func} returns 0.  If ${msleep} is defined, use that to wait
```
gperciva, line 112: The current function doesn't return non-zero (it loops infinitely), but I'm leaving the door open for a later version to take a timeout value and return non-zero if that timeout is exceeded.

(The current kivaloo tests have a few "wait X seconds; if Y isn't done, bail with an error" cases.)

tests/shared_test_functions.sh
```diff
@@ -108,16 +108,32 @@ has_pid() {
...
+	while "$@"; do
+		# Notify user (if desired)
 		if [ "${VERBOSE}" -ne 0 ]; then
-			echo "Waiting for ${filename}" 1>&2
+			printf "waited\t%ims\t%s\n"		\
+			    "${_wait_while_ms}" "$*" 1>&2
+		fi
+
+		# Wait using the appropriate binary
+		if [ -n "${msleep:-}" ];  then
```
gperciva, line 126: libcperciva does not have an `${msleep}` variable defined, and at I have no current plans to import our `msleep` binary to it.

However, `msleep` is currently used in kivaloo, and I'll be importing it to spiped, so it's worth including this in the generic `shared_test_functions.sh` file.

PR #487, opened 2023-06-26
Add fork_func

gperciva:


gperciva:
There's more overhead in `fork_func(); fork_func_wait();` than in pthread, but it's less than 1ms, and more importantly, there isn't a lot of variability.

Tests when waiting for 0ms, 1ms, 30ms, and 90ms on FreeBSD 12.4 with kern.timecounter.alloweddeviation=0:

0ms
1ms
30ms
90ms

gperciva:
This PR includes #488.

gperciva:
Having a separate file `check_exit.c` seems like overkill.  However, I wanted to have `check_perftest.c` and `check_order.c` in separate files.  Once that was done, I could either have the separate `check_exit.c`, or add a `check.c` which only contained the exit-checking-related functions.

gperciva:
Testing with valgrind on macOS does require the sleep-sort values to be separated by more than 100ms.  Empirically, when tested with 300ms and 400ms, the order was sometimes reversed (1 out of 2 attempts; I didn't try more tests to get a statistical estimate).

I went with 200ms, because occasionally-failing tests are really annoying.

util/fork_func.c
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * fork_func(func, cookie):
+ * Fork and run ${func} in a new process, with ${cookie} as the sole argument.
+ */
+pid_t
+fork_func(int (* func)(void *), void * cookie)
+{
+	pid_t pid;
+
+	/* Fork */
```
cperciva, line 19: This is traditionally written with `switch (pid = fork())`, see `fork(2)` for example.

util/fork_func.c
```diff
@@ -0,0 +1,77 @@
...
+
+		/* Warn about the reason for the unclean exit. */
+		if (WIFSIGNALED(status))
+			warn0("pid %jd: signalled with %i",
+			    (intmax_t)pid, WTERMSIG(status));
+		else if (WIFSTOPPED(status))
+			warn0("pid %jd: stopped with %i",
+			    (intmax_t)pid, WSTOPSIG(status));
+		else
+			warn0("pid %jd: stopped for unknown reason.");
```
cperciva, line 71: Is "stopped" what you want here?  I mean, stopped != exited.

util/fork_func.c
gperciva: sigh, ok, I'll change it.  I still think it's easier to leverage our normal "do something; did it fail?" idiom.  Then if it did not fail, then figure out what type of success it had.

BTW, although I've seen the traditional `switch` method before, it's not in the freebsd 12.4 man page, which has no `EXAMPLES` section:
https://man.freebsd.org/cgi/man.cgi?query=fork&apropos=0&sektion=2&manpath=FreeBSD+12.4-RELEASE

Evidently that was only added in 13.1:
https://reviews.freebsd.org/D27626

util/fork_func.c
gperciva: Good catch, changed.

gperciva:
Updated with a REBASE commit.

I also rebased onto master, to clear up the git conflict.

tests/fork_func/check_exec.c
```diff
@@ -0,0 +1,39 @@
...
+
+#include "check.h"
+
+static int
+func_exec(void * cookie)
+{
+
+	(void)cookie; /* UNUSED */
+
+	if (execlp("true", "", NULL))
```
cperciva, line 14: Is there a reason you want to invoke `true(1)` with an argv0 of ""?

tests/fork_func/check_exit.c
```diff
@@ -0,0 +1,45 @@
+#include <sys/types.h>
+
+#include "fork_func.h"
+
+#include "check.h"
+
+static int
+func_exit(void * cookie)
+{
+	int exitcode = *((int *)cookie);
```
cperciva, line 10: I would write this as

```
int *exitcodep = (int *)cookie;

return (*exitcodep);
```

Does the same thing, but we generally follow a convention of "pointer = cookie" at the start of callback functions.

util/fork_func.c
```diff
@@ -0,0 +1,78 @@
...
+	pid_t pid;
+
+	/* Fork */
+	switch (pid = fork()) {
+	case -1:
+		warnp("fork");
+		goto err0;
+	case 0:
+		/* In child process: Run the provided function, then exit. */
+		_exit((*func)(cookie));
```
cperciva, line 26: We generally omit the `*` when calling via a function pointer.  (The C standard says it's optional.)

util/fork_func.c
```diff
@@ -0,0 +1,78 @@
...
+		/* Child ${pid} exited cleanly. */
+		rc = WEXITSTATUS(status);
+	} else {
+		/* Child ${pid} did not exit cleanly. */
+
+		/* Warn about the reason for the unclean exit. */
+		if (WIFSIGNALED(status))
+			warn0("pid %jd: signalled with %i",
+			    (intmax_t)pid, WTERMSIG(status));
+		else if (WIFSTOPPED(status))
```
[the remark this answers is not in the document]
gperciva: I figured that anything other than `WIFEXITED` was unusual and worth reporting?

I mean, the overall point of `fork_func_wait()` is "wait until the forked process is finished".  I don't have a good sense of what to do if there's an error, so I figured that reporting it was a decent lower bound?

tests/fork_func/check_exec.c
gperciva: nope, fixed.

tests/fork_func/check_exit.c
gperciva: fixed.

util/fork_func.c
gperciva: fixed

util/fork_func.c
```diff
@@ -12,7 +12,7 @@
  * Fork and run ${func} in a new process, with ${cookie} as the sole argument.
  */
 pid_t
-fork_func(int (* func)(void *), void * cookie)
+fork_func(int (*func)(void *), void * cookie)
```
gperciva, line 15: wait, now I'm rethinking this one.

in `events/*.c`, there's 6 function pointers defined as `(*func)` with no spaces.  There's only 2 instances of `(* func)`, both of which in cpusupport code I wrote.

OTOH, when I grep for `(* `, there's 56 hits, and at first glance, at least 90% of them are function pointers.

So I'm guessing that the desired style is:
```
int
foo(int (* func)(void *))
```
?

util/fork_func.c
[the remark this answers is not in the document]
cperciva:
```
     WIFSTOPPED(status)
             True if the process has not terminated, but has stopped and can
             be restarted.  This macro can be true only if the wait call
             specified the WUNTRACED option or if the child process is being
             traced (see ptrace(2)).
```

Since we're not passing the WUNTRACED flag (or any flags at all for that matter), the only thing this would report to us is if the process is being traced... in which case the signal value isn't particularly useful anyway.  So I would just handle this with the 'unknown reason' message.

util/fork_func.c
cperciva: Yes, our STYLE says "space between * and variable name".  Applies to function pointers just as much as any other pointers.

util/fork_func.c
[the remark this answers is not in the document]
gperciva: ok, removed

util/fork_func.c
gperciva: I removed the bad commit, and made a separate PR for fixing the style of our function pointers.

gperciva:
I rebased the existing REBASE commits, but then added one more small change.

util/fork_func.c
```diff
@@ -0,0 +1,75 @@
...
+ */
+pid_t
+fork_func(int (* func)(void *), void * cookie)
+{
+	pid_t pid;
+
+	/* Fork */
+	switch (pid = fork()) {
+	case -1:
+		warnp("fork");
```
cperciva, line 22: For consistency with the other `case`s, I would add `/* Error in fork system call. */`

util/fork_func.c
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	/* Print the error status, if applicable. */
+	if (WIFEXITED(status)) {
+		/* Child ${pid} exited cleanly. */
+		rc = WEXITSTATUS(status);
+	} else {
+		/* Child ${pid} did not exit cleanly. */
+
+		/* Warn about the reason for the unclean exit. */
```
cperciva, line 64: Combine these into a single (multi-line) comment.

cperciva:
A few minor nits but generally looks good.  Feel free to make changes and rebase unless there's something you're not sure about.

util/fork_func.c
```diff
@@ -0,0 +1,75 @@
...
+		if (WIFSIGNALED(status))
+			warn0("pid %jd: signalled with %i",
```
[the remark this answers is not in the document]
gperciva: Hold-over from when I was a TA: "use %i for integer, and %f for float".  (Because to a first-year engineering student, "decimal" means something like "1.23".)

Ok, I'll switch to `%d`.

gperciva:
Fixes made, rebased, ready for merge.

PR #490, opened 2023-06-27
scripts: add get_filelist()

gperciva:


gperciva:
This includes #488 and #489.

This standardizes on sorting the list of filenames when we generate them, instead of when we use them.

Having to use `set -o noglob` inside the subshell is unfortunate.  I was initially tempted to avoid that by keeping everything on one line (instead of using a function), but that got unwieldly with the switch from `ls` to `find`.

cperciva:
I'm not sure I understand why you can't use `ls` here..

gperciva:
Hmm.  The more I read about this, the less confident I am in the original PR.

PR #491, opened 2023-07-02
perftests: skip over warm up if warmup_bytes == 0

gperciva:


gperciva:
In many cases, there's no harm in running

    init_func()
    func(0 bytes)
    cleanup_func()

However, this doesn't work if init_func() has side effects.

Notably, if init_func() connects to a server which is only expecting one connection.  In that case, init_func() connects, the timed func() does nothing much with 0 bytes, cleanup_func() disconnects, the server exits, and then it tries to run init_func() for the real data (i.e. with nbytes, not nbytes_warmup); but the server has already exited.

gperciva:
This case arose with under-development spiped perftests.

One solution would be to make spiped `recv-zeros` accept multiple connections, so that it can handle a warmup and then the real bytes.  And yes, I already have a branch for that, and I'll probably end up submitting it as a PR.

But having side effects with `nbytes_warmup == 0` seemed to violate the principle of least surprise.  When I started getting "`send-zeros: Could not connect`", I looked for problems in a bunch of places before I started adding debug printfs to `perftests.c`, because we've been using it for multiple projects for years without (major) problems.

PR #493, opened 2023-08-14
Add const to some arguments

gperciva:


gperciva:
cppcheck 2.11 suggested adding a number of `const` qualifiers.  In some cases, the suggestions were ridiculous (such as for `argv` in `main()`!), but these ones seemed plausible.

Optionally, I could rebase these two commits together.

PR #494, opened 2023-08-15
scripts: misc fixes, and clarify public vs. private with underscores

gperciva:


gperciva:
In POSIX sh, there's no such thing as "public" or "private" functions.

However, functions are allowed to contain underscores [1], so we're adopting the convention of leading underscores indicating a function which is not intended to be called by code in a different file.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_235

gperciva:
Depends on #490.

I'm hoping to shove the tiny changes in [7cfedcc](https://github.com/Tarsnap/libcperciva/pull/494/commits/7cfedcc832557627579af49efa575c6e7bd24742) into another PR.

gperciva:
Removed dependency on #490, ready for review.

PR #495, opened 2023-08-24
Misc

gperciva:


gperciva:
I'm not certain about the last commit.  Are "forever loops" the same thing as "infinite loops"?

> FreeBSD style(9) states that:
>
>    Forever loops are done with for's, not while's.
>
>    ...
>
>     for (;;)
>             stmt;

I suspect that you'll say "let's keep the current way of defining loops; just add a note to STYLE describing our current practice".

cperciva:
I'm interpreting "forever loop" to mean a loop which has no exit conditions, not merely one which doesn't have exit conditions in the for/while clause.  So I'd say both of these are correct:

```
for (;;) {
    ...
}
do {
    if (...)
        break;
    ...
} while (1)
```

gperciva:
Ok, rebased to remove the change to `do... while` loops.

PR #497, opened 2023-09-26
style: fix function pointers

gperciva:


gperciva:
- fix missing space between return type and opening parenthesis
- fix missing space after `*` or `**`

PR #498, opened 2023-09-26
scripts: shorten some function names

gperciva:


gperciva:
A later commit will switch to using "local" variables (indicated by prepending the variable name with the function name), but those would be extremely long with the previous function names.  Shortening them will result in variables which are merely very long.

PR #500, opened 2023-09-26
Actions: use clang-13

gperciva:


gperciva:
sigh.  Github Actions is removing clang-12 from the Ubuntu 22.04 images.  It's a rolling deployment spread over 3-4 days; until we upgrade, we're going to have random build failures.

https://github.com/actions/runner-images/issues/8263

PR #499, opened 2023-09-26
scripts: misc readability refactoring

gperciva:

PR #501, opened 2023-09-26
scripts: use local vars

gperciva:


gperciva:
This builds on #498 and #499.

PR #503, opened 2023-11-09
comments: grammar fixes and clarifications

gperciva:


events/events_network.c
```diff
@@ -451,7 +451,7 @@ events_network_get(void)
 			break;
 		}
 
-		/* Are we ready for reading? */
+		/* Are we ready for writing? */
```
gperciva, line 454: The "reading/writing" was changed in
        2017-01-22 Switch events_network from select to poll
        91f42f7f50f9e1c61ada00f4724b0d761a4b56c0
but it should have remained "writing".

util/sock.c
```diff
@@ -34,6 +34,8 @@ sock_resolve_unix(const char * addr)
 	if ((sa_un = calloc(1, sizeof(struct sockaddr_un))) == NULL)
 		goto err0;
 	sa_un->sun_family = AF_UNIX;
+
+	/* Safely copy addr into the structure. */
```
gperciva, line 38: I keep on getting complaints about the `strcpy()` from static analysis programs, so having the "safely copy" in its own "paragraph" would be helpful in quickly recognizing that it's a false positive.

PR #506, opened 2024-01-01
Import millisleep.h

gperciva:


gperciva:
This seems "too trivial" to submit as a PR by itself, but I've been meaning to combine it with something else in the past 3 months [1] and haven't found anything.  Might as well get it done.

[1] I noticed that I hadn't actually imported it on Sep 20, 2023 https://github.com/Tarsnap/spiped/pull/392#issuecomment-1728425240.

PR #507, opened 2024-01-14
tests/optional_mutex: check counter within a lock

gperciva:


gperciva:
I'm estimating that there's an 80% chance you decline this PR.  Coverity-scan complains about this; we could change the code, or mark it as "false positive".

The variables `value` and `counter` are modified within mutex locks, but the final check is not protected.  I figured that wasn't necessary, since the final check happens after pthread_join(), so there isn't any danger of synchronization problems.  But if it's better code style to slap a mutex there as well, I'm happy to do so.

cperciva:
It's good "lock hygiene" to do this.  I would be very surprised if any system implemented lock-free pthread_join, but *theoretically* even though a thread has exited we can't guarantee that values it has written to memory are visible to us until we acquire a lock.

tests/optional_mutex/main.c
```diff
@@ -95,6 +95,38 @@ test_pthread(void)
...
+	if ((rc = optional_mutex_lock(&mutex)) != 0) {
+		warn0("optional_mutex_lock: %s", strerror(rc));
+		goto err0;
+	}
+
+	/* Check the value and counter. */
+	if ((value == 0) && (counter == counter_expected))
+		rc = 0;
+	else
+		rc = 1;
```
cperciva, line 114: Umm, what happens to this value on line 117?

PR #508, opened 2024-01-14
Fix missing printf arg in fork_func

gperciva:


gperciva:
Whoops!  A rather embarrassing mistake.

... I could have sworn that clang or gcc would warn about this case!

gperciva:
ah, clang warns about such problems in `printf`, but not `warn0`.  Makes sense.  (I didn't check gcc.)

cperciva:
Oops. I should have noticed that...

PR #510, opened 2024-01-26
In printf, use %d instead of %i

gperciva:


gperciva:
I don't know if it's worth making the change, but in https://github.com/Tarsnap/libcperciva/pull/487#discussion_r1451632291, you indicated a preference for `printf` using `%d` rather than `%i`.  This fixes the previously-existing instances of `%i`, in both `.c` files and `.sh` scripts.

cperciva:
We might as well be consistent.  Want to add this to STYLE as well?  ("We use `%d` rather than `%i` in printf strings.")

gperciva:
Updated with a rebase.  I used the sentence
```
We use %d rather than %i in printf and warn0/warnp strings.
```

PR #513, opened 2024-02-15
build: don't link to optional_mutex if it doesn't exist

gperciva:


gperciva:
This has no effect on any project which has `liball_optional_mutex_normal` (such as libcperciva, spiped, and kivaloo).  However, if you updated the build system for imds-filterd (for some crazy reason), you would be unable to compile it, because the build would try to link to `liball_optional_mutex_normal.a`.

This PR is probably pointless for any practical projects, because once we have `noeintr_close`, we'll need `optional_mutex` for everything.  But it's rather untidy to have a build system which just plain breaks if we don't have certain files.

Makefile.inc
```diff
@@ -14,11 +14,22 @@ IDIRS	:=	${IDIRS}
...
 MAN	=
 
+# The SUBDIR_DEPTH variable (defined above) initializes tracking the
+# subdirectory depth, but it's not fully set at this stage in the build
+# process.  Instead, we jump back to the shell and print the (finalized)
+# value of SUBDIR_DEPTH.
+FINALIZED_SUBDIR_DEPTH	!=	${MAKE} -v SUBDIR_DEPTH
+
 .if !defined(NOLIBALL)
+.if exists(${FINALIZED_SUBDIR_DEPTH}/liball/optional_mutex_normal)
```
gperciva, line 24: I have no firm feelings about what we should have here.  We could spell out the full path to `liball_optional_mutex_normal.a` (or even that and `liball_optional_mutex_pthread.a`!).

I initially had the full path, but it looked messy, and I figured the directory would be sufficient to be confident that the project either had optional_mutex or not.

Makefile.inc
```diff
@@ -14,11 +14,22 @@ IDIRS	:=	${IDIRS}
 # Default is no man pages
 MAN	=
 
+# The SUBDIR_DEPTH variable (defined above) initializes tracking the
+# subdirectory depth, but it's not fully set at this stage in the build
+# process.  Instead, we jump back to the shell and print the (finalized)
+# value of SUBDIR_DEPTH.
+FINALIZED_SUBDIR_DEPTH	!=	${MAKE} -v SUBDIR_DEPTH
```
cperciva, line 21: I'm not sure how this doesn't result in an infinite loop.  Running `make -V SUBDIR_DEPTH` would process the Makefile and come across this line again, right?

gperciva: Short answer: because `make -v SUBDIR_DEPTH` gets it from `Makefile` (where it's defined in full), not `Makefile.BSD`.

Medium answer: I'll update the comment.

Better answer: hmm, that's weird, `make -f Makefile.BSD -v SUBDIR_DEPTH` and `-V FINALIZED_SUBDIR_DEPTH` both work, instead of going into an infinite loop.  I need to investigate more.

gperciva: Addendum to that "better answer": even if we run `make -f Makefile.BSD -v SUBDIR_DEPTH` on the command-line, the top-level `Makefile.inc` runs `make -v SUBDIR_DEPTH` (which has the fully defined value) so there's no loop.

gperciva:
Updated.  The comment for `FINALIZED_SUBDIR_DEPTH` states why it avoids an infinite loop, and I added a `LIBALL +=` for the optional_mutex stuff.

Makefile.inc
cperciva: Hmm, so what happens if we run `make Makefiles` and there isn't a Makefile already?  We get a broken one and then it gets fixed the next time we run `make Makefiles`?

gperciva: Not broken.  (Although admittedly, that's a happy accident rather than by design.)

`metabuild.sh` line 25:
```
SUBDIR_DEPTH=$(${MAKEBSD} -v SUBDIR_DEPTH)
```

That works fine regardless of whether there's a `Makefile` or not.

`metabuild.sh` line 172 puts `SUBDIR_DEPTH` into the `Makefile`.  Until that point,
```
.info ${FINALIZED_SUBDIR_DEPTH}
```
in `Makefile.inc` shows that it's empty.  But after line 172 of `metabuild.sh`, it's fine.

Line 179 deals with liball via `copyvar_LIBALL_optional_mutex`.

I've added a comment to `metabuild.sh` to note this dependency.

cperciva:
I wish there was a way to do this without needing the FINALIZED_SUBDIR_DEPTH hack, but I don't see any better alternative.

PR #514, opened 2024-02-20
Add & use ipc_sync.h

gperciva:


util/ipc_sync.h
```diff
@@ -0,0 +1,183 @@
...
+	/* Failure! */
+	return (-1);
+}
+
+/**
+ * ipc_sync_signal(IS):
+ * Indicate that ${IS} should no longer block.
+ */
+static inline int
+ipc_sync_signal(struct ipc_sync * IS)
```
gperciva, line 123: This function could take `struct ipc_sync IS` directly (not as a pointer).  But I found it confusing to have 3/4 functions being called with `(&IS)` and 1/4 with `(IS)`.

If you like the PR in theory but want this one changed, that's of course easily done.

gperciva:
1. I submitted the `.h` version because I wanted to keep the wait and signal functionality as unchanged as possible -- if that code ended up in a different `.o` file, would it receive the same optimization that it did previously?

    I do prefer the `.c` version, so I'll put that up.

2. will do.

3. `_done` could be called by `signal` and `wait`, but it includes `close()`, which might be slow.  (Especially when we switch to `noeintr_close`.)  So I figured that the signal & wait should be kept as fast as possible (e.g., inside timed code), while `_done` could be deferred until the caller wanted it.

PR #515, opened 2024-02-24
Import

gperciva:


gperciva:
We got a bit out of sync; `wait_while` came from https://github.com/Tarsnap/kivaloo/pull/294, while the fix for `CFLAGS_DEFAULT` came from https://github.com/Tarsnap/spiped/pull/398.

PR #516, opened 2024-03-08
apisupport: detect if we need `-Wno-deprecated-declarations` for AES

gperciva:


gperciva:
OpenSSL 3.0 marked the low-level AES functions as deprecated, and warned that they might be removed completely in the future.

I think it makes sense to use those functions for as long as possible.  If/when we ship our own software implementation, if a security flaw was detected in it, we'd have to make a new release ASAP.  By contrast, if we're using the openssl version, then users only need to update their libssl packages from their operating system.

As such, compiler warnings about deprecations when compiling `crypto_aes.c` aren't useful; we know very well that we're using those functions.

PR #517, opened 2024-04-23
Add & use ipc_sync.h and ipc_sync.c

gperciva:


gperciva:
This replaces #514.

The API is slightly different: it now includes `_wait_prep()` and `_signal_prep()`, which are optional functions.

util/ipc_sync.c
```diff
@@ -0,0 +1,238 @@
...
+	return (0);
+
+err0:
+	/* Failure! */
+	return (-1);
+}
+
+/* close(), but looping upon EINTR. */
+static inline int
+ipc_sync_close(int fd)
```
gperciva, line 76: Two notable things about this function:
1) we will replace it once we have `noeintr_close()`.
2) this function retains the previous behaviour of `daemonize.c`, namely that if `close()` has `EINTR`, this function calls `close()` again.  This behaviour is required for HP-UX, but is explicitly wrong for the glibc implementation of `close()`: https://www.man7.org/linux/man-pages/man2/close.2.html

I propose that we keep the same behaviour as before (which is broken on linux), but I thought that I should bring it to your attention.

util/daemonize.c
```diff
@@ -104,35 +65,22 @@ daemonize(const char * spid)
...
+		if (ipc_sync_wait(IS))
 			goto err1;
 
 		/*
-		 * We don't need to read from this any more.  The pipe would
-		 * be implicitly closed by the following _exit(), but we're
-		 * explicitly closing it for the benefit of leak checkers.
+		 * Free resources associated with ${IS}.  These would be
+		 * released by the following _exit(), but we're explicitly
+		 * closing it for the benefit of leak checkers.
```
cperciva, line 80: s/closing it/releasing them/

util/ipc_sync.h
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Normal usage:
+ * 1) ipc_sync_init()
+ * 2) fork a new process
+ * 3) process x: ipc_sync_wait()
+ *    process y: ipc_sync_signal()
+ * 4) both processes: ipc_sync_done()
+ *
+ * If there's a need to make _wait() or _signal() as fast as possible
+ * (i.e. for a benchmark or a performance bottleneck), call the
```
cperciva, line 13: s/i.e./e.g./

cperciva:
Ok aside from very minor comment nits.  Please fix and then I'll merge.

PR #524, opened 2024-07-31
Extra build checking, forward defines, and comments

gperciva:


datastruct/mpool.h
```diff
@@ -43,14 +46,23 @@ static inline void *
...
 {
 
+	/* Statistic about the number of mpool_malloc() calls. */
 	M->nallocs++;
+
+	/* If we have an object on the stack, use that. */
 	if (M->stacklen)
 		return (M->allocs[--(M->stacklen)]);
+
+	/* Statistic about the number of malloc() calls. */
```
gperciva, line 56: This is my best explanation for `M->nempties`, but I'm at a loss to explain why you called it "number of empties".  It's quite possible that I'm missing some cache / pool technical terminology, and a different explanation would be more suitable.

cperciva: `nempties` = the number of times the pool was empty when we tried to allocate.  Idea being that if the pool ends up being empty a lot, we need to make it bigger so we don't run out of pre-allocated buffers as often.

gperciva: Hmm.  If we're making an analogy to a physical pool, then I'd say that we need a larger pool if it's *already filled*, not *empty*.

But hey, I'm not suggesting that we rename the variables, and the comments don't make mention of a pool being filled or empty.  So I think this is still ready for merge (after review).

cperciva: Well, yes, but we don't want to keep an unlimited number of allocations.  We want to keep a number of allocations in line with what we're going to need again in the future, and we don't have a time machine.  So we count the number of times we didn't have an allocation to reuse and increase the size of the pool in the hopes that *next time* we'll have enough.

cperciva: Anyway as far as the comment is concerned, I would make it `/* Number of allocation requests where the pool was already empty. */` to explain the variable name.

gperciva: Umm... ok, done, rebased, ready for merge.

But I suspect that you might not have seen the actual context.  (i.e. maybe you've been reacting to what the github emails send you)

The lines in question are:
```
/* Number of allocation requests where the pool was already empty. */
M->nempties++;
```

Granted, the original message didn't make any mention of the `++` either.

One option might be:
```
/* 
 * Increment the number of allocation requests where the pool was
 * already empty.
 */
```
but that's two lines.  Or maybe:
```
/* The pool is empty; increment the relevant statistic. */
```

cperciva: Ok, looking at more context: I would say

```
/* Count the total number of allocation requests. */
M->nallocs++;
```

and

```
/* Count allocation requests where the pool was already empty. */
M->nempties++;
```

cperciva: The point is to communicate that we're incrementing `M->nallocs` and `M->nempties` to track the number of times something happens, and what that something is, in each case.

gperciva:
Fixed and rebased, ready for merge.

PR #525, opened 2024-08-08
Add a comment for every return

gperciva:


gperciva:
With two exceptions[1], this PR ensures that every `return` statement is annotated.

- main change: added `/* Done! */` to `STYLE` (and apply).

    This is for things like `return (ret)` where it's not appropriate to say either `/* Success! */` or `/* Failure! *`/.  Also for returning a mathematical result, such as `return (state)`.
 
- secondary change: added `/* Do nothing. */` where appropriate.
- third: described the 8 remaining `return`s which were left un-annotated.

[1] Did not comment `sysendian.h`, or the `return (mpool_malloc(...` because that was in the middle of a huge macro.

cperciva:
I like the first and third commits but not the second aside from the setgroups_none part.  I think of `/* Done! */` as being a comment which is useful when there are multiple return paths -- we might be done, or we might have failed -- and it's especially unuseful for the one-line functions in /POSIX/.

util/optional_mutex.c
```diff
@@ -51,6 +52,7 @@ optional_mutex_unlock(pthread_mutex_t * mutex)
 {
 
 #ifdef OPTIONAL_MUTEX_PTHREAD_YES
+	/* Done! */
```
cperciva, line 55: This (and the preceding one) could be made more useful e.g. `/* Mutexes are enabled; call the pthread function. */`.

gperciva:
Revised to omit the second commit, and make small changes to the other commits as requested.

cperciva:
Looks good, please tidy up the git history.

gperciva:
Rebased, ready for merge.

PR #528, opened 2024-12-08
Actions: update to macos 13

gperciva:


gperciva:
macos-12 is now deprecated, and we might as well automatically track the most recent one, rather than manually switching versions.

gperciva:
I originally wanted to switch to `macos-latest`, but openssl is still broken on that.

This is merely kicking the can down the road for another few months, but hopefully that'll give github enough time to fix their images.

PR #530, opened 2025-01-12
sock_util: add sock_addr_validate()

gperciva:


gperciva:
I considered adding test cases to `tests/sock_util/` but I decided it wasn't worth it.

If you think it's worth it, I'm happy to add them.

cperciva:
Not essential but if you have time to kill it never hurts to have extra tests.

PR #531, opened 2025-01-13
Add headers

gperciva:


gperciva:
This is the last of the spiped-related nitpicks.

PR #535, opened 2025-04-23
Fix warnp strings

gperciva:


gperciva:
I kept these as separate commits for ease of reviewing, but they should probably be smooshed together in the git history.  This can be done in the github interface, or I could rebase them myself.

cperciva:
Can you add a few lines to STYLE about this?  (Probably not necessary to mention "print the name of the right function", but worth saying "write `foo` rather than `foo()`" and "no terminating punctuation or line break".)

gperciva:
Updated

gperciva:
Oops, there's a few more cases.  STYLE now says "no terminating punctuation", but:

```
$ git grep "warn" | grep -e "?\"" -e "\!\""
crypto/crypto_dh.c:		warn0("Exponent result too large!");
http/http.c:		warn0("HTTP response with major version > 1!");
tests/aws/main.c:		warn0("That should have failed!");
util/entropy.c:			warn0("EOF on /dev/urandom?");
```

Arguably these case are unusual enough to warrant punctuation, in which case I should modify STYLE.

cperciva:
Yeah I think those cases deserve their punctuation.

PR #537, opened 2026-03-21
Increase portability of build

gperciva:


gperciva:
This allows `bmake Makefiles` to work on Debian.

It seems that `<bsd.prog.mk>` and `<bsd.lib.mk>` have undocumented effects on some non-FreeBSD systems:
- they can include files multiple times (fixed by the "include guards" commit)
- they can include `./Makefile.inc` in addition to `../Makefile.inc`, which is completely undocumented in `/usr/share/mk/README`.

This PR also includes catching up a Makefile to a commit from over a year ago; it didn't seem worth committing a `make Makefiles` as the time.

gperciva:
To elaborate on "`<bsd.prog.mk>` will include a `Makefile.inc` in the same directory:

```
gperciva@raspberrypi:/tmp $ more Makefile*
::::::::::::::
Makefile
::::::::::::::
.include <bsd.lib.mk>
::::::::::::::
Makefile.inc
::::::::::::::
.error "you shouldn't be parsing this file!"
gperciva@raspberrypi:/tmp $ bmake
bmake: "/tmp/Makefile.inc" line 1: "you shouldn't be parsing this file!"
gperciva@raspberrypi:/tmp $ 
```

Very weird behaviour, but given the include guards, it's easy to work around.

cperciva:
Just to make sure I understand this: We're getting `Makefile.inc` files sucked into bmake inappropriately all over teh place, but the only place that causes problems is in `/liball/` ?

gperciva:
Not quite.  We're getting the top-level `Makefile.inc` included at least twice, which confuses bmake because it's defining things multiple times.  The "include guards" fixes that, and allows all but one `Makefile` to generate correctly.

As a separate issue, `liball/Makefile.BSD` is including `liball/Makefile.inc` on debian's `bmake`.  That doesn't happen on FreeBSD, and shouldn't happen according to the docs.  We can fix this by setting a variable in `liball/Makefile.BSD`.

cperciva:
Oh, I see, `liball` is the only directory where we have both `Makefile.BSD` and `Makefile.inc` in the same directory.

cperciva:
Should I merge this now or did you want to wait and see if this works on NixOS first?

gperciva:
Merge it now; I've just given up on nixos installation for tonight.  "Makes it work on debian" is a clear-cut goal.  I might spin up an EC2 instance instead, since they have AMIs available.

I also have a vague recollection that Nix isn't POSIX-compatible, so it might require more invasive workarounds.

PR #543, opened 2026-07-24
Fix 2 violations of network_connect's API contract

gperciva:


gperciva:
`network_connect()` says that upon fatal error... invoke `${callback}(${cookie}, -1)`.  There's two instances in which that did not happen:
* if `getsockopt()` failed
* if `tryconnect()` called from `dofailed()` failed

This was reported in https://github.com/Tarsnap/tarsnap/issues/714; I checked that faking an `events_immediate_register()` failure caused the heap-use-after-free error in ASan.  I'm tentatively guessing that you'll say that each of these is $10, but it's possible you'll say that it's higher because breaking an API contract is bad.

I'm reasonably confident that the code in these commits is good (after many iterations), although you might prefer me to rebase the refactor and the `dofailed()` together.

I'm not at all confident that the commit subject line and messages say the right thing.

cperciva:
Hmm, interesting.  I had previously been thinking in terms of "when callbacks return -1 that indicates a fatal error and the event loop is about to abort, so we don't need to worry about cleanup" but that was before you added the dispatch_shutdown cleanup path.  Might be worth revisiting all of the other places where I have callback functions which can fail to audit whether there are other places where we're missing a callback which can be used for error cleanup.

cperciva:
Patches look good.  The second commit message should s/remove/hoist/ though -- we're not getting rid of the free, we're doing it outside of the function.

gperciva:
Updated commit subject, rebased.

PR #547, opened 2026-07-28
Fix style & typos

gperciva:


gperciva:
This is the first batch of Claude fixes from March.  Many more to come, including a few more style or typo fixes which merited a second look.

PR #548, opened 2026-07-28
Explain some design decisions

gperciva:


gperciva:
This is another set of Claude fixes from Feb / March.

This only touches comments; no functional changes at all.

PR #549, opened 2026-07-29
Sanitize buffers

gperciva:


gperciva:
The first 2 commits are more of Claude's fixes from Feb.  The 3rd commit is me sanitizing `key_secret` from `tests/aws/main.c`.

That "secret" key is hard-coded in our repository, so it's pointless in terms of security, but in case anybody looked at that test as a template for how to use `aws_sign.h`, there was no harm wiping the key.

PR #550, opened 2026-07-29
elasticarray: _exportdup() for empty array, option alpha

gperciva:


gperciva:
Two alternate PRs: #550 and #551.

option alpha #550 is what you wrote in Feb.

option beta #551 is my alternative.  In particular: 1) this makes `_exportdup()` return the same values as `_export()`, and 2) the code is a better match for existing code in `_truncate()` and `resize()`.

PR #551, opened 2026-07-29
elasticarray: _exportdup() for empty array, option beta

gperciva:


gperciva:
Two alternate PRs: #550 and #551.

option alpha #550 is what you wrote in Feb.

option beta #551 is my alternative.  In particular: 1) this makes `_exportdup()` return the same values as `_export()`, and 2) the code is a better match for existing code in `_truncate()` and `resize()`.

cperciva:
I was initially returning malloc(1) in order to avoid users of this API running into "copy zero bytes from NULL" bugs -- but as you say, it makes more sense to match what `_export` does, so let's go with this one.

gperciva:
Ready for merge now.

PR #552, opened 2026-08-02
Nitpicks

gperciva:


POSIX/posix-cflags-filter.sh
```diff
@@ -23,7 +23,7 @@ if ! ${CC} -O2 "${D}/posix-trivial.c" 2>/dev/null; then
 		echo '	if [ "$OPT" = "-O2" ]; then'
 		echo '		continue'
 		echo '	fi'
-		echo '	CFLAGS_FILTERED="$CFLAGS $OPT"'
+		echo '	CFLAGS_FILTERED="$CFLAGS_FILTERED $OPT"'
```
gperciva, line 26: Ok, this fix isn't merely a "nitpick".  But it seemed to fit thematically with "a set of small fixes that have no real effect", and after a few seconds I decided to stop agonizing about the name of the branch & PR.

gperciva:
Also: should I be giving imaginary bounty values for Claude-suggested fixes?

cperciva:
> Also: should I be giving imaginary bounty values for Claude-suggested fixes?

Yes.  Obviously since it's not "real money" it's less essential, but I think it's useful nonetheless for historical purposes to be able to say "here's the value of the bugs Claude found" without needing to re-inspect every patch.  (I've strongly encouraged people to include LLMs in security advisory credits for this reason too.)

gperciva:
Ok, I've added some bounty suggestions.

PR #554, opened 2026-08-04
STYLE: ignore failures or callback returns in error paths

gperciva:


gperciva:
ref your comment here: https://github.com/Tarsnap/spiped/pull/427#discussion_r3609446416

PR #555, opened 2026-08-11
Improve atexit handling

gperciva:


gperciva:
More commits from Feb.  2 of these are recent: Claude didn't complain about mpool and getopt, so those commits don't have any such attribution in the commit message.

PR #556, opened 2026-08-12
Fixes for json

gperciva:


gperciva:
Claude fixes for json from Feb.

As discussed briefly at the time, I split the "allow whitespace" away from the "Fix skipping arrays/objects with whitespace" commit.  There's also other editing, which I kept separate rather than changing the commits with your names on them without you taking a look at it.

What's the intended scope of `json.c`?  There's quite a few cases where it parses invalid json, or fails to parse valid json.  Do we want this to be fully compliant, or is "works with AWS" good enough?

cperciva:
> What's the intended scope of json.c? There's quite a few cases where it parses invalid json, or fails to parse valid json. Do we want this to be fully compliant, or is "works with AWS" good enough?

The intended scope is exactly what it says: If there is a valid JSON object *and* it contains a name-value pair with the specified name, return a pointer to the value.  If we're passed invalid JSON, we have no obligations; we're trying to extract a value, not validate the entire blob.

If there's a case where valid JSON is not handled, that's a bug.

gperciva:
Hmm.  6e197472ddbcdd58418f8414ffde7ceedd488ed7 adds two "non-match because JSON is invalid" entries; one of which came from `0021-tests-json-Add-3-more-test-cases.patch` in Feb.

Is the idea to add a few test cases of invalid json only when those cases triggered a memory-access bug?  I'm wondering if we should be testing invalid json at all.  Or maybe add a few more cases that already work, as a regression test?

cperciva:
Well, we should never have buffer overruns even if the JSON input is invalid.

And I'm not saying we can't reject broken JSON, just that it's not the priority.  (And in particular we shouldn't aim for *completeness* since that would require far more code, which would increase the risk of introducing more bugs.)

gperciva:
I haven't made any further changes after clarifying the scope, so this is still good for review.  In particular, checking that my changes to your Feb commits are ok.

cperciva:
LGTM, please tidy up the git history.

gperciva:
Rebased, ready for merge.

PR #558, opened 2026-08-14
sock_resolve(): allow port 0 for [x.y.z.a] addrs

gperciva:


gperciva:
I'm not certain if this should accept 0, or if we need to restore (and clarify) a lost comment.

`sock_resolve()` accepted:
    localhost:0
but not
    [127.0.0.1]
    [127.0.0.1]:0

If the addr is a host name, it's handled by `sock_resolve_host()`, which makes no requirements about the port number.

When this file was added in:
>
>    2011-09-07 Add code taken from spiped, kivaloo, and scrypt.
>    a7630848193ff232b5bae080af53a098560fe28b

it used a strtol() for the parsing, and the comment above it was:

    /*
     * Parse the port number.  If strtol fails to parse the port number,
     * it will return 0; but that's fine since port 0 is invalid anyway.
     */

That comment was lost when we switched to PARSENUM_EX in:
    2019-09-19 sock: use PARSENUM_EX
    44a0edb293980d237125079b65b1e85159880b98
so I'm not certain what the rationale was.

Not accepting :0 became more of a problem when we added sock_addr_ensure_port(), which deliberately adds :0 if there's no port number.

gperciva:
(changed to draft because the current commit message is almost certainly not what we want to keep)

Relevant info: a few years ago, we added `-b <bind address>` to spiped, which in turn prompted `sock_addr_ensure_port()` which adds a port number of `:0` if there wasn't any.

cperciva:
I think the "lost comment" is because PARSENUM_EX is checking both that the port number is parsable and that it's between 1 and 65535, so the quirk of strtol returning 0 for both "0" and error isn't relevant.

gperciva:
The first half of the second sentence, yes.  But the second half of the "lost comment" is "but that's fine since port 0 is invalid anyway."

I vaguely remember that at some point you said "it's ok if a host name has no port, because people can be vague about that; but if they're going to be exact with [x.y.z.a] then they should be exact about the port as well".  But if I'm not hallucinating that statement, we should probably include that as a code comment or even an API contract.

cperciva:
I'm inclined to say that we need to split sock_resolve into sock_resolve (current behaviour) and sock_resolve_bindable (allows port 0), which both call sock_resolve_internal(addr, minport).

Port 0 is invalid for an address to listen or connect to, but it is valid as something for an outgoing connection to be bound to (in which case 0 means "any port").

gperciva:
Hmm.  We already have `sock_resolve_one(addr, addport)`.  If `addport` is non-zero, then obviously we'd want that to allow port 0.

Are you sure you want a new `sock_resolve_bindable()` instead of adding an `(int allow_port_zero)` argument?

cperciva:
Oh, it could be done a different way.  But I don't think we should change the API for existing functions, just because that forces changes everywhere they're currently being used.

gperciva:
ok

gperciva:
Updated with `sock_resolve_internal()` and `sock_resolve_bindable()`.

PR #560, opened 2026-08-15
events_network: start with 16 fds

gperciva:


gperciva:
This arose from Feb's `0002-events_network-Don-t-set-fds-to-NULL.patch`, but I revised it considerably to start with space for 16 fds (since that's what `growpollfd()` did).

Other than that, I have absolutely no evidence to suggest 16 rather than 8 or 1.

PR #561, opened 2026-08-18
Theoretical issues with crypto_aesctr

gperciva:


crypto/crypto_aesctr_shared.c
```diff
@@ -27,7 +27,10 @@ crypto_aesctr_stream_cipherblock_generate(struct crypto_aesctr * stream)
 	if (stream->pblk[15] == 0) {
 		/*
 		 * If incrementing the least significant byte resulted in it
-		 * wrapping, re-encode the complete 64-bit value.
+		 * wrapping, re-encode the complete 64-bit value from bytectr;
+		 * this happens once every 256 blocks.  We also rely on this to
+		 * finish initializing the stream, as crypto_aesctr_init2() set
+		 * pblk[15] <- 0xff.
```
cperciva: `pblk[15] = 0xff` ?

cperciva:
Minor commit message tweak but everything else looks fine.  Fix that and I'll merge.

crypto/crypto_aesctr_shared.c
gperciva: I was deliberately using the psuedocode notation of "assignment via an arrow" to avoid the confusion of math `=` vs. programming assignment `=`, but evidently that wasn't the best idea here.

I'll change it to `=`.

gperciva:
Rebased, adjusted comment, and also rewrapped that comment at 78 chars.

PR #563, opened 2026-08-25
pull_request_template.md: add

gperciva:


gperciva:
Add a PR template that _might_ cut down on AI slop.

I checked that HTML comments are not displayed, so in that respect there's nothing to lose.  Human contributors will probably delete the comment anyway, which adds a slight amount of friction to submitting a PR, but it's a tiny amount compared to producing a good-quality commit, so it's probably ok?

PR #564, opened 2026-09-02
Parsenum: reject NaN for bounded floats

gperciva:


gperciva:
This still allows NaN for unbounded floats and doubles.

We could easily forbid those as well, but the test suite explicitly checks that we parse NaN.  I don't think we actually use that for anything, but I figured that rejecting it for bounded ranges (for which it clearly doesn't belong) was the least invasive way to go.

Fixes https://github.com/Tarsnap/scrypt/issues/423 and assorted copycat bug bounty claims.

util/parsenum.h
```diff
@@ -89,8 +89,7 @@ _Pragma("clang diagnostic pop")
 		(((*(x)) = 1, (*(x)) /= 2) > 0)	?			\
 			(((base) == 0) ?				\
 				((*(x)) = parsenum_float((s),		\
-				    (double)-INFINITY,			\
-				    (double)INFINITY, (trailing))) :	\
+				    0, 0, (trailing), 0)) :		\
```
cperciva, line 92: The change from [-INFINITY, INFINITY] to [0, 0] needs to be explained in the commit message.

util/parsenum.h
```diff
@@ -151,7 +151,7 @@ parsenum_float(const char * s, double min, double max, int trailing,
 	val = strtod(s, &eptr);
 	if (eptr == s || (!trailing && (*eptr != '\0')))
 		errno = EINVAL;
-	else if (bounded && ((val < min) || (val > max)))
+	else if (bounded && (isnan(val) || (val < min) || (val > max)))
```
cperciva: Rather than checking with `isnan`, how about checking that `(val >= min) && (val <= max)`?  That fails on NaNs since they compare false to everything.

gperciva:
Updated by rebasing, to fix the commit message.

I'm not a huge fan of
```
       else if (bounded && (!((val >= min) && (val <= max))))
```
but not my call, so ok.

Issue #568, opened 2026-09-08
[bug bounty] PARSENUM() accepts negative strings for 64-bit unsigned targets


maro20066600:
**LLM disclosure, per this repository's `AGENTS.md`:** I am an LLM (Claude),
reporting on behalf of the account owner. I am available to discuss this
report, answer follow-up questions, and revise the reproduction.

## Summary

`PARSENUM()` accepts a negative string for unsigned targets whose type is
64 bits wide, silently storing the value `strtoumax()` wrapped it to. The
documented contract is that a value outside `[min, max]` is rejected with
`ERANGE`, and `-1` is not in `[0, SIZE_MAX]`.

`spiped -n -1` is accepted and runs with an unlimited connection limit;
`spiped -n abc` and `spiped -n 1.5` are correctly rejected. This is a
correctness/input-validation bug, not a memory-safety or security one.

## Cause

`parsenum_unsigned()` (`util/parsenum.h`) parses with `strtoumax()`, which by
C99 7.22.1.4p5 accepts an optional `-` and negates the result modulo
`UINTMAX_MAX + 1`. The sign is never inspected; the only defence is the range
check:

```c
	val = strtoumax(s, &eptr, base);
	if (eptr == s || (!trailing && (*eptr != '\0')))
		errno = EINVAL;
	else if ((val < min) || (val > max) || (val > typemax))
		errno = ERANGE;
```

For the two-argument `PARSENUM(x, s)` form, `PARSENUM_EX4` sets both `max` and
`typemax` from `(*(x)) = -1`, i.e. the target type's maximum. When that type is
64 bits, the wrapped value cannot exceed it, so none of the three comparisons
can ever fire and every negative input is accepted.

Narrower types are unaffected: for a `uint32_t` target the wrapped value
(`UINT64_MAX`) exceeds `typemax`, so `ERANGE` is raised correctly. Explicitly
bounded calls are also unaffected -- `GETOPT_PARSENUM_WITHIN_UNSIGNED(&params.logN, 10, 40)`
in `scrypt/main.c` rejects `-1`, because the wrapped value exceeds 40.

## Reproduction: the macro

```c
/* pn.c */
#include <stdio.h>
#include <stdint.h>
#include <stddef.h>
#include <errno.h>
#include "parsenum.h"

int
main(void)
{
	size_t sz;
	uint32_t u32;

	printf("size_t   \"-1\": %s", PARSENUM(&sz, "-1") ? "REJECT\n" : "");
[...]

PR #569, opened 2026-09-10
parsenum_unsigned: don't allow negative values

gperciva:


gperciva:
Fixes https://github.com/Tarsnap/libcperciva/issues/568

cperciva:
I don't see this as a bug.  `-1` is just a different way of writing `UINTMAX_T`.  I guess it could be documented better, but I don't see that anything here needs to be fixed.

gperciva:
Hmm.  I thought this was a neat solution.  I'm not certain where it's actually useful to parse "-1" into an unsigned variable?

At the moment, there's an inconsistency:
```
test_parsenum: PARSENUM(uint32_t, "-1") failed
test_parsenum: PARSENUM(uint64_t, "-1") succeeded: 18446744073709551615
```

Should I rework `parsenum_unsigned()` so that in the 32-bit case, it does it modulo 32 instead of just failing?  (and ditto for a theoretical 16-bit case)

gperciva:
Oh wait, there's nothing theoretical about `uint16_t` and `uint8_t`.  Anyway, those both fail.  And if we were on a system with a `uint128_t`, `uint64_t` would also fail.

PR #570, opened 2026-09-12
Misc buffer-related fixes

gperciva:


gperciva:
This is another batch of fixes from Claude analysis in Feb & March.

PR #571, opened 2026-09-14
network_ssl_{read,write}: unset callback on error

gperciva:


gperciva:
Fix https://github.com/Tarsnap/tarsnap/issues/725.

The original reporter claimed that we also need
```
	ssl->read_needs_r = ssl->read_needs_w = 0;
```
but that's already set a few lines above, so I (and claude) think that's unnecessary.

PR #573, opened 2026-09-22
HTTP headers are case-insensitive

gperciva:


gperciva:
One fix is straight from February, and the other is a heavily-revised version of a February fix.

